PR #3, opened 2015-05-01
Multiple updates

rth:
Hi Dan,

first of all, thank you so much for putting this module online. I have been looking for a python implementation of the BASEX algorithm for a while and it is really helpfull!

I did update several things and which are included in this pull request.

 1. Rearanged the code base so it can be used as a standalone module. 
     It can be installed with
             
             python setup.py install --user

     or if you want to change the source files without having to reinstall all the time,

            python setup.py develop --user

     and then importing the base class as `from BASEX import BASEX` ( see for instance the updated Readme [file](https://github.com/luli/pyBASEX) )
 2. Added some unit tests that can be run from the `pyBASEX` folder with,
  
            nosetests -s BASEX/test/

 3. Translated the `BASIS2.m` matlab script to a python  function in  `BASEX/basis.py` and also did some optimisation of it. So there are three different matlab versions `BASIS{,1,2}.m`. I'm not sure what the difference in the output is, but the `BASIS2.m` seems to be producing the basis set you were using. Now I have tests that the output of my `generate_basis_sets` function is consistent with  `BASIS2.m`  up to n=200 , checked by unit tests in `BASEX/tests/test_basis_set.py` (which would suggest that the algorithm is correct), larger number were just too slow to generate with `BASIS2.m` . However I'm not able to exactly reproduce the 1000x1000 basis set file with this python function. I'm not sure where the problem is. It could be that this is somehow machine dependent, and since it was taking forever to regenerate this basis set with matlab, I didn't rerun it on my machine The inverse Abel transform obtained with the new basis looks consistent be consistent. So I would say it is working now, but it might be usefull to add a few more unittests, particularly if there is a way to check the validity of generated `M` and `Mc` matrices. Altough, I have just translated the .m file, I have not checked if is correct mathematically with the respect to the original paper.

 4. So I changed a bit the workflow, since we have the ability to regenerate the basis set and is fairly fast (e.g. 1min 30sec for a n=1000 and nbf=500 basis), as illustrated in the Readme. So there in a fist step we load or generate the basis, and then calculate the inverse Abel transform. This is done so we can for instance, run the transform in a loop on a list of different files, without having to reload the basis set all the time. I also updated all the examples under `examples/`

 5. Other minor changes include for instance using a matrix view of numpy arrays, with `X.view(np.matrix)`, so we can do simply `X*Y`, instead of `X.dot(Y)` and `X.I` instead of `np.inv(X)`, so the formulas are more readable.

 6. Added the ability to take in 1D `rawdata`, mostly for testing.

 7. Minor fixes, so the module is compatible with both Python 2.7 and 3

  BTW, it's a 2002 paper not 2012  :) 

Please let me know if this works for you, and sorry if there are lot of things changing at the same time which is difficult to follow. Some functions I updated might need to be more documented,


DanHickstein:
Dear Roman,

Thank you so much for taking the time to make these changes to the pyBasex
program! Your Python skills are a clear improvement over my hacking :)

I will try to take a look at the modifications that you have made tomorrow
and incorporate the pull request.

When you say that the basis sets that you have generated are not the same
as those generated by the matlab program, how different are they? Some
small fraction? It's possible that there is just some numerical difference
between what is being done in the matlab implementation and what you have
done in python. As you say, if they produce nearly the same transformation,
it's probably not a significant difference. But, we should be careful.

Thanks,

Dan




On Fri, May 1, 2015 at 1:26 PM, Roman Yurchak <notifications@github.com>
wrote:

> Hi Dan,
>
> first of all, thank you so much for putting this module online. I have
> been looking for a python implementation of the BASEX algorithm for a while
> and it is really helpfull!
>
> I did update several things and which are included in this pull request.
>
>    1.
>
>    Rearanged the code base so it can be used as a standalone module.
>    It can be installed with
>
>         python setup.py install --user
>
>    or if you want to change the source files without having to reinstall
>    all the time,
>
>        python setup.py develop --user
>
>    and then importing the base class as from BASEX import BASEX ( see for
>    instance the updated Readme file <https://github.com/luli/pyBASEX> )
>     2.
>
>    Added some unit tests that can be run from the pyBASEX folder with,
>
>        nosetests -s BASEX/test/
>
>     3.
>
>    Translated the BASIS2.m matlab script to a python function in
>    BASEX/basis.py and also did some optimisation of it. So there are
>    three different matlab versions BASIS{,1,2}.m. I'm not sure what the
>    difference in the output is, but the BASIS2.m seems to be producing
>    the basis set you were using. Now I have tests that the output of my
>    generate_basis_sets function is consistent with BASIS2.m up to n=200 ,
>    checked by unit tests in BASEX/tests/test_basis_set.py (which would
>    suggest that the algorithm is correct), larger number were just too slow to
>    generate with BASIS2.m . However I'm not able to exactly reproduce the
>    1000x1000 basis set file with this python function. I'm not sure where the
>    problem is. It could be that this is somehow machine dependent, and since
>    it was taking forever to regenerate this basis set with matlab, I didn't
>    rerun it on my machine The invers e Abel t ransform obtained with the new
>    basis looks consistent be consistent. So I would say it is working now, but
>    it might be usefull to add a few more unittests, particularly if there is a
>    way to check the validity of generated M and Mc matrices. Altough, I
>    have just translated the .m file, I have not checked if is correct
>    mathematically with the respect to the original paper.
>    4.
>
>    So I changed a bit the workflow, since we have the ability to
>    regenerate the basis set and is fairly fast (e.g. 1min 30sec for a n=1000
>    and nbf=500 basis), as illustrated in the Readme. So there in a fist step
>    we load or generate the basis, and then calculate the inverse Abel
>    transform. This is done so we can for instance, run the transform in a loop
>    on a list of different files, without having to reload the basis set all
>    the time. I also updated all the examples under examples/
>    5.
>
>    Other minor changes include for instance using a matrix view of numpy
>    arrays, with X.view(np.matrix), so we can do simply X*Y, instead of
>    X.dot(Y) and X.I instead of np.inv(X), so the formulas are more
>    readable.
>    6.
>
>    Added the ability to take in 1D rawdata, mostly for testing.
>
> Please let me know if this works for you, and sorry if there are lot of
> things changing at the same time which is difficult to follow. Some
> functions I updated might need to be more documented,
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/DanHickstein/pyBASEX/pull/3
> Commit Summary
>
>    - Transforming BASEX to a standalone package added setup.py
>    - Rewrote linear algebra operations with np.matrix, and added unit
>    tests
>    - Added the generation of a basis set in python
>    - Optimized the basis set generation and added tests
>    - Adapted examples to the new workflow
>    - Updated the readm
>    - Added support for 1D input arrays
>
> File Changes
>
>    - *A* .gitignore
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-0> (8)
>    - *D* BASEX.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-1> (335)
>    - *D* BASEX.pyc
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-2> (0)
>    - *A* BASEX/__init__.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-3> (12)
>    - *A* BASEX/basis.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-4> (108)
>    - *A* BASEX/core.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-5> (349)
>    - *A* BASEX/data/ascii/basis100_1.bst
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-6> (101)
>    - *A* BASEX/data/ascii/basis100pr_1.bst
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-7> (101)
>    - *A* BASEX/data/ascii/basis40_1.bst
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-8> (41)
>    - *A* BASEX/data/ascii/basis40pr_1.bst
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-9> (41)
>    - *R* BASEX/data/basex_basis_1000_500_orig.npy
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-10> (0)
>    - *A* BASEX/io.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-11> (55)
>    - *A* BASEX/tests/__init__.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-12> (7)
>    - *A* BASEX/tests/test_basis_set.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-13> (61)
>    - *M* README.md
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-14> (48)
>    - *D* additional_files/basis1000.npy
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-15> (0)
>    - *D* basis1000.npy
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-16> (0)
>    - *D* example.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-17> (17)
>    - *D* example_data/Xenon_800_nm_Abel_transform.png
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-18> (0)
>    - *D* example_data/Xenon_800_nm_comparison.pdf
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-19> (0)
>    - *D* example_data/Xenon_800_nm_speeds.txt
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-20> (501)
>    - *R* examples/data/Xenon_800_nm.raw
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-21> (0)
>    - *R* examples/data/Xenon_800_nm.tif
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-22> (0)
>    - *A* examples/example.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-23> (24)
>    - *A* examples/example_main.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-24> (93)
>    - *A* setup.py
>    <https://github.com/DanHickstein/pyBASEX/pull/3/files#diff-25> (14)
>
> Patch Links:
>
>    - https://github.com/DanHickstein/pyBASEX/pull/3.patch
>    - https://github.com/DanHickstein/pyBASEX/pull/3.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/DanHickstein/pyBASEX/pull/3>.
>



-- 
Daniel D. Hickstein
DanHickstein@gmail.com
Phone: 301 787 1853

rth:
Hi Dan,

thanks for your reply. 

About the basis set generation, actually could you wait a couple more days before accepting it? I need to do some more tests, as it is the output seems to be actually significatly different from the original version, which could be due to the basis set routine. At the moment I'm trying to regenerate the 1000x1000 basis set with matlab routines but it's taking really long: it's been running for 10 hours now, and it only  got to n=300 out of 500 . I should be done by tomorrow, and I'll let  you know if there is a problem there or not.

Thanks, 
Best,

DanHickstein:
Oh yeah, those basis set generating programs take forever. "BASIS1.m" only
takes a few minutes, but I've never seen "BASIS2.m" actually finish. I have
not idea how the 1000x1000 basis sets that were included were actually
generated. I've got BASIS2.m running on our cluster right now - we will see
if it ever finishes :)



On Sat, May 2, 2015 at 12:17 PM, Roman Yurchak <notifications@github.com>
wrote:

> Hi Dan,
>
> thanks for your reply.
>
> About the basis set generation, actually could you wait a couple more days
> before accepting it? I need to do some more tests, as it is the output
> seems to be actually significatly different from the original version,
> which could be due to the basis set routine. At the moment I'm trying to
> regenerate the 1000x1000 basis set with matlab routines but it's taking
> really long: it's been running for 10 hours now, and it only got to n=300
> out of 500 . I should be done by tomorrow, and I'll let you know if there
> is a problem there or not.
>
> Thanks,
> Best,
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/DanHickstein/pyBASEX/pull/3#issuecomment-98383933>.
>



-- 
Daniel D. Hickstein
DanHickstein@gmail.com
Phone: 301 787 1853

DanHickstein:
Hi Roman,

Here's are the basis sets calculated using BASIS2.m (as well as BASIS1.m)
as calculated on our Linux cluster:

https://www.dropbox.com/sh/thhhc6hx5bpz52e/AABDpASoWezPa-u7uSjbCUzra?dl=0

Thanks,

Dan


On Sat, May 2, 2015 at 5:05 PM, Daniel Hickstein <danhickstein@gmail.com>
wrote:

> Oh yeah, those basis set generating programs take forever. "BASIS1.m" only
> takes a few minutes, but I've never seen "BASIS2.m" actually finish. I have
> not idea how the 1000x1000 basis sets that were included were actually
> generated. I've got BASIS2.m running on our cluster right now - we will see
> if it ever finishes :)
>
>
>
> On Sat, May 2, 2015 at 12:17 PM, Roman Yurchak <notifications@github.com>
> wrote:
>
>> Hi Dan,
>>
>> thanks for your reply.
>>
>> About the basis set generation, actually could you wait a couple more
>> days before accepting it? I need to do some more tests, as it is the output
>> seems to be actually significatly different from the original version,
>> which could be due to the basis set routine. At the moment I'm trying to
>> regenerate the 1000x1000 basis set with matlab routines but it's taking
>> really long: it's been running for 10 hours now, and it only got to n=300
>> out of 500 . I should be done by tomorrow, and I'll let you know if there
>> is a problem there or not.
>>
>> Thanks,
>> Best,
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/DanHickstein/pyBASEX/pull/3#issuecomment-98383933>.
>>
>
>
>
> --
> Daniel D. Hickstein
> DanHickstein@gmail.com
> Phone: 301 787 1853
>



-- 
Daniel D. Hickstein
DanHickstein@gmail.com
Phone: 301 787 1853

rth:
Hi Dan,

thanks for generating these basis sets, that is very helpull.  So validated that  `BASEX/basis.py` does generate the same output as `BASIS2.m` to machine precision. I also completes in 9 sec instead of a day or so because of the optimizations (simplified the inner loops) . 

However, it turns out `BASIS2.m` was not the script that I should have used, since the inverse Abel does not make sens (unless I made a error somewhere else).   I can probably get to the bottom of it eventually, but it would help if it was possible to ask the person who wrote those scripts what is the difference between these scripts for basis generation, and which is the correct one. Since, both the output you send me of the `BASIS1.m`  and  `BASIS2.m`  don't seem to match the basis set orginally included.  

Thanks,

DanHickstein:
Oh dear. None of them match??

This is the problem with sorting through some code that was whipped up many
years ago. At some point we may just need to start over and code the
algorithm straight from the rev sci instum paper.

Anyway, I discovered that in the original matlab implementation
(additional_file/original_matlab_implementation.zip), there is another
basis generation program (BASIS.m). This one appears to be even less
efficient than the others, but I have it running on the cluster and will
send you the basis set that it generates.

But, maybe it's sufficient to compare the basis sets generated with a
smaller number of basis functions from basis.py with BASIS.m.

Thanks,

Dan



On Mon, May 4, 2015 at 4:46 PM, Roman Yurchak <notifications@github.com>
wrote:

> Hi Dan,
>
> thanks for generating these basis sets, that is very helpull. So validated
> that BASEX/basis.py does generate the same output as BASIS2.m to machine
> precision. I also completes in 9 sec instead of a day or so because of the
> optimizations (simplified the inner loops) .
>
> However, it turns out BASIS2.m was not the script that I should have
> used, since the inverse Abel does not make sens (unless I made a error
> somewhere else). I can probably get to the bottom of it eventually, but it
> would help if it was possible to ask the person who wrote those scripts
> what is the difference between these scripts for basis generation, and
> which is the correct one. Since, both the output you send me of the
> BASIS1.m and BASIS2.m don't seem to match the basis set orginally
> included.
>
> Thanks,
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/DanHickstein/pyBASEX/pull/3#issuecomment-98877701>.
>



-- 
Daniel D. Hickstein
DanHickstein@gmail.com
Phone: 301 787 1853

DanHickstein:
Hi Roman,

I put the basis set calculated using the "BASIS.m" program in the dropbox:

https://dl.dropboxusercontent.com/u/3816350/Basex%20basis%20sets/basis_generating_matlab_programs/dan_basis0_basis1000_1.bst
https://dl.dropboxusercontent.com/u/3816350/Basex%20basis%20sets/basis_generating_matlab_programs/dan_basis0_basis1000pr_1.bst

Compared to the basis sets that I was using, I found the following levels
of agreement:

For the "basis1000" files:
BASIS.m   agrees to within 1e-8
BASIS1.m agrees to within 9e-323 (perfect)
BASIS2.m agrees to within 9e-323 (perfect)

For the "basis1000pr" files:
BASIS.m   agrees to within 1e-6
BASIS1.m agrees to within 1e-7
BASIS2.m agrees to within 42 (horrible)

So, it's not entirely clear how the basis sets that I was using were
generated. It's possible that different versions of matlab are enough
numerically different to generate basis sets that are different to 1e-7 and
these basis sets were generated using the BASIS1.py provided. Since
BASIS2.py agrees perfectly for the "basis1000" file, but not for the
"basis1000_pr" file, it's clear that the math is slightly different. It's
possible that these are just different basis sets. In the C++ version of
basex (I will send this to you for reference), there is an option to use
bot a "narrow" and "wide" basis set, which I believe refers to the width of
the gaussian functions used in the transform. But in this case I would
expect both files to be different.

Looking at the actual transforms from the different basis set (see
screenshot) shows that all of the basis sets produce things that look
reasonable. The transforms using the basis sets generated by BASIS1.m are
almost exactly identical except for a few tiny differences in a handful of
pixels. Similar for BASIS.m, but the differences are slightly larger.

The basis sets from BASIS2.m look strikingly different. The image has much
more electron density at higher radii and much less at lower radii. This
image actually looks much better to me in some ways. I think that we should
prepare a test image based on a function that we can analytically take the
inverse abel transform. We should see which of these basis sets is the
closest to the analytical solution.

Thanks,

Dan


[image: Inline image 1]


On Mon, May 4, 2015 at 5:27 PM, Daniel Hickstein <danhickstein@gmail.com>
wrote:

> Oh dear. None of them match??
>
> This is the problem with sorting through some code that was whipped up
> many years ago. At some point we may just need to start over and code the
> algorithm straight from the rev sci instum paper.
>
> Anyway, I discovered that in the original matlab implementation
> (additional_file/original_matlab_implementation.zip), there is another
> basis generation program (BASIS.m). This one appears to be even less
> efficient than the others, but I have it running on the cluster and will
> send you the basis set that it generates.
>
> But, maybe it's sufficient to compare the basis sets generated with a
> smaller number of basis functions from basis.py with BASIS.m.
>
> Thanks,
>
> Dan
>
>
>
> On Mon, May 4, 2015 at 4:46 PM, Roman Yurchak <notifications@github.com>
> wrote:
>
>> Hi Dan,
>>
>> thanks for generating these basis sets, that is very helpull. So
>> validated that BASEX/basis.py does generate the same output as BASIS2.m
>> to machine precision. I also completes in 9 sec instead of a day or so
>> because of the optimizations (simplified the inner loops) .
>>
>> However, it turns out BASIS2.m was not the script that I should have
>> used, since the inverse Abel does not make sens (unless I made a error
>> somewhere else). I can probably get to the bottom of it eventually, but it
>> would help if it was possible to ask the person who wrote those scripts
>> what is the difference between these scripts for basis generation, and
>> which is the correct one. Since, both the output you send me of the
>> BASIS1.m and BASIS2.m don't seem to match the basis set orginally
>> included.
>>
>> Thanks,
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/DanHickstein/pyBASEX/pull/3#issuecomment-98877701>.
>>
>
>
>
> --
> Daniel D. Hickstein
> DanHickstein@gmail.com
> Phone: 301 787 1853
>



-- 
Daniel D. Hickstein
DanHickstein@gmail.com
Phone: 301 787 1853

rth:
Hi Dan,

thanks for looking into this. I arrived at the same conclusions with respect to the different `BASIS*.m` files.

I have not looked at `BASIS.m`, but as to the other two I belive the algorithm is the same, there is just one parameter that changes which explains the difference in results. Essentially, each element of  `basis1000pr` is calculated as a finite series (the third loop in the `.m` files).  The number of elements for the third loop is given by this   [line](https://github.com/luli/pyBASEX/blob/master/BASEX/basis.py#L78). Now in `BASIS1.m`, the number of elements in this third loop is set arbitrairly to 4000 (it's a bit more complicated then that in reality, it's a number which starts at 4000 and is incremented with the index of the first loop), which is presumably sufficient for the series to be convergent (as far as I can tell from the code, I have not cheched the article in detail).
In  `BASIS2.m`,  the algorithm is less efficient, and because it was so slow already, I guess, it was decided to decrease the number of elements in this third loop from 4000 to 100 (again arbitrairly), which does therefore result in a different output of this script.

In conclusion, as fars as I can tell `BASIS1.m`, is the correct script, and as you indicated it is consistent with the basis originally included. Now, the small error of `1e-6`, could be due to the fact that in earlier version a slightly different value was used instead of 4000  to calculate the series.



Now, I have validated the python script to be consistent with  `BASIS1.m`, to a `1e-7` relative precision (unit tests under `BASEX/test/test_basis_set.py`) 

There are currently two ways to intialse the `BASEX` class. Either with,
       
        inv_ab = BASEX(n=1001, nbf=500, basis_dir='./' )
which will then compute the  basis (a minute or so with this last version) and save it for further use in the `basis_dir`.

Or providing directly the malab generated basis sets (optionally gzipped), as following

      inv_ab = BASEX(use_basis_set="../BASEX/data/ascii/original_basis1000{}_1.txt.gz",)
 which is constant with previous version, and can be used to check the rest of the code undependetly of the basis set generation.

I have run the a 1D test with the gaussian function (see `example/example_gaussian.py`), and apart from the absolute scaling factor, the  result looks correct when compared to the analytical solution. 

So I thing the current version, should be ready to merge. In any case, it's possible to use the module  independently of the basis set generation function. I do concur that checking the code with respect to the algorithm would be helpfull, at some point...  Also making some authomatic tests comparing with an analytical solution would be nice, but that requires to find out why this absolute scaling factor is there.

Best,

Roman

DanHickstein:
Hi Roman,

This all sounds good. Have you tried generating different sized basis sets?
When I decrease the number of basis functions, the transform looks tiny,
when I think that the preferred behavior would be for it to just get lower
resolution.

More importantly, I tried to generate a 500x500 basis set with 250 basis
functions, and then there is an error with the data centering routine. Can
you try this?

Thanks,

Dan


On Thu, May 7, 2015 at 1:30 PM, Roman Yurchak <notifications@github.com>
wrote:

> Hi Dan,
>
> thanks for looking into this. I arrived at the same conclusions with
> respect to the different BASIS*.m files.
>
> I have not looked at BASIS.m, but as to the other two I belive the
> algorithm is the same, there is just one parameter that changes which
> explains the difference in results. Essentially, each element of
> basis1000pr is calculated as a finite series (the third loop in the .m
> files). The number of elements for the third loop is given by this line
> <https://github.com/luli/pyBASEX/blob/master/BASEX/basis.py#L78>. Now in
> BASIS1.m, the number of elements in this third loop is set arbitrairly to
> 4000 (it's a bit more complicated then that in reality, it's a number which
> starts at 4000 and is incremented with the index of the first loop), which
> is presumably sufficient for the series to be convergent (as far as I can
> tell from the code, I have not cheched the article in detail).
> In BASIS2.m, the algorithm is less efficient, and because it was so slow
> already, I guess, it was decided to decrease the number of elements in this
> third loop from 4000 to 100 (again arbitrairly), which does therefore
> result in a different output of this script.
>
> In conclusion, as fars as I can tell BASIS1.m, is the correct script, and
> as you indicated it is consistent with the basis originally included. Now,
> the small error of 1e-6, could be due to the fact that in earlier version
> a slightly different value was used instead of 4000 to calculate the series.
>
> Now, I have validated the python script to be consistent with BASIS1.m,
> to a 1e-7 relative precision (unit tests under
> BASEX/test/test_basis_set.py)
>
> There are currently two ways to intialse the BASEX class. Either with,
>
>     inv_ab = BASEX(n=1001, nbf=500, basis_dir='./' )
>
> which will then compute the basis (a minute or so with this last version)
> and save it for further use in the basis_dir.
>
> Or providing directly the malab generated basis sets (optionally gzipped),
> as following
>
>   inv_ab = BASEX(use_basis_set="../BASEX/data/ascii/original_basis1000{}_1.txt.gz",)
>
> which is constant with previous version, and can be used to check the rest
> of the code undependetly of the basis set generation.
>
> I have run the a 1D test with the gaussian function (see
> example/example_gaussian.py), and apart from the absolute scaling factor,
> the result looks correct when compared to the analytical solution.
>
> So I thing the current version, should be ready to merge. In any case,
> it's possible to use the module independently of the basis set generation
> function. I do concur that checking the code with respect to the algorithm
> would be helpfull, at some point... Also making some authomatic tests
> comparing with an analytical solution would be nice, but that requires to
> find out why this absolute scaling factor is there.
>
> Best,
>
> Roman
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/DanHickstein/pyBASEX/pull/3#issuecomment-99989627>.
>



-- 
Daniel D. Hickstein
DanHickstein@gmail.com
Phone: 301 787 1853

rth:
Hmm, yes I have not tested this that much in other configurations. I do get the same errors as you...

The fact that, the transforms looks tiny when the number of basis functions is decreased is extremely strange, and I do not understand it. I would also have expected a lower resulution too... 

As to the other error, yes that happens if the size of the data is larger then than `n`. The problem is in the centering function, puts the data into an  array of size `2n` by `2n`, but then if the data size is larger then `n`, it's going to get outsize of the array boundaries. Although, this issue is not really specific to this pull request, I have mery adapted the function that was already there. For instance, if you take input data of size 2000 x 2000 in the previous version (v3) with a 1000 basis set, you will get   the same error. I could look into this, though probably not this week.

DanHickstein:
I probably should have made this comment when I merged this. But, for the record I'll say that your diagnosis of the problem is completely correct. I fixed the centering function so that it should be more robust in the future. I merged this pull request and everything seems to be working well. Thanks for all of the improvements!!

Issue #9, opened 2015-11-17
Code clean up


rth:
- make sure all functions are documented
  - improve code styling  (PEP 0008)
  - possibly simplify the Readme

rth:
Closing this issue, as this is not so relevant anymore. The Readme still would need to be updated as other implementations are added but it's a separate issue.

PR #10, opened 2015-11-17
Pep code style improvements

rth:
A few minor changes to improve the style of the code ([PEP 0008](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements)), related to issue #9 :
 
   - renamed the module name to lowercase. So on github it is still called pyBASEX but when installed it is used as,
  
        from  basex import BASEX
      also updated all the examples.

   - added a few more white spaces for readability
   - changed version format from 0.5 to 0.5.0 (usual convention) and added it to the `__init__.py`


DanHickstein:
These look great! 

Regarding the name-change (lowercase is fine), you've got me thinking - pybasex is actually already a python package. I'll start a separate issue regarding that...

DanHickstein:
Hmmm, so, I think that we need to make the folder lowercase as well. right?

rth:
So it is lowercase in the reposotory. However, I think on local computer you would now have 2 identical folders  `basex` and a `BASEX` , after this PR, where the later is not referenced in git anymore and can be safely deleted. Well that's what I had on my computer anyway, not sure why it happens...

DanHickstein:
Ah, okay, works now.

PR #15, opened 2015-11-17
Adding continuous integrations with Travis CI

rth:
This PR adds the `.travis.yml` setup as well as simplifies the mechanism for running tests. 
The result of the Travis build is avalable at https://travis-ci.org/PyAbel/PyAbel
Unit tests can be run either from inside the module with,

    nose -s abel/tests/

or if the module is installed, from any location with,
   
    python -c 'import abel.tests; abel.tests.run_cli(coverage=True)'

This latter part is used in Travis CI, and also show tests coverage of the code base (to use it locally one might have to install `pip install coveralls`). 

For now, only the test for generating the basis set is run.

P.S: sorry this PR is a bit messy as there are parts related to other issues: there were just to many simultaneous changes in the repo today ))


DanHickstein:
It passed the tests! Looking good! :+1:

Issue #8, opened 2015-11-16
Update tests?


DanHickstein:
@rth : test_basis_set.py does not really do anything since the relevant basis sets are no longer included. Should we update the tests so that they are actually a meaningful test of BASEX? 

Two tests we could use:
1) Include the transformed image with the pyBASEX and compare the user-generated image with the image that we generated. 

2) Compare the BASEX transform with the analytical inverse abel of a gaussian (within the scaling factor at this point).

rth:
I agree that  updating tests would be a good idea, including both tests proposed above. I will add the second one. As to the first one, could you look into it (I'm not sure which user-generated image you have in mind) ? The only thing, I suppose, would be to take a low resolution test image, so  a) it does not take too much space in the repo b) tests run sufficiently fast.

As to testing the basis set generation, I would still keep it, as this allows to test this functionality independently of the rest (particularly checking compatibility of different python versions etc).  Are you sure about "the relevant basis sets are no longer included. " ? I still see basis sets under `BASEX/data/ascii/`.  Do we need those? As far as I am concerned I would just remove them and, 
   - let the basis be generated the first time the user runs the calculation (this way it is resolution independent).
   - I can also add a small (e.g. 10x10)  sample basis file under `BASEX/tests/data` for unit tests

What do you think?

DanHickstein:
Oh, I didn't see that we still had those basis sets there. I believe that those basis sets are identical, so it would make sense to get rid of one set. And I suppose that the users are not going to care if this implementation of BASEX matches some old matlab version. If anything, they will care if it matches the C implementation of BASEX. That version only runs on Windows so I can't do it right now, but perhaps I should prepare a sample image and transform it using the Basex.exe and we can use this for one of our tests, since it is considered to be the gold standard. (Also, I'm curious if they worked out the issue with the scaling factor.)

Also, I like your idea of saving a small 'reference' basis set in the tests folder for comparison purposes. 

I think that we should include a basis basis set or two with the program. If someone is on an older computer, the generation can take a while, and I want to give the impression that this implementation is fast (it is!). So, perhaps we should include a 500x500 or 1000x1000 basis set with the example files.

rth:
So just to sum up the previous conversation and commits:
  - Was added: unit tests for BASEX with a symmetric step
  - Addressed in a separate issue #19 : adding unit tests for a gaussian function (this therefore can be ignored here)
  - remains to do to close this issue: 
[...]

PR #16, opened 2015-11-18
Added unit tests for BASEX with a symmetric step 

rth:
As discussed in issue #8, instead of a Gaussian function, I added unit tests for a symmetric step function that also have an analytical abel transform, but can be more easily compared as the reconstruction should just be flat.

This symmetric step function and its Abel transform are calculated in `abel.analytical.SymStep`. I also added an example under `examples/example_step.py` that produces the figure below:

figure_1
Then under `abel.benchmark.SymStepBenchmark` we take the ratio of the reconstructed step (here for `6<x<14`) and the original signal. The mean of this ratio gives the scaling factor (which should be 1, if the scaling is correct), while the standard deviation should be close to 0.0 (meaning that the reconstruction is flat). I added a unit test to check that BASEX produces a correct reconstruction up to a scaling factor.

This PR also
 -  removes unnecessary basis sets `abel/data/ascii/*gz`. 
 - There is already a 500x500 basis set included under `examples/` that users with a slow computer could use. I'm not sure about including the 1000x1000 basis set, as it is 16 MB which will significantly increase the size of the repo...
 - Also added the ability not to save basis sets on disk if the `basis_dir` argument to the `BASEX` class is set to `None` (mostly for unit tests).


DanHickstein:
Looks great! I'd say you can go ahead and merge this one.

Issue #4, opened 2015-05-02
Absolute scaling of the inverse abel transform with BASEX


rth:
Hi Dan, 

I'm interested in the absoute scaling of the output of the BASEX algorithm: so when we process some physical signal, I'm trying to make sure that we to obtaint the correctly scaled inverse abel transform.

This uses the pyBASEX v0.3 (i.e. before the pull request). 

For instance, if we take a  gaussian signal `f(r) = exp(-r**2)` the forward Abel tranform is given analytically by  `F(r) = pi**0.5*f(r)`, so if we were to apply the BASEX algorithm to `F(r)` function it should yield exactly `f(r)`. Now when I'm doing this I obtain the plot below,

![figure_1](https://cloud.githubusercontent.com/assets/630936/7442144/72e5b9ee-f104-11e4-8e35-ab3e4f2039a1.png)

The code to make this figure is given in the following [gist](https://gist.github.com/rth/416280215353196c97b1) (and it uses pyBASEX v0.3, before the pull request)

So it looks really good, but for  a constant  scaling factor of approximatily `8.86` . I already scaled the BASEX output by `1/dx`  where `dx = 0.02` is my pixel size  which should be the only necessary scaling I belive.  So I'm a bit confused about about the `8.86` scaling factor. I suppose it should be some combination of `dx`, `N=1001`, and maybe `pi` to some power, but up to now I was not able to find which one. Would you have any ideas about it?

Thanks,

DanHickstein:
Dear Roman,

I have notice this discrepancy before, but I have not tried to figure out
where it is coming from. It would be great if the BASEX reconstruction
could match up with the analytical inverse abel. I think that the same
problem exists with the compiled C version of the BASEX software that most
people use.

Hopefully you can figure out what is going on, otherwise it's probably a
good idea to just include a scaling factor (as you suggest) to get it to
match up.

Thanks,

Dan


On Sat, May 2, 2015 at 12:10 PM, Roman Yurchak <notifications@github.com>
wrote:

> Hi Dan,
>
> I'm interested in the absoute scaling of the output of the BASEX
> algorithm: so when we process some physical signal, I'm trying to make sure
> that we to obtaint the correctly scaled inverse abel transform.
>
> This uses the pyBASEX v0.3 (i.e. before the pull request).
>
> For instance, if we take a gaussian signal f(r) = exp(-r**2) the forward
> Abel tranform is given analytically by F(r) = pi**0.5*f(r), so if we were
[...]

PR #17, opened 2015-11-18
Added a scaling factor to BASEX

rth:
Following the issue #4 , this PR adds a fixed scaling factor of `8.053` on the output of the BASEX algorithm as to match the analytical results.

Here, I used the symmetric step function example (see PR #16 )  that depends on a number of parameters, including:
 - `n` : number of points along the r axis
 - `r_max`: range of the symmetric r interval
 - `r1`, `r2`: bounds of the step function for (r > 0)
 - A0:  height of the step

So the test code I used is accessible here https://gist.github.com/rth/9885931342bf8a5ad056 ( requires PR #16). Essentially, it calculates the ratio between the height of the reconstructed signal to the original one (e.i. the scaling factor). Then sweeps along different input parameters are performed in order to determine which ones influence the scaling factor. 

The original result is here:  http://pastebin.com/EWsRVhMB  where we want to have `ratio_mean == 1` if the scaling is correct. 

So by normalizing the output by `8.053/pixel_size` we do obtain the correct scaling in all the tested cases: http://pastebin.com/S6FtmR4m (up to a precision given by `ratio_std` as there is some noise in the reconstructed signal). 

@DhrubajyotiDas I have also tried the [scaling value you used](https://github.com/DhrubajyotiDas/pyroBASEX/blob/master/BASEX/core.py#L160) of `1.1122` (note that you scaled the `Ci` matrix, while I scale the output signal). So this would be equivalent of scaling the output signal by  `1.1122*n` I think? In any case I was not able to get good agreement for all tested parameters with this value (see results at http://pastebin.com/1Zj5EwAS )

Anyway, this is a bit weird, at some point somebody would need to look up what factor is missing in the math, but for now I guess that's all we can do...

### DIFF
--- a/abel/core.py
+++ b/abel/core.py
@@ -51,9 +51,9 @@ from .io import parse_matlab
 
 class BASEX(object):
 
-    def __init__(self, n=501, nbf=250, basis_dir='./', use_basis_set=None,
-                    calc_speeds=False, pixel_size=1.0, correct_scaling=True,
-                    verbose=True):
+    def __init__(self, n=501, nbf=250, basis_dir='./',
+                    use_basis_set=None, verbose=True, calc_speeds=False,
+                    pixel_size=1.0, scaling_correction=True):
         """ Initalize the BASEX class, preloading or generating the basis set.
 
         Parameters:
@@ -70,7 +70,7 @@ class BASEX(object):
                   Gzip compressed text files are accepted.
           - calc_speeds: determines if the speed distribution should be calculated
           - pixel_size: size of one pixel in the radial direction (optional)
-          - correct_scaling: correct the result by a heuristic scaling factor
+          - scaling_correction: correct the result by a heuristic scaling factor
                             as to be consistent with the analytical inverse Abel transform.
                             This requires to set pixel_size to the correct value.
           - verbose: Set to True to see more output for debugging
@@ -80,7 +80,7 @@ class BASEX(object):
 
         self.verbose = verbose
         self.calc_speeds = calc_speeds
-        self.correct_scaling = correct_scaling
+        self.scaling_correction = scaling_correction
         self.pixel_size = pixel_size
 
         self.n = n
@@ -173,7 +173,8 @@ class BASEX(object):
 
     def _get_scaling_factor(self):
 
-        return 1./self.pixel_size
+        MAGIC_NUMBER = 8.053   # see https://github.com/PyAbel/PyAbel/issues/4
+        return MAGIC_NUMBER/self.pixel_size
 
 
 
@@ -241,7 +242,7 @@ class BASEX(object):
         if self.ndim == 1:
             recon = recon[0, :] # taking one row, since they are all the same anyway
 
-        if self.correct_scaling:
+        if self.scaling_correction:
             recon *= self._get_scaling_factor()
 
         if self.calc_speeds:


DanHickstein:
Yeah, I think that this is the best thing that we can do at this point. I suggest that we go ahead and merge this.

It's a little weird that nobody noticed this difference in absolute scaling in the past. But I think that a similar problem exists with the BASEX.exe that most photoelectron spectroscopy researchers use. I will play with this more today when I can use the lab computer where BASEX is installed, but here is an example of some of some of my data transformed using the BASEX.exe. The magnitude of the projection and the transform should roughly match, but instead I needed to multiply the transform by ~200. So it's way off. 

image

Also, same issue with the [matlab POP algorithm](http://www.mathworks.com/matlabcentral/fileexchange/41064-polar-onion-peeling/content/POP.m). Here I needed to multiply the transform by 40 to obtain agreement. 

image

So, a nice feature of the BASEX algorithm in PyAbel is that gives you the physically meaningful absolute value, even if it's with an empirically derived scaling factor.

rth:
Hmm, when you say "the magnitude of the projection and the transform should roughly match" shouldn't it somehow be related to the `pixel_size` or the number of points `n`? Or is it compensated for?

Say if you take a signal with `n` points and perform BASEX calculations which will yield some amplitude. Now if you up sample the original signal by a factor of 2 (i.e. same signal on `2n` points with half the original `pixel_size`) the amplitude of the BASEX calculation will change (which I account for in this PR). Do you observe that in the BASEX.exe program or is it just an offset by some constant factor ?

DanHickstein:
You are basically saying that if you change the bin size of a histogram, then the peak heights are going to change. That is certainly true and something we should watch out for. (And, potentially the cause of the mysterious 8.053 offset could be something like this, where we are making some sort of coordinate transform and not taking into account the Jacobian.)

But, in the two images that I posted, the number of pixels is preserved, so I think that this should not be a problem. But you are correct, I should say that the magnitude of the projection and the inverse abel transform should look about the same *so long as the pixel size is preserved.*

DhrubajyotiDas:
@rth I've noticed that the BASEX output oscillates severely at near-vertical edges of the transform, i.e., at `|x| = 6` or `14` in the example function in https://github.com/PyAbel/PyAbel/pull/16

Why not stick with gaussian example functions where the outputs are smooth everywhere except at `x = 0`? I believe if you omit the very edges of the range of `x`, the element-wise ratio between the analytical transform and the BASEX transform is very close to single-valued, and a mean or median of that should be sufficient to see if the scaling factor is accurate or not.

As another data point to this discussion, here's my results with the `1.1122` multiplied to the `Ci` matrix for an exponential

image

and a few other polynomial functions (from doi:10.1016/j.sab.2005.11.009)

image

DanHickstein:
Thanks @DhrubajyotiDas, that's a good point that we should be careful to ignore the edge regions when calculating the scaling factor. I think that @rth took this into account by only considering the middle 60% of each step, so we should be okay. But yes, just using a function that produces a smooth transform would be a good solution. 

If I just multiply the ``Ci`` matrix by ``1.1122`` in the PyAbel master branch, I cannot obtain agreement. Did you produce the graphs using the code here or in your pyroBASEX repo?

rth:
@DhrubajyotiDas  Good point, I'll add a issue to add tests with Gaussian functions (that I initially changed in part because of this artifact at `x=0` in the reconstruction).

I was going to make the same comment as @DanHickstein :  would you have the script to produce that Gaussian function + reconstruction? I tried to run the `examples/example_gaussian.py` with pyroBASEX but could make it work fully because of the change in the interface. I'm wondering if the correction factor you get might be different because of the modification of the basis set generation? Or it is possible I'm doing something wrong. In any case, it would probably be good to understand this difference in normalization  before merging the PR.. Thanks.

DhrubajyotiDas:
I produced the graphs using the version of BASEX in my pryoBASEX repo. Here's the script to reproduce the Gaussian line-of-sight function reconstruction. 

https://gist.github.com/DhrubajyotiDas/bd68a205c94a340f336e

[Here's a link](https://dl.dropboxusercontent.com/u/6335217/basex_basis_21_201_20_101.npy) to the basis set that was generated and used in the example. If you are trying to reproduce the example, it might be useful to compare your generated basis set to the one I'm using if you notice disagreements.

The `1.1122` scaling factor is already incorporated here, so the BASEX reconstruction directly matches the analytical reconstruction. You can comment out the [relevant line](https://github.com/DhrubajyotiDas/pyroBASEX/blob/master/BASEX/core.py#L160) and retry to see the *native* disagreement.

Some side notes: `n_vert` and `nbf_vert` in my example are not set to `1` because I was investigating if those had any impact on the scaling factor (they didn't).

rth:
@DhrubajyotiDas  Great, thanks! I just checked that both PyAbel and pyroBASEX have the correct scaling   ( https://gist.github.com/rth/d2454b787d71ffe0ef5d ) : one with the 8.053 factor and the other with `Ci*1.1122` included in the code. So because the latter approach does not seem to work  with PyAbel, it stands to reason that this scaling factor was somehow changed when you modified the basis generation function. 

Well, it is empirical anyway, so as long as it works it should be fine. I'm merging this PR. Thanks for help in debugging this issue!

PR #21, opened 2015-11-19
Added a list of dependencies

rth:
Added a list of dependencies 
    ` "numpy >= 1.6"`,   ` "setuptools >= 16.0"`,   ` "scipy >= 0.14"`  
which should not be too constraining and we can lower the version numbers if needed. 

Checked that 

     pip install git+https://github.com/PyAbel/PyAbel.git

works and the current tests pass on Linux and Windows (with conda). Same applies to `python setup.py install`  Pip will also try to install missing dependencies although in the case of numpy/scipy this might not be very desirable (compiling from sources is long).


DanHickstein:
Looks good, thanks!

Issue #18, opened 2015-11-19
Allow for wider basis functions: investigate why nbf != n/2 +1 doesn't produce the expected result


DanHickstein:
In most cases, it makes sense to use a number of basis functions so that there is one basis function per pixel. Currently ``nbf`` defaults to ``250`` and we do not mention that ``n`` and ``nbf`` should be changed together. This behavior could be made more obvious by making ``nbf`` default to ``'auto'``, in which case ``nbf`` will be set to ``(n-1)/2``. 

Actually, now that I actually try to use ``nbf != (n-1)/2``, the behavior is not at all what I expected. As @DhrubajyotiDas mentioned in https://github.com/PyAbel/PyAbel/issues/14, a smaller number of basis functions should provide larger functions and better smoothing of the data. Instead, changing the number of basis functions simply provides a basis set that only transforms the center of the image.

``n=501, nbf=250:``
![image](https://cloud.githubusercontent.com/assets/1107796/11282671/cd0968d8-8ebd-11e5-8a41-a8873acb63ab.png)

``n=501, nbf=100``
![image](https://cloud.githubusercontent.com/assets/1107796/11282668/c7df1920-8ebd-11e5-9a9f-c41b83b1429f.png)

rth:
Aww, yes that is a good point. I'm not sure at all that the basis set generation routine is correct when `nbf != (n-1)/2` (n odd)  as we have only tested it against an example basis set that has `2*nbf +1 = n`. I mean it is identical to the matlab implementation, but I don't know if that one was correct in this case. For now, maybe we can just set `nbf='default'`  (and set `nbf = n//2` (integer division) in the code)   as you mentioned?

DanHickstein:
Yes, we should have ``nbf='auto'`` and set it to ``n//2``. We should probably include a warning if the user tried to set nbf to anything different until we figure out what is going one with the basis set generation.

DhrubajyotiDas:
Just noticed something: Typically when `n` is an odd number, say, `5`, then there's `5//2+1 = 3` "unique" points (`r = 0, +/-1, +/-2`) that need to be resolved with the basis functions `nbf`. Currently the `nbf = auto` option will set it to `5//2 = 2` basis functions, when it should default to `5//2 + 1 = 3` basis functions.

DanHickstein:
So, the ``n`` value actually works strangely as well. In the docstring, we say:
```
Abel inverse transform will be performed on a `n x n` area of the image
```
[...]

Issue #14, opened 2015-11-17
Discussion on the general user interface in PyAbel


DanHickstein:
In order to simplify and leave room for other transform methods, should we combine the basis set generation routines (basis.py) with the main basex code (core.py)? Some of the general image processing functions that are currently living in core.py (generate_speeds and all functions below that) should probably go into their own file since they will likely be common to many of the transform methods. 

Additionally, would it be better to allow the basex transform to just be one function call? Currently, there is one call to make/load the basis sets and another function to perform the transform. This might confuse most users, who just want to call a single function to perform a basex transform on their image. Why not just have one function that takes the image and does the transform?

rth:
For the first part, I agree completely, well we can move all related part from `basis.py` and `core.py` to `basex.py` and more the rest under e.g. `utils.py`, `postprocessing.py` (for the `generate_speeds`), etc. 

As to the second point, I agree that we can merge loading of basis set and the BASEX calculation in one function call (although personally I find it more intuitive in 2), but there is also a number of other steps to consider that should be common to all inverse Abel transforms:

  1. Pre-processing (optional):
      - Possibly filtering: median, etc.
      - Centering the image
  2. Inverse Abel transform
     - in the case of BASEX in 1 step or 2 steps
  3. Post-filtering:
     - Cleaning up the reconstruction with Gaussian, median filters, etc.
  4. Post-processing depending on the application:
    - calculating speeds, etc.

So right now all of these steps are performed in 2 function calls, and I don't think it would help to merge them in 1 function call . I mean we can do a wrapper if somebody find that convenient, but in the work-flow it could be useful to separate these different steps as function calls.  I mean that's probably what would be done in Matlab as well. 

 No, what  do you think?

Say for basex what do you think about something roughly as follows,

     from abel.basex import BASEX

     # img is the input image
     b = BASEX(img, n=500, nbf=250, basis_dir='./', verbose=True)
     b.pre_process(center=(681,491), median_size=2, gaussian_blur=0)
     res = b.run()
     res2 = b.post_process(median_size=2) # optional
     IM = b.calculate_speeds() # or some other calculation depending on the application
     

where `calculate_speeds`, `pre_process`, and `post_process` would be common to all implementations. For a different implementation what would change is the first 2 lines:
   
     from abel.some_other_implementation import OnionPeeling
      
     b = OnionPeeling(img, some_other_parmaters)
     # the remaining steps are same as above

In addition, we could also create a wrapper independent of the implementation something like:
[...]

PR #22, opened 2015-11-20
Major code refactoring

rth:
Following discussion in issue #14 , this PR performs a major reorganization of the code base to make the use of other inv. Abel backends more simple. This includes lots of internal changes (moving code around, breaking apart internal functions to get more modular operations, renaming variables, etc) however the idea was to keep the public user interface unchanged, and it can be accessed with `from abel.basex import BASEX` same as before. This high level function/class can then be changed at a later point as discussed in issue #14. The aim of this PR was mostly to clean up the internal functions (also related issue #9 ). 

Detailed changes include:
  - added a default argument `nbf='auto'` and printing a warning when `nbf!=n//2` as discussed in issue #18 
  - merged `core.py` with `basis.py` into `basex.py`, while other functions were moved to `tools.py`
  - removing some input argument of `BASEX` as to simplify its use:
      - completely removed `use_basis_sets` argument: this was used to load basis sets from matlab files mostly for debugging. This is still possible using dedicated function in `abel.io`, but most users won't care about it.
      - renamed `pixel_size` argument to `dr` 
  - exposed 3 internal  elementary functions related to BASEX that are fully independent:
      - `basex_generate_basis_sets`, `basex_get_basis_sets_cached`, `basex_transform` (the names are a bit long, might need to be simplified)
  -  Extended a bit the mechanism for defining the analytical functions under `analytical.py`
  - simplified and generalized benchmarks  under `benchmark.py`
  - adapted the unit tests to this new setup. Also added a few superficial unit tests to increase code coverage.
  - Fixed broken examples
  - The module version is now defined in `abel/_version.py` and is loaded automatically in `setup.py`
  - Changed the file format for saving the basis sets. It now includes the version number of PyAbel with which it was generated. So for instance, if say in release 0.7 we detect  that there was a bug in the basis set generation routines and fix it, we can print warning in future releases when such basis sets are loaded and ask users to recalculate them with the last version of the software. This is backward compatible, so files generated with `<=PyAbel-0.5` can still be loaded as before.

Please review these changes and let me know if you have comments, or if not you can merge the pull request. Because a lot of code was moved between different files the diff is a bit confusing, so it might also be useful to check directly this version at https://github.com/rth/PyAbel/tree/code_refactoring_v1


DanHickstein:
This looks good. The functions will all be prefixed with ``abel.basex.``, so I would think that ``basex_get_basis_sets_cached`` could be shortened to ``get_basis_sets_cached``, and similarly for the other functions. Otherwise this looks good as far as I can tell!

rth:
@DanHickstein  Yes, I was not sure about that. In part I would tend to do the same thing. But following the same logic `basex_transform` under `abel.basex` should be renamed in simply `transform` which is not ideal when it is imported as 
    
    from abel.basex import transform
as we kind of loose information about the fact that it is related to basex. But you are right, that is still probably more clear with respect to he basis set generation functions, as those names are too long.

DanHickstein:
Actually, it's fine with me to use  ``from abel.basex import transform``. I don't think that the function name itself needs to give you all the information. The package and module name provide the rest. 

Eventually, as we discussed before, we just need to have two different functions that are basically
```python
abel.basex.center_and_transform_and_calulate_radial_distribution()
```
and
```
abel.basex.transform_only()
```
but with cooler sounding names. We were going to call these ``BASEX`` and ``basex_transform``, but if we are getting rid of the ``basex`` prefix, then we might need to be more creative.

Issue #20, opened 2015-11-19
Reduce the size on disk of PyAbel


rth:
Right now when downloaded PyAbel repository takes 95 MB wich is quite a lot for such package. Downloading the PyAbel.zip from github takes 42 MB on disk when unzipped.

This means that there is like 50MB of huge files (probably basis sets) in the git history that are not used anymore and can be safely  removed (need to look up how to do that). 

Out of the remaining 42 MB, the distribution is as follows:
   - `examples/` -> 13 MB  including
       - `basex_basis_501_250.npy` -> OK
       - `examples/data`  -> do we need both the `Xenon_800_nm.raw` and `Xenon_800_nm.tiff` ? Can we not remove the `.raw` ? or is it useful? 
   - `additional_files/` -> 30 MB
        - `basis1000_1.txt` and `basis1000pr_1.txt`:  -> 2x7.7MB = 15MB do we still need to include these? Can we not just put them somewhere and provide a download link for the few people who want to play with the .m files? Or at least gzip them. 
        - `original_matlab_implementation.zip`: -> 14MB This is useful, but it  includes several versions of `basis1000*.txt` files. Not sure if that it is a probem, or we can keep it as it is now.

DanHickstein:
Good questions. For the old Matlab code and basis sets, I think that it would be a great idea to just archive these somewhere and provide a link. There is no reason most users would be interested in them. But, I think that it's nice to keep them in an obvious place where developers of PyAbel can go back and look at them in case they are interested. It was certainly interesting for us when we were playing with the basis set generation functions. 

Can we archive them on GitHub somehow while still keeping them out of the main repo? Perhaps just fork what we have now to a PyAbelLegacy repo and then delete them here?

And fine with me to use the tiff instead of the raw file. Currently the example loads the raw, so we just need to switch this.

We could also crop/downsample the Xenon sample image to 500x500.

rth:
So, just for the record, as this was discussed in other issues/PR; the corresponding files were removed in PR #28, PR #29 and the git history cleaned in issue #32. The additional files were moved to a new PyAbel/PyAbelLegacy repo.

PR #28, opened 2015-11-21
Code refactoring v2

rth:
This PR concludes the code reorganization as discussed in issue #14 . Changes include:
  - merges `BASEX.__init__` and  `BASEX._call__` into a single function, and removed all the pre/post filtering with numpy (which can be added back later if necessary). The name BASEX is not consistent with Python naming conventions for a function, but we can address that later.
  - regenerated `examples/basex_basis_501_250.npy` in the new format (PR #22 )
  - removed everything under `additional_files/` that was moved to a separate repo `PyAbel/PyAbelLegacy`, as discussed in issue #20 . This does not remove those files from the git history though, for now. 
  - adapted tests and examples to the new `BASEX` function.

If you are ok with this, you can merge this PR.


DanHickstein:
This looks good, and I will go ahead and merge. 

For the name, do you mean that we should change uppercase ``BASEX`` to lowercase ``basex``? This is fine with me. 

How does the git history work? Anyone who wants to clone the repo is stuck downloading extra information that they don't need?

PR #29, opened 2015-11-21
Modified the example to use a smaller image

DanHickstein:
Also added better comments the example and generally polished it.


rth:
Looks good. I'll delete the different removed files from the git history tomorrow.

Issue #25, opened 2015-11-21
Add comprehensive documentation


rth:
Some mechanism for including documentation with sphinx for the PyAbel module should be added.

One possibility is to use put in a separate repository on e.g. `PyAbel.github.io` . Alternatively it is possible to put it in the same repository and use `https://htmlpreview.github.io/?link_to_the_html_files.html`  as it is done for the [pyinstaller documentation](https://htmlpreview.github.io/?https://github.com/pyinstaller/pyinstaller/blob/develop/doc/Manual.html).  Not sure what are the drawbacks/advantages of each method.

DanHickstein:
Yes, I was thinking that it would be great to add some nice documentation. Also not sure the best way to do this. The pyinstaller page that you linked looks useful but a little bit ugly. Maybe we could include a brief introduction and overview information (with some nice images) on the PyAbel.github.io page, and then link to the sphinx generated documentation...

rth:
I agree that would be nice. I'm not so keen on the the Pyinstaller docs myself; I can work on the sphinx generated documentation if you want. As to the overview page, I'm not sure what would be the best layout/template for it  nor where it can be hosted, but it sound like a good idea.!

DanHickstein:
I think that there are two important aspects to the documentation. One is that we should have a way to get the docstrings nicely formatted on a webpage, as is the standard for Numpy, Scipy, and other Python modules. We don't really know how to do this yet, but the docstrings that are in the code are more-or-less okay.

A second important aspect is to provide additional information about the transforms, such as background information, advantages of the transform, relevant equations, deviations from papers (i.e., we correct typos), and general user advice. I think that the wiki provided by GitHhub is a quick and easy place to provide this information, using that standard Markdown syntax that we are accustomed to. So, I would say that we should start including the PyAbel documentation using the wiki: https://github.com/PyAbel/PyAbel/wiki

Each transform method should have it's own page, and there can also be a page for general information about abel transforms.

What does the team think about this? 

@rth @DhrubajyotiDas @stggh

stggh:
Looks good. 

Perhaps, with links for  `BASEX`  etc in the home page summary, to the relevant wiki, as per the right-side 'pages' navigation box, for easy access, and for those  that do not notice the right-side navigation box.

DhrubajyotiDas:
Good idea @stggh. I've added links to the individual sections from the home page summary and fleshed out the `three_point` section.

DanHickstein:
[...]

PR #31, opened 2015-11-23
A few small tweaks to the readme

DanHickstein:
Also, I removed “#!/usr/bin/python” from the top of abel/test/run.py
since it doesn’t seem necessary.


rth:
Looks great. For the way functions parameters are formatted this will probably depend on documentation (issue #25 ) so sphynx can pick it up correctly, and generate function definitions.

Issue #19, opened 2015-11-19
Add polynomial functions with known Abel solutions


rth:
It would be nice to add unit tests for Gaussian functions as mentioned in PR #17 . An example of the implementation could be (but we could do something else too):
- add a definition for a general Gausian under `analytical.py` and its Abel transform (e.g. similarly to `abel.analytical.SimStep` )
- add a benchmark under `benchmarks.py`  that checks e.g. that the reconstruction should be identical to the original signal up to a scaling factor (excluding origin and  |x | >> sigma)  or some other criteria
- finaly add a unit test checking that BASEX implementation passes this benchmark

At some later point, adding  a few other polynomial functions from doi:10.1016/j.sab.2005.11.009 as mentioned by  @DhrubajyotiDas would be great too.

DhrubajyotiDas:
I've coded up the polynomial function checks so it'd be easier to incorporate them into the standard unit tests. It can be found here: https://gist.github.com/DhrubajyotiDas/f8f531ca2cfa3fee6345

**Note**: The original citation `doi:10.1016/j.sab.2005.11.009` has a few typographical errors, which I had to manually correct. In fact `Case:4` in that reference is not included in the code I pasted above because I couldn't figure out where the error was.

![image](https://cloud.githubusercontent.com/assets/6530840/11288857/4ab99ece-8ef5-11e5-91e4-92910e39a191.png)

rth:
That is nice!

rth:
Just for future reference, at some point, we could also add solutions from the Hansen & Law paper, coded by @stggh is [this comment](https://github.com/PyAbel/PyAbel/pull/41#issuecomment-163433698).

DhrubajyotiDas:
@rth I believe the function in [this comment](https://github.com/PyAbel/PyAbel/pull/41#issuecomment-163433698) "Curve A" is the same as Case 3 in [my linked notebook](https://gist.github.com/DhrubajyotiDas/f8f531ca2cfa3fee6345).

rth:
@DhrubajyotiDas You are right, sorry, I missed that.

DanHickstein:
This issue should be addressed by #201, and can probably be closed once that is merged.

PR #33, opened 2015-11-24
For nbf='auto', set nbf=n//2 + 1   and added gaussian function

rth:
As @DanHickstein  rightly noted in issue #18 , for an image with a radial size `n`, there should be `nbf=n//2 + 1` basis functions, not `nbf=n//2` as it is now. Sorry about that. 

This PR sets `nbf=n//2 + 1`, when `nbf='auto'`. Interestingly this also changes the scaling completely, and required scaling is now `1.1122` (instead of  `8.053` before), which is identical to the value @DhrubajyotiDas found. So I guess this makes sens, but something we should take into account, and if somebody would want to check the math at some later point to understand what's going on, it would be more than welcome. Right now hard coding `nbf=n//2 + 1`, and outputting lots of warnings if that's not the case seems to be the only thing we can do.


rth:
Actually while we are at it, I also added an analytical Gaussian function in this PR, as discussed in issue #19 . So if we close all issues related to Version 0.6, we can maybe do  [a release on github v0.6](https://github.com/PyAbel/PyAbel/releases) that has the last working version with the BASEX implementation only, and then  keep working on adding other implementations in v0.7?

DanHickstein:
This would fix the last issue for v0.6, right? 

See my other comment about the transform being performed on an ``(n-1)x(n-1)`` area instead of ``nxn``. If we resolve this, then I think that v0.6 is ready for release.

rth:
Yes, let me just open one last issue for bundling basis sets, as I realize now that the basis set shipped under `examples/` would be for `nbf = n//2`, which is not used anymore...

DanHickstein:
You know, in the interest of keeping the repo smaller, I think that it would be fine to remove the basis sets and just generate them on the fly. As long as we print that statement to the user that it won't take a long time everything, I think they will understand.

rth:
I just added a another unit test to check that the shape of the BASEX result is correct.

I agree that if we include basis sets files, then realize there is a bug, add new ones, etc.. it is going to be an issue for the repo size. So I removed the `examples/basex_basis_501_250.npy` file, and we can also indicate that they can download basis sets e.g. from 
   https://github.com/PyAbel/PyAbelLegacy/tree/master/basis_sets/BASEX
as there we can just delete the repo periodically and read clean files again. Ideally would be just to have some way of hosting static files outside of a repo, but I don't think github allows it.

If you are ok with this, you can merge it.

DanHickstein:
Looks good to me! And I doubt that generating the basis sets will be that problematic for the users. The 500x500 is pretty quick even on my laptop.

Issue #34, opened 2015-11-25
recon image has shape (width+1,height+1)


stggh:
```python
$ ipython
Python 2.7.10 (default, Oct 14 2015, 16:09:02) 

IPython 2.3.0 -- An enhanced Interactive Python.

In [1]: import numpy as np

In [2]: from abel.basex import BASEX

In [3]: im = np.loadtxt("O-ANU2048.txt")

In [4]: (n,m) = np.shape(im)

In [5]: print (n,m)
(2048, 2048)

In [6]: recon,speed = BASEX(im,center=(int(n/2),int(m/2)),n=n,basis_dir='./',verbose=True,calc_speeds=True)
Loading basis sets...           
0.72 seconds
Reconstructing image...         
16.19 seconds
Generating speed distribution...
2.07 seconds

In [7]: print (np.shape(recon))
(2049, 2049)
```

PS: Great work guys!

DhrubajyotiDas:
Could you retry with an image that has width = odd integer?

rth:
Hi Steve,

thanks for pointing this out. 

As far as I understand, the current BASEX implementation only supports `n` odd, and [will raise an error if that is not the case](https://github.com/PyAbel/PyAbel/blob/master/abel/basex.py#L272). When given an image with an odd `n`, the BASEX transform produces a result of the same shape, as checked by [this test](https://github.com/PyAbel/PyAbel/blob/master/abel/tests/test_basex.py#L47). Now, what must be happening here is that the centering function pads the original image to make it odd, which result in a transform of shape (2049, 2049). 

What I'm wondering about is there actually a physical reason why we can't generate basis sets for n=even?  If so, we probably should ensure that the centering function produces the result of the shape (`n x n`) and raise an error if it is odd in the `BASEX` function and not just in the `generate_basis_sets` routine. What do you think about it ?

stggh:
Thanks for the very fast response. The odd size image is a good point. I have trimmed the image by one pixel and the new basis calculation in progress.

Unfortunately, I am unfamiliar with the basex algorithm and do not know any reason for the odd pixel image size. 

My group uses the Hansen and Law J. Opt. Soc. Am A2, 510-520 (1985). I have written a version that could be included, as an alternative to the basex method (and onion peel). I am just waiting on approval for the upload from my colleague. The agreement is excellent apart from a small displacement in the radial position of the speed profile.
[...]

PR #35, opened 2015-11-26
HansenLaw - another inverse Abel transformation algorithm

stggh:
Hansen and Law  J. Opt. Soc. Am A2, 510-520 (1985) recursive inverse Abel transform

Note, tweaks for your own coding style may be required.

Code:  abel/HansenLaw.py
Example:   example/example_HansenLaw.py      using data/O-ANU2048.txt.bz2

Fuller documentation to be added later.

Probably requires an even-number pixel image(!). Not tested on the other example images.


rth:
This is great! Thanks a lot for contributing! 

We should fix the issue #34  with the centering function for even integers, release a stable version 0.6 with a finalized  BASEX implementation (hopefully within a days or two) and then we can merge this under the development version 0.7 and keep working on adding other implementations, and making it all consistent.

DanHickstein:
This is awesome! Thank you so much Steve! 

@rth is right, we should eat our vegetables, fix the issue with centering, and release version 0.6. *Then* we can merge this cool new code! 

I had a look at HansenLaw.py, and I had a few thoughts:

Line 13: No need to compare this method to basex, it's a quality method that stands on its own. Perhaps provide some description of what are the key advantages of this method (and mention that it's a recursive method).

Line 21: Can you provide a link to the paper and (especially) the PhD thesis?

Line 34: Can you provide more comments in the HansenLaw_transform function? If possible, it would be really useful for future hackers if you could mention which sections correspond to which equations in the article (if this makes sense).

The speed of the transform is probably hampered by the fact that this algorithm has three nested loops. One over ImgRow (size=2048), one over N (size=2048), and K (size=9?). If we could vectorize one or more of these, we might see a big speedup. For example, what stops us from running the algorithm on the whole image at once instead of row-by-row?

If we really can't eliminate the nested loops, we could try something like [numba](http://numba.pydata.org/) to speed them up.

Line 56: I like the idea of the mask, since it will be very useful for users with imperfect datasets (i.e., damage in one quarter of their detector). I am a little confused about passing a single number to define the mask. What about just using a tuple, like ``mask=(0,1,0,1)``? Or, ``quadrants=(False,True,False,True)``. That might be more intuitive for most users. Or, even a string: ``mask='0101'``?

Line 95: typo in comment.

The dataset you supplied is simply incredible. I work with multiphoton ionization of neutrals, so I have never seen such perfect rings. In the interest of keeping the repo small, can we downsample to 1024x1024? 

I think that it still looks pretty nice:

image

Again, thank you so much for contributing this! I want to use this right away in my own research :)

stggh:
Thanks, for all the encouragement. 

@Dan - I agree with all of your excellent recommendations and will make the changes. The whole image transformation in parallel is a great idea. For the example code the image size may be 1024x1024 or smaller. In the end it only affects the ultimate resolution of the O- photoelectron spectral transitions:  FWHM(1024)=1.18 meV, 2048=0.98 meV, 3072=0.95 meV. My spectrometer does produce high quality velocity-map images, camera 2048x2048 pixel. In the measurement electron events are centroided, to sub-pixel accuracy, and so there is no (accumulated) background image noise.  The image is further circularized, to reduce any residual image distortion. Circularization is achieved through comparing adjacent angular slice radial intensity profiles, and then applying a radial scaling correction to each electron coordinate (x/sf,y/sf). This is an essential step to obtain high-resolution spectra.

Image circularization should be on our todo list. I will explain the details another time. My current algorithm operates on electron coordinates, rather than pixel intensities, so some adaptation, or an alternative method, will be required.

Also, todo, anisotropy parameters, for which the old basex.exe analysis was poor (or it may have been the user).

stggh:
I have pushed another code update: 
Hansen and Law: updated code according to @DanHickstein pull request 1 conversation, including parallel processing, tuple quandrant selection, single reform image block. Example reduced image size 1024x, changed to O2- as more strcuture. Renamed functions according to issue 10 (consistent naming scheme) - I think!

Parallel processing results in a speed up ~x3 (using 3 processor cores).

DanHickstein:
This look good Steve! Thanks for making those changes!

Is it possible to just pass the transform function the whole image instead of just the row? The ``for n in range(N-1)`` loop could just iterate over columns rather than pixels. Is there a reason that doesn't work?

stggh:
Dan, 
Transforming the whole image is the job of iabel_hansenlaw. 

The hansenlaw transform (in iabel_hansenlaw_transform) starts at the outer edge of the image and recurses to the centre.  It makes no sense to hand it the whole image. This would also not fit well with the parallel processing of each row, that produces the massive speed up.

Sorry, if I have missed the point of your suggestion.

DanHickstein:
I just looked at your new example image - it's even more fantastic! 

For the anisotropy parameters, do you simple extract those from the transformed image (like we do with calc_speeds), or do you generate them directly from the transform somehow? It would be great to add a function to the tools.py to calculate anisotropy. I think that will be a feature that a lot of photoelectron/photoion people will be looking for. 

Also, the image circularization sounds very exciting. I have not attempted this in my research, but it's something that's also very important in many cases.

Do you want to open separate issues for these?

DanHickstein:
So, should we merge this now that we are in version 0.7?

stggh:
Yes please.  I will open issues for the other items.

rth:
Yeah, that would be great. @stggh you just need to [rebase it](https://github.com/rth/PyAbel/blob/contribution_guidelines/CONTRIBUTING.md#before-merging) on the current master, as this fork cannot be automatically merged at present (I just rebased my other fork that was at a similar point in the history without any issues, so it does work :) ).

stggh:
I receive an error:
```
git fetch upstream
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

rth:
@stggh Aww, in your `.git/config` in the upstream section you can replace `git@github.com:PyAbel/PyAbel.git` by  `https://github.com/PyAbel/PyAbel.git` to use https instead of ssh (well same format as what is used in your `[remote "origin"]` section. I will update the documentation file to take this into account.

rth:
Great, let's merge this!

DanHickstein:
When I changed calc-speeds, I broke the Hansen-Law function. Fix is in this PR: https://github.com/PyAbel/PyAbel/pull/43

PR #37, opened 2015-11-29
Fixed centering function

DanHickstein:
We will need to think more about centering, but in the short run, this PR fixes the bug where the center was shifted by one pixel for odd-sized imaged. The behavior is also correct for even-sized images (using the convention discussed in https://github.com/PyAbel/PyAbel/issues/34 if we return ``im[:-1,:-1]`` (Currently we are returning an odd-sized image even if we get an even sized image in. This seems good for what we currently know about basex, but in the long run we should change it to allow both even and odd images.)

Here is some code I used for testing:
```python
import numpy as np
import abel

# # Odd:

n=5
data = np.zeros((n,n))

data[2,2]=1


# # # Even:
# n=4
# data = np.zeros((n,n))
#
# data[2,2]=1
# data[2,1]=1
# data[1,1]=1
# data[1,2]=1

center = (2,2)

fig = plt.figure(figsize=(12,6))
ax1 = plt.subplot(121)
ax2 = plt.subplot(122)

d2 = abel.tools.center_image(data,center=center,n=n)

ax1.imshow(data,interpolation='nearest',origin='lower')
ax2.imshow(d2,  interpolation='nearest',origin='lower')

ax1.set_title('Raw data')
ax2.set_title('Center=(%i,%i)'%(center[0],center[1]))

plt.show()
```

image

image

### DIFF
--- a/abel/tools.py
+++ b/abel/tools.py
@@ -78,24 +78,20 @@ def center_image(data, center, n, ndim=2):
     elif ndim == 2:
         cx, cy = np.asarray(center, dtype='int')
         
-        #make an array of zeros that is large enough for cropping or padding:
+        # Make an array of zeros that is large enough for cropping or padding:
         sz = 2*np.round(n + np.max((Nw, Nh)))
         im = np.zeros((sz, sz))
+        
+        # Set center of "zeros image" to be the data
         im[sz//2-cy:sz//2-cy+Nh, sz//2-cx:sz//2-cx+Nw] = data
         
-        im = im[ sz//2-n_2:n_2+sz//2+1, sz//2-n_2:n_2+sz//2+1]
+        # Crop padded image to size n 
+        # note the n%2 which return the appropriate image size for both 
+        # odd and even images
+        im = im[ sz//2-n_2:n_2+sz//2+n%2, sz//2-n_2:n_2+sz//2+n%2]
         
-        #print(np.shape(im))
     else:
         raise ValueError
-
-    if n%2==0: # if n is odd
-        return im[:-1,:-1]
-        
-    else:      # if n is even
-        return im
-
-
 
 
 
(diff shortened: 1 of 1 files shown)


rth:
This looks good. 

I propose to first merge the PR #38, that adds some unit tests on how the centering function should behave. Then close and open this PR (to restart the Travis build) to see if it passes the added tests. I suspect there is another bug in the centering function (offset by 1 for the center for both odd and even images), which should be detected by unit tests. Another commit to fix that might be necessary.

DanHickstein:
Was that result with this new version of the centering function? When I make a 9x9 array with a center at (5,5), I can successfully center and crop to 3x3....

rth:
@DanHickstein No that was with the original version. We should wait for the Travis build to complete to see what we have with this one...

DanHickstein:
Oh, well, it's failing because it doesn't produce even-sized images...

rth:
So before we had 4 test failing, this PR fixed one (centering of odd images), and now we have only 3. See [travis log](https://travis-ci.org/PyAbel/PyAbel/jobs/94184887)  Well if I did not make mistakes in the unit tests in the first place.  So we can either merge this and fix the remaining 3 in a different PR, or keep working in this one, but then you would be the only one who can commit. It's up to you @DanHickstein .

DanHickstein:
Yup, I removed the last row and column and it passes now.

rth:
Great. An alternative solution, could be to replace 
  
       im = im[ sz//2-n_2:n_2+sz//2+1, sz//2-n_2:n_2+sz//2+1]
       + removing of last row

with 

       im = im[ sz//2-n_2:n_2+sz//2+n%2, sz//2-n_2:n_2+sz//2+n%2]
not sure which one would be more readable / simple to understand?

DanHickstein:
I like the one-liner solution.

Issue #30, opened 2015-11-23
Incorporate polar onion peeling algorithm into PyAbel


DanHickstein:
It would be nice to implement one of the inverse Abel transform methods that uses a polar basis set to allow for smooth transforms of noisy data. [Polar onion peeling (POP)](http://dx.doi.org/10.1063/1.3126527) is one of the most popular methods for this. 

There is a nice Matlab implementation of the POP algorithm [here](http://www.mathworks.com/matlabcentral/fileexchange/41064-polar-onion-peeling). Adi Natan, the author of that code, said that it would be fine if we ported it to python and incorporate it in to PyAbel. He also has the matlab code that generates the basis sets, so we would not be limited to the 1024x1024 basis set that comes with the program.

rth:
That would be great! It would be interesting  to see how different implementations behave with noisy data.

DanHickstein:
Now that #270 has added rBasex, there is probably little motivation to also add pBasex.

MikhailRyazanov:
Now I need to say here that POP is not really correct. Namely, this procedure:
![image](https://user-images.githubusercontent.com/12854133/79825006-07877c80-8355-11ea-92ac-b3e819449238.png)
that generates the projection of an anisotropic ring is valid only if the ring is infinitely narrow (that is, _b_<sub>_r_</sub>(_R_) → Abel[δ(_R_ − _r_)]). In reality, for extended basis functions with different angular dependences, their projections have different radial parts, not just by the (_R_/_r_)<sup>_n_</sup> factor.

Here is a comparison of the actual projections and their POP approximations for the triangular basis functions used in rBasex (the POP article is not clear about the shape of their basis functions, but they are unlikely to be narrower than this):
![POP](https://user-images.githubusercontent.com/12854133/79825885-7d8ce300-8357-11ea-820c-1a1e02ec3837.png)
_p_<sub>6; 0</sub>(_R_) here is what they call _b_<sub>6</sub>(_R_) — projection of the isotropic basis function centered at _r_ = 6, and _p_<sub>6; _n_</sub>(_R_) are projections of _anisotropic_ basis functions with cos<sup>_n_</sup> θ angular dependences.
As can be seen, all POP approximations have noticeable errors, larger for higher angular orders.

[...]

Issue #27, opened 2015-11-21
Enable use of up/down asymmetric images


DhrubajyotiDas:
To summarize some of the discussion which already took place in https://github.com/PyAbel/PyAbel/issues/14, currently PyAbel presupposes the `raw_data` being transformed has up/down symmetry, i.e., `f(r,-z) = f(r,z)`. This is mostly because in photoelectron spectroscopy, images are theoretically up/down symmetric, and so a lot of reconstruction algorithms make this approximation in order to gain better signal-to-noise. 

The underlying BASEX algorithm does not require the `raw_data` being transformed to have up/down asymmetry. The only requisite symmetry is the lateral `f(-r) = f(r)` condition, which is common to all Abel inversion algorithms. This symmetry condition is a feature of many experimental data sets, such as images of coflow and counterflow flames, or even in photoelectron spectroscopy images, where hot spots on the detector might break up/down symmetry. Therefore it would be useful to enhance PyAbel to perform in the more general no-up-down-symmetry situation.

My fork of PyAbel, [PyroBASEX](https://github.com/DhrubajyotiDas/pyroBASEX) has already made the changes to the basis-set generation and implementation codes to account for this situation. The overall changes can be summrized as: `n` has been decomposed into `n_horz` and `n_vert`, and the relevant `nbf_horz` and `nbf_vert` basis functions are generated for each direction independently. The basis functions in the vertical direction do not need to be forward-transformed, and therefore their calculation is very quick. 

As far as incorporating the code changes into PyAbel goes, most of it ought to be straightforward. To preserve code legibility in PyAbel, it might be useful to have two separate internal basis-set generation routines. From the end-user's perspective, a `vertical_symmetry=<True, False, or unspecified>` will need to be added to whatever function call is being made to do the transform. Additionally, the format of the specification of the `center` of the raw data will need to be updated to handle a vector in addition to a tuple. 

One user-interface concept is the one discussed in https://github.com/PyAbel/PyAbel/issues/14

* `vertical_symmetry=True` -> Use PyAbel code
* `vertical_symmetry=False` or `unspecified` -> Use pyroBASEX code

Examples

1)  `n=500` -> Use pyroBASEX, `n_horz, n_vert = 500,500`
2)  `n=500, vertical_symmetry=True` -> Use PyAbel as usual
3)  `n=(500,500), vertical_symmetry=True` -> Same as above, convert `n` from tuple to integer `n=500`
4)  `n=(500,500)` -> Same as 1)
[...]

Issue #24, opened 2015-11-21
Add Fourier-Hankel implementation for direct and inverse Abel transforms


rth:
Probably won't be too hard to implement at some point, if somebody is interested.  See [FHA cycle](https://en.wikipedia.org/wiki/Hankel_transform#Relation_to_the_Fourier_and_Abel_transforms) 

The only thing that need to be cheked is whether the Hankel transform can be done exclusively with `numpy`, `scipy` (don't know if `scipy.linalg.hankel` is related) or whether we would need to use a separate package  [`hankel`](https://pypi.python.org/pypi/hankel/0.1.0). 

In which case it can be added as an optional dependency, and this implementation should raise an error if `hankel` module is not installed (trying to keep the number of additional dependencies low). Or any other way you prefer to handle this.

DanHickstein:
Yes, it would be fun to include the FH version of the transform for comparison purposes! 

The ``scipy.linalg.hankel`` says that it constructs a "Hankel Matrix". Wikipedia says that the [Hankel Transform](https://en.wikipedia.org/wiki/Hankel_transform) is "Not to be confused with the determinant of the [Hankel matrix](https://en.wikipedia.org/wiki/Hankel_matrix) of a sequence."

So, I think that we may need to use that package that you suggest, or figure something else out.

DanHickstein:
This should be pretty easy to implement, but, when I looked around the literature, I could not find a clear equation for the Fourier-Hankel transform, despite it being mentioned frequently. @stggh, do you know of a paper that gives the equation/algorithm for the FH transform?

stggh:
Python code for the Hankel transform is available at this [github](https://github.com/steven-murray/hankel). 

The HankelTransform function requires a function (image) input rather than (grid) data. At the moment I am confused by the evaluation of the Hankel transform in radial (1D) space, and how to return the 2D image, although, the (slightly buggy) examples appear to cover this case.

Thoughts, suggestions?

DanHickstein:
Oh, that is pretty cool that you found code for the Hankel transform. It looks like that code is just for the 1D case, right? We would just need to loop over all rows (or, hopefully, vectorize somehow). 

It looks like that package was developed reasonably recently, so the author might be able to help us wrangle it into a Fourier-Hankel transform function.

stggh:
[...]

Issue #23, opened 2015-11-21
`hedp.math.abel` direct and inverse Abel implementation


rth:
Move  the  [hedp.math.abel](https://github.com/luli/hedp/blob/master/hedp/math/abel.py) implementation  to PyAbel .  Still thinking what names would be appropriate.

DanHickstein:
Yes, this is a great idea and very useful to have the forward Abel transform for testing thing out. Since this is a "direct" numerical implementation of the analytical formula, should it live in ``abel.direct.???``

DanHickstein:
I'm excited to see this implemented in PyAbel. How does it compare to the more sophisticated methods? Just way more noise along the center?

DanHickstein:
I think that we can close this issue, since the direct implementation is now included. I made a separate issue (https://github.com/PyAbel/PyAbel/issues/58) to look into the difference between the Python and Cython results.

PR #41, opened 2015-12-01
Discussion on the contribution guidelines and unit tests

rth:
This Pull Request aims to discuss the contribution guidelines for adding new implementation in v0.7. This is relevant for issues #23, #24 , #27 , #30, and PR #35 .

It adds a [`CONTRIBUTING.md`](https://github.com/rth/PyAbel/blob/contribution_guidelines/CONTRIBUTING.md) which you can view with the previous link.

The general formulation and English might need to be revised, the main point of this PR is that I think that before merging an new fabel/iabel implementation to the code base it should pass some generic unit tests to check that it behaves as expected. 

Since the original developer is the person best suited to understand how his implementation work, I think it is better to detect inconsistencies with the general way PyAbel works before merging, rather then when the code is included in PyAbel. This could also allow and emphasize basic assumptions we don't really think about (what's the center, odd even images,  etc),. 

What remains is then to agree on some very light and non constraining tests that every implementation should be able to pass. I outline 3 possible tests in the [`CONTRIBUTING.md`](https://github.com/rth/PyAbel/blob/contribution_guidelines/CONTRIBUTING.md#naming-conventions).  What do you think about those? Any other tests that could be useful?

I will add the actual code for unit test to this PR once we converge on what those unit tests should be.

Please let me know what you think about this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ build/
 examples/data/Xenon_800_nm*
 examples/*.pdf
 examples/*.png
-examples/*basis*
+examples/*basex_basis*
 examples/*.dat
+*basex_basis*
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ The implementation named `<implementation>`, located under `abel/<implementation
 
 As to detect issues early and avoid regressions, the submitted implementation should have the following properties and pass the corresponding unit tests,
 
- 1. The reconstruction has the same shape as the original image for the parity it supports. When provided with an image size with a parity it does not support a clear exception should be raised.
+ 1. The reconstruction has the same shape as the original image for the parity (even/odd shape of the image) it supports. When provided with an image size with a parity it does not support a clear exception should be raised.
 
  2. Given an array of 0 elements, the reconstruction should also be a 0 array.
   
@@ -44,11 +44,12 @@ which performs an equivalent call.
 
 ##### Dependencies
 
-The current list of dependencies can be found in [`setup.py`](https://github.com/PyAbel/PyAbel/blob/master/setup.py). Unless it cannot be avoided, refraining from adding new dependencies is preferred. 
+The current list of dependencies can be found in [`setup.py`](https://github.com/PyAbel/PyAbel/blob/master/setup.py). Please refrain from adding new dependencies, unless it cannot be avoided.
+
 
 ##### Before merging
 
-As to keep a clean git history, before merging your pull request please rebase your fork on the last master on PyAbel/. This could be done  [as explained in this post](https://stackoverflow.com/questions/7244321/how-to-update-a-github-forked-repository),
+As to keep a clean git history, before merging your pull request please rebase your fork on the last master on PyAbel. This could be done  [as explained in this post](https://stackoverflow.com/questions/7244321/how-to-update-a-github-forked-repository),
    
     # Add the remote, call it "upstream" (only the fist time)
     git remote add upstream git@github.com:PyAbel/PyAbel.git
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,3 +45,28 @@ which performs an equivalent call.
 ##### Dependencies
 
 The current list of dependencies can be found in [`setup.py`](https://github.com/PyAbel/PyAbel/blob/master/setup.py). Unless it cannot be avoided, refraining from adding new dependencies is preferred. 
+
+##### Before merging
+
+As to keep a clean git history, before merging your pull request please rebase your fork on the last master on PyAbel/. This could be done  [as explained in this post](https://stackoverflow.com/questions/7244321/how-to-update-a-github-forked-repository),
+   
+    # Add the remote, call it "upstream" (only the fist time)
+    git remote add upstream git@github.com:PyAbel/PyAbel.git
+
+    # Fetch all the branches of that remote into remote-tracking branches,
+    # such as upstream/master:
+
+    git fetch upstream
+
+    # Make sure that you're on your master branch 
+    # or any other branch your are working on
+
+    git checkout master  # or your other working branch
+
+    # Rewrite your master branch so that any commits of yours that
+    # aren't already in upstream/master are replayed on top of that
+    # other branch:
+
+    git rebase upstream/master
+
+See [this wiki](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) for more information.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,7 +49,7 @@ The current list of dependencies can be found in [`setup.py`](https://github.com
 
 ##### Before merging
 
-As to keep a clean git history, before merging your pull request please rebase your fork on the last master on PyAbel. This could be done  [as explained in this post](https://stackoverflow.com/questions/7244321/how-to-update-a-github-forked-repository),
+If possible, before merging your pull request please rebase your fork on the last master on PyAbel. This could be done  [as explained in this post](https://stackoverflow.com/questions/7244321/how-to-update-a-github-forked-repository),
    
     # Add the remote, call it "upstream" (only the fist time)
     git remote add upstream git@github.com:PyAbel/PyAbel.git
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,9 @@ As to detect issues early and avoid regressions, the submitted implementation sh
  3. The implementation should be able to calculated the inverse (or forward) transform of a Gaussian function defined by a standard deviation `sigma`, with better than a `10 %` relative error with respect to the analytical solution for `0 > r > 2*sigma`.
 
 
+Unit tests for a given implementation are located under `abel/tests/test_<implemenation>.py`, which should contain at least the following 3 functions `test_<implementation>_shape`, `test_<implementation>_zeros`, `test_<implementation>_gaussian`. See `abel/tests/test_basex.py` for a concrete example.
+
+
 The test suite can be run from within the PyAbel package with,
   
     nose -s  abel/tests/ --verbosity=2  --with-coverage --cover-package=abel
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,4 +70,8 @@ As to keep a clean git history, before merging your pull request please rebase y
 
     git rebase upstream/master
 
+    # push the changes to your fork
+ 
+    git push -f
+
 See [this wiki](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) for more information.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+# Contributing to PyAbel
+
+
+
+
+### Adding a new forward or inverse Abel implementation 
+
+We are always looking for new implementation of forward or inverse Abel transform, therefore if you have an implementation that you would want to contribute to PyAbel, don't hesitate to do so. 
+
+
+In order to allow a consistent user experience between different implementations, and insure an overall code quality, please consider the following points in your pull request.
+
+##### Naming conventions
+
+The implementation named `<implementation>`, located under `abel/<implementation>.py` should use the following naming system for top-level functions,
+
+ -  `fabel_<implemenation>`  :  forward transform (when defined)
+ -  `iabel_<implemenation>` :  inverse implementation (when defined)
+ -  `_bs_<implementation>` :  function that generates  the basis sets (if necessary)
+ -  `bs_<implementation>_cached` : function that loads the basis sets from disk, and generates them if they are not found (if necessary).
+
+
+##### Unit tests
+
+As to detect issues early and avoid regressions, the submitted implementation should have the following properties and pass the corresponding unit tests,
+
+ 1. The reconstruction has the same shape as the original image for the parity it support. When provided with a image size with a parity it does not support a clear exception should be raised).
+
+ 2. Given an array of 0 elements, the reconstruction should also be a 0 array.
+  
+ 3. The implementation should be able to calculated the inverse (or forward) transform of a Gaussian function defined by a standard deviation `sigma`, with better then a `10 %` relative error with respect to the analytical solution for `0 > r > 2*sigma`.
+
+
+The test suite can be run from within the PyAbel package with,
+  
+    nose -s  abel/tests/ --verbosity=2  --with-coverage --cover-package=abel
+
+or, from any folder with,
+    
+   python  -c "import abel.tests; abel.tests.run_cli(coverage=True)"
+
+which performs an equivalent call. 
+  
+
+##### Dependencies
+
+The current list of dependencies can be found in [`setup.py`](https://github.com/PyAbel/PyAbel/blob/master/setup.py). Unless it cannot be avoided, refraining from adding new dependencies is preferred. 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,11 +24,11 @@ The implementation named `<implementation>`, located under `abel/<implementation
 
 As to detect issues early and avoid regressions, the submitted implementation should have the following properties and pass the corresponding unit tests,
 
- 1. The reconstruction has the same shape as the original image for the parity it support. When provided with a image size with a parity it does not support a clear exception should be raised).
+ 1. The reconstruction has the same shape as the original image for the parity it supports. When provided with an image size with a parity it does not support a clear exception should be raised.
 
  2. Given an array of 0 elements, the reconstruction should also be a 0 array.
   
- 3. The implementation should be able to calculated the inverse (or forward) transform of a Gaussian function defined by a standard deviation `sigma`, with better then a `10 %` relative error with respect to the analytical solution for `0 > r > 2*sigma`.
+ 3. The implementation should be able to calculated the inverse (or forward) transform of a Gaussian function defined by a standard deviation `sigma`, with better than a `10 %` relative error with respect to the analytical solution for `0 > r > 2*sigma`.
 
 
 The test suite can be run from within the PyAbel package with,
@@ -37,7 +37,7 @@ The test suite can be run from within the PyAbel package with,
 
 or, from any folder with,
     
-   python  -c "import abel.tests; abel.tests.run_cli(coverage=True)"
+    python  -c "import abel.tests; abel.tests.run_cli(coverage=True)"
 
 which performs an equivalent call. 
   
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # PyAbel
 
 [![Build Status](https://travis-ci.org/PyAbel/PyAbel.svg?branch=master)](https://travis-ci.org/PyAbel/PyAbel)
+[![Appveyor Status](https://ci.appveyor.com/api/projects/status/github/PyAbel/PyAbel?branch=master&svg=true)](https://ci.appveyor.com/project/PyAbel/PyAbel)
+
 
 PyAbel is a Python package for performing Abel and (primarily) inverse Abel transforms. The Abel transform takes a cylindrically symmetric 3D object and finds the 2D projection of that object. The more difficult problem -- the inverse Abel transform -- takes the 2D projection and finds the central slice of the 3D object by assuming cylindrical symmetry in the vertical direction.
 
--- a/README.md
+++ b/README.md
@@ -6,11 +6,11 @@ PyAbel is a Python package for performing Abel and (primarily) inverse Abel tran
 
 The PyAbel package offers several options for completing the inverse Abel transform:
 
-1) The BASEX algorithm creared by Dribinski, Ossadtchi, Mandelshtam, and Reisler [[Rev. Sci. Instrum. 73 2634, (2002)](http://dx.doi.org/10.1063/1.1482156)]. The BASEX implementation uses Gaussian basis functions to find the transform instead of analytically solving the inverse Abel transform.
+ 1. The BASEX algorithm creared by Dribinski, Ossadtchi, Mandelshtam, and Reisler [[Rev. Sci. Instrum. 73 2634, (2002)](http://dx.doi.org/10.1063/1.1482156)]. The BASEX implementation uses Gaussian basis functions to find the transform instead of analytically solving the inverse Abel transform.
 
-2) The "Hansen and Law" recursive method described in [[J. Opt. Soc. Am A 2 (4) 510 (1985)](dx.doi.org/10.1364/JOSAA.2.000510)]
+ 2. The "Hansen and Law" recursive method described in [[J. Opt. Soc. Am A 2 (4) 510 (1985)](dx.doi.org/10.1364/JOSAA.2.000510)]
 
-3) In the future, we hope to have more options for the forward and inverse abel transforms.
+ 3. In the future, we hope to have more options for the forward and inverse abel transforms.
 
 ### Symmetry
 
--- a/abel/analytical.py
+++ b/abel/analytical.py
@@ -36,9 +36,6 @@ class BaseAnalytical(object):
 
         if symmetric:
             self.r = np.linspace(-r_max, r_max, n)
-            if n % 2 == 0:
-                raise ValueError("Error: When using a symmetric interval, the number of discretization points\n\
-                n = {} must be odd".format(n))
         else:
             self.r = np.linspace(0, r_max, n)
 
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -62,5 +62,5 @@ def absolute_ratio_benchmark(analytical, recon):
     """
     mask = analytical.mask_valid
     err = analytical.func[mask]/recon[mask]
-    return np.mean(err), np.std(err), np.sum(mask)
+    return err
 
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -9,6 +9,85 @@ import numpy as np
 
 from .tools import get_image_quadrants
 
+class AbelTiming(object):
+    def __init__(self, n=[201, 401], n_max_bs=500):
+        """
+        Benchmark performance of different iAbel/fAbel implementations.
+
+        Parameters
+        ----------
+          - n: a list of arrays sizes for the benchmark (assuming 2D
+            square arrays (n,n) )
+          - n_max_bs: since the basis sets generation takes a long time,
+            do not run this benchmark for implementations that use basis sets
+            for n > n_max_bs
+        """
+        from .basex import get_basis_sets_cached, basex_transform
+        from .hansenlaw import iabel_hansenlaw
+        from timeit import Timer
+
+        self.n = n
+
+        NREPEAT = 10
+
+        res_fabel = {}
+        res_iabel = {'BASEX':     {'bs': [], 'tr': []},
+               'HansenLaw': {'tr': []}
+               }
+        for ni in n:
+            x = np.random.randn(ni,ni)
+            if ni <= n_max_bs:
+                bs = get_basis_sets_cached(ni, basis_dir=None)
+                res_iabel['BASEX']['bs'].append(
+                    Timer(lambda: get_basis_sets_cached(ni, basis_dir=None)).timeit(number=1))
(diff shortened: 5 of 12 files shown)


DanHickstein:
This looks really good! It's a great idea to make these guidelines clear.

You may want to define "parity (even/odd shape of the image)" when you first mention it. 

Somewhere we should also mention that in case someone cannot meet all of these guidelines with their code, they can simply create a new issue and we would be happy to help integrate their transform.

Also, I would change "Unless it cannot be avoided, refraining from adding new dependencies is preferred." to "Please refrain from adding new dependencies, unless it cannot be avoided."

DanHickstein:
Should we go ahead and merge this so that the contributing.md file is now included? We can always improve it later...

rth:
Well, let me just a few unit tests to this pull request and we can merge it. I'll try to do it later today. 

I'd rather limit the number of PR, as it's easier to look it up if all related commits are in a single place and  besides after a while it takes more effort/time to open a PR, explain it, rebase, discuss/review etc, rather than actually code the functionality.

DanHickstein:
Sounds good!

rth:
Ok, I pushed all  the commits to this PR so this could be merged after review. The additional changes include:
  
  - Simplified  unit tests and benchmarks.
  - added `appveyor.yml` to run unit tests on Windows alongside with Travis-CI (on Linux). @DanHickstein Could you please activate http://www.appveyor.com/ for PyAbel (free for opensource)? This is not absolutely necessary, but is nice to have so everything remains platform independent. 
  - Btw, in examples, I also  replaced unix type path as `data/Xenon_ATI_VMI_800_nm_649x519.tif` with `os.path.join('data', 'Xenon_ATI_VMI_800_nm_649x519.tif')` so they also work on Windows. 
  - Cleaned up and simplified unit tests for BASEX in  `abel/tests/test_hansenlaw.py`
  - Added unit tests for `abel/tests/test_hansenlaw.py`: @stggh in the `test_hansenlaw_gaussian` the ratio between the the numerical reconstruction and the analytical result for a Gaussian is ~0.70-0.75 (i.e. it passes that test only with a 30% relative tolerance). Not sure if I'm doing something wrong, or if there is an issue with absolute scaling. 

Don't hesitate to adapt/extend those unit tests in the future, if you feel that some other things should be tested for.  

On a completely unrelated topic, @DanHickstein I see that you added a basis set under examples  `basex_basis_1025_513.npy` (17 MB) a few days ago. The problem is that for me it errors when I try to use it (very strange since `.npy` are supposed to be platform independent) with a UnicodeDecodeError,
     
    rth@byzance examples % python example_hansenlaw_basex.py                                                                           [contribution_guidelines]
    Loading data/O2-ANU1024.txt.bz2
    image size 1024x1024
    Performing Hansen and Law inverse Abel transform:
    HL: Calculating inverse Abel transform:  image size 1024x1024
    HL: Individual quadrants
    HL: Calculating inverse Abel transform ...
    0.22 seconds
    Generating speed distribution ...
    0.17 seconds
    Performing basex inverse Abel transform:
    Loading basis sets...       
      File "/home/rth/src/PyAbel/abel/basex.py", line 209, in get_basis_sets_cached
        M, Mc, M_left, M_right, M_version  = np.load(path_to_basis_file)
      File "/usr/lib64/python3.4/site-packages/numpy/lib/npyio.py", line 393, in load
        return format.read_array(fid)
      File "/usr/lib64/python3.4/site-packages/numpy/lib/format.py", line 602, in read_array
        array = pickle.load(fp)
    UnicodeDecodeError: 'ascii' codec can't decode byte 0x8b in position 3953: ordinal not in range(128)

      File "example_hansenlaw_basex.py", line 62, in <module>
        verbose=True, calc_speeds=True)
      File "/home/rth/src/PyAbel/abel/basex.py", line 91, in BASEX
        M, Mc, M_left, M_right = get_basis_sets_cached(n, nbf, basis_dir, verbose)
      File "/home/rth/src/PyAbel/abel/basex.py", line 212, in get_basis_sets_cached
        M_left, M_right, M, Mc = np.load(path_to_basis_file)
      File "/usr/lib64/python3.4/site-packages/numpy/lib/npyio.py", line 393, in load
        return format.read_array(fid)
      File "/usr/lib64/python3.4/site-packages/numpy/lib/format.py", line 602, in read_array
        array = pickle.load(fp)
    UnicodeDecodeError: 'ascii' codec can't decode byte 0x8b in position 3953: ordinal not in range(128)

Here is the original file http://r0h.eu/d/basex_basis_1025_513_v1.npy (17 MB), while when I regenerate this basis set, I get a file with a different size http://r0h.eu/d/basex_basis_1025_513_v2.npy (23 MB). Not sure to what this could be due. But anyway, as experience shows that we keep changing those basis set files every couple of weeks, so it would probably be better to keep them out of git history (as any other large files), and just provide a download link, as we discussed before, no? We can also add those basis sets manually when [creating releases on GitHub](https://help.github.com/articles/distributing-large-binaries/) which would indeed be nice for new users :)

DanHickstein:
This looks awesome!

The basis set push was an accident. Can we add ``*basex_basis*`` to ``.gitignore`` to keep me from doing this again?

I'll activate the appveyor now.

rth:
Great thanks! I also removed that basis set. The AppVeyor build is a bit slow (2min per build x 4) so out of the 4 builds (Python 2.7/3.4 and 32/64bit), maybe we can keep only the Python 3.4 32/64bit if that's too long to wait  as the Py2/Py3 compatibility is already tested by Travis Ci.

DanHickstein:
Awesome! I love seeing all of the green check marks:) 

TravisCI does the builds in parallel, right? Weird that AppVeyor does not...

In this case, it makes sense to just test one python version on Windows. It's hard to imagine that there are Py2/Py3 issues that would only show up on one OS.

rth:
@DanHickstein Yeah, I think you have to pay to get parallel builds with AppVeyor, but well it's already nice of them to do what they do for free. 

I added a last few changes,
 - I removed the 2 Python 2.7 builds on AppVeyor (so now the build should complete in less than 5 min)
 - Added an AppVeyor badge to the README.md
 - Added a benchmark to calculate execution time for different implementations under `abel.benchmark.AbelTiming` . This runs the transform multiple times with `timeit` and does the average of the execution time, so it is a bit more reliable particularly for small image sizes. It can be used as follows,

        In [1]: from abel.benchmark import AbelTiming
           ...: time = AbelTiming(n=[401, 801, 1601], n_max_bs=1000)
           ...: # this will take a while for such large array sizes
           ...: # mostly because we have to generate the corresponding basis sets
           ...: print(time)
           ...:
        PyAbel benchmark run on Intel(R) Core(TM) i7-4600U CPU @ 2.10GHz
 
         =========     Direct Abel implementations ==========
         time to solution [s] -> transform (basis sets generation)
 
         =========    Inverse Abel implementations ==========
         time to solution [s] -> transform (basis sets generation)
 
         Implementation |    n = 401          |    n = 801          |    n = 1601 
         ----------------------------------------------------------------------------
              HansenLaw |  3.4e-02            |  1.8e-01            |  5.8e-01    
                  BASEX |  1.2e-02 ( 1.3e+01) |  1.6e-01 ( 8.4e+01) |      nan (     nan)

Ok, I'm merging this. We can fix things later if something needs to be improved...

DanHickstein:
Oh, benchmarks! Fun!

Here's my Macbook Pro (``Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz``)


```
PyAbel benchmark run on i386

 =========     Direct Abel implementations ==========
 time to solution [s] -> transform (basis sets generation)

 =========    Inverse Abel implementations ==========
 time to solution [s] -> transform (basis sets generation)

 Implementation |    n = 401          |    n = 801          |    n = 1601         
 ----------------------------------------------------------------------------
          BASEX |  5.8e-03 ( 7.4e+00) |  1.3e-01 ( 4.9e+01) |      nan (     nan) 
      HansenLaw |  3.0e-02            |  1.5e-01            |  5.5e-01     
```

So I beat you by a little bit! :)

Also, note that ``platform.processor()`` returned ``i386`` for my processor. I think that it's not really platform independent, which is disappointing. There are some hacks on StackOverflow: http://stackoverflow.com/questions/4842448/getting-processor-information-in-python

stggh:
(1) Timing (excellent @rth)
>Model name:            Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz
>PyAbel benchmark run on x86_64
>
>=========     Direct Abel implementations ==========
> time to solution [s] -> transform (basis sets generation)
>
>=========    Inverse Abel implementations ==========
> time to solution [s] -> transform (basis sets generation)
>
> Implementation |    n = 401          |    n = 801          |    n = 1601         
>  ----------------------------------------------------------------------------
>          BASEX |  1.4e-02 ( 2.1e+01) |  1.2e-01 ( 1.2e+02) |      nan (     nan) 
>      HansenLaw |  8.4e-02            |  2.4e-01            |  6.5e-01       

Improves on the 2nd run: 
>  ----------------------------------------------------------------------------
>          BASEX |  1.6e-02 ( 1.7e+01) |  7.5e-02 ( 1.0e+02) |      nan (     nan) 
>      HansenLaw |  4.5e-02            |  1.4e-01            |  5.2e-01      

(2) @rth Unit tests:
`in the test_hansenlaw_gaussian the ratio between the the numerical reconstruction and the analytical result for a Gaussian is ~0.70-0.75 `

 Interesting. In the past the absolute intensity had no importance for my (relative) data. I will check this.

(3) @DanHickstein   `small changes that you made to the example`

 There is an example_hansenlaw_Xe.py and some comments that I will put back.
Plus the additional symmetry variables for ```iabel.hansenlaw()```

(4) Is there any particular structure planned for the Wiki? 

It would be nice to have an ipython notebook example, which would provide a much simpler introduction, rather than the somewhat complex looking ```examples/example_hansenlaw.py``` etc.

DanHickstein:
Yes, that would be great if you can add in the ``example_hansenlaw_Xe.py``. Actually, it might be nice if you did the same basex/hansenlaw comparison that you did for the O2 data. 

I did not have any specific plans for the wiki or for the documentation in general, but adding some documentation is on the priority list. We started some discussion in https://github.com/PyAbel/PyAbel/issues/25, but this was more to do with the function documentation (probably using Sphinx). If we want to include some iPython notebook examples/tutorials (**awesome idea**), then I think that we could probably just include them in the repo, or on a separate Github.io repo. In any case, if you want to make a notebook example and make a PR to add it to the repo, then this seems like a good short-run solution.

rth:
@DanHickstein  Yeah, `platform.processor()` doesn't seem to be platform independent. I'll see if we can use of the hacks proposed in that SO link. 

@stggh 
1. It's strange that timing changes with the second run (I don't see that on my laptop), which would mean that's the actual values are not reliable/reproducible. I think this might be due to CPU throttling, but anyway this is more to give some relative comparison between different implementations, and check the scaling with the image size.
2. We observed as similar problem in BASEX (also because the absolute scaling didn't matter) and now use an empirical scaling constant to be consistent with the  analytical solution.
3.  - 
4. I totally agree about using ipython notebooks. We do need to discuss how to organize things between the actual documentation, whether we would use the github wiki, where to host the overview page with nice pictures @DanHickstein  that we discussed. On the other hand, I'm a bit cautious about notebooks under version control, as they do contain large binary blobs (images) that would change from update to update and every version would be saved into the git history, thus increasing the repo size (still fighting for the repo size here :) ). There is a long discussion about it athttps://stackoverflow.com/questions/18734739/using-ipython-notebooks-under-version-control but that does sound a bit complicated. So we need to think about where to put those notebooks. Maybe gist, or just render through Github.io as you mention...

stggh:
(2)  Symmetric Gaussian unit test:
I passed the 1-d Gaussian directly to ```abel.hansenlaw_transform()```
```
def test_hansenlaw_1d_gaussian():
    """Check a gaussian solution for HansenLaw"""
    n = 101
    r_max = 50
    n2 = n//2

    ref = GaussianAnalytical(n, r_max, symmetric=True,  sigma=10)
    tr = ref.abel[None].reshape(n,)  # convert to 1-d array

    recon1d = np.atleast_1d(iabel_hansenlaw_transform (tr[:n2]))   # 1/2 of 1-d array
    recon1d = recon1d.reshape(n2)  # 1-d array, again!
    ratio = n2

    plt.plot(tr[:n2],label='Abel Gaussian')
    plt.plot(recon1d*ratio,'ro',label='HansenLaw')
    plt.legend(loc=2)
    plt.savefig("HLsymmGauss.png",dpi=50)
    plt.show()


if __name__ == "__main__":
    test_hansenlaw_1d_gaussian()
```
and the agreement is pretty good, apart from the 2nd last point (N-2). Note, for this particular test I increased n = 101. I will continue to investigate this and the associated changes that should be made to ```hansenlaw```.
The scaling is, nicely, the size of the 1-d vector passed to HL. 
hlsymmgauss

stggh:
(4) ipython notebook. 
The good:  .ipynb files are rendered directly by github
The bad:  @rth  is correct that the included images may rapidly increase file size
There is ipynb-compress <url>https://pypi.python.org/pypi/ipynbcompress</url> which may help keep the size under control

DanHickstein:
Gosh, these edge cases are annoying! 

I wonder if the problem with the last pixel (N-1 and N-2 are the same...) has to do with the fact that the r-coordinates don't quite line up between the the analytical and the hansen-law solution. I got it to look a little bit better by starting the x-coordinate at 0.5 instead of 0. 

Also, note that in your code you are comparing the projection with the abel transform, when you actually want to compare it to ref.func (the original function). They are both Gaussian, but separated by a scaling factor of sigma. 


Here is what I've been playing with:
```python
from abel.analytical import GaussianAnalytical
from abel.hansenlaw import iabel_hansenlaw_transform

import numpy as np
import abel.hansenlaw

r = np.arange(0.5,25.5,1)
sigma = 10
A0=1

orig = A0*np.exp(-r**2/sigma**2) 
proj = sigma * np.sqrt(np.pi) * A0*np.exp(-r**2/sigma**2)

orig = orig[::-1] # flip them
proj = proj[::-1]

recon = iabel_hansenlaw_transform(proj)[0]

ratio = 3.1
# ratio = np.mean(orig/recon)

plt.plot(r,orig,label='Abel Gaussian')
plt.plot(r+0.5,recon*ratio,'ro',label='HansenLaw')

plt.legend(loc=2)
plt.savefig("HLsymmGauss.png",dpi=50)
plt.show()
```

image

___
Regarding the notebooks, it sounds like they should certainly stay out of the main repo to reduce the potential for increasing the repo size and other Git craziness. This shouldn't post a problem, since we can just link to them from the readme (or other documentation).

stggh:
Hansen and Law's paper provides analytical function examples, input <math>f(r)</math> and the transform output functions <math>G(R)</math>. Here is the situation for "curveA":
hlcuvean 510 002
hlcuvean 510 020
The algorithm is very sensitive to &Delta;  which definitely needs to be sub-pixel.

Apologies for my poor code:
```
#!/usr/bin/env python
from abel.analytical import GaussianAnalytical
from abel.hansenlaw import iabel_hansenlaw_transform

import numpy as np
import abel.hansenlaw

N = 51
delta = 1.0/(N-1)
rl = np.arange(0,0.5+delta,delta)   # 0 <= r <= 0.5
rr = np.arange(0.5+delta,1.0,delta) # 0.5 < r < 1.0

f = lambda r: 1-2*r**2 if np.all(r) <= 0.5 else 2*(1-r)**2

def g (R):
    alpha = np.sqrt(1-R**2)
    R2 = R**2

    if np.all(R) <= 0.5:
        beta  = np.sqrt(0.25-R**2)
        return (2/3)*(2*alpha*(1+2*R2)-beta*(1+8*R2))-\
               4*R2*np.log((1+alpha)/(0.5+beta))
    else:
        return (4/3)*alpha*(1+2*R2)-4*R2*np.log((1+alpha)/R)

orig = np.concatenate((f(rl),f(rr)),axis=0)
proj = np.concatenate((g(rl),g(rr)),axis=0)
orig = orig[::-1] # flip them
proj = proj[::-1]

recon = iabel_hansenlaw_transform(proj)[0]

ratio = np.pi/delta

r = np.concatenate((rl,rr),axis=0)
orig = orig[::-1]
recon = recon[::-1]
plt.plot(r,orig,'b',label='"Curve A"',lw=2)
plt.plot(r,recon*ratio,'r--',label='HL inverse Abel',lw=2)
# pretty labels
plt.annotate("$N={:d}$".format(N),(0.02,0.80),fontsize=20)
plt.annotate("$\Delta={:.3f}$".format(delta),(0.02,0.75),fontsize=20)
plt.annotate("$f(r) = $",(0.35,0.81),fontsize=20)
plt.annotate("$\{$",(0.5,0.785),fontsize=30)
plt.annotate(r"$1-2r^2,$   $0 \leq r \leq 0.5$",(0.55,0.85),fontsize=20)
plt.annotate(r"$2(1-r)^2,$ $0.5 \leq r \leq 1.0$",(0.55,0.75),fontsize=20)

plt.axis(ymax=1.1)
plt.xlabel("$r$")
plt.ylabel("$f$")

plt.legend()
plt.savefig("HLCuveAN={:d}{:.3f}.png".format(N,delta),dpi=50)
plt.show()
```

stggh:
@DanHickstein code for a 0.01 &Delta;, no grid offset:
hlsymmgauss

```
delta = 0.01
r = np.arange(0,25+delta,delta)
```
Some form of pixel interpolation should be applied before the HL inverse Abel transform is applied. 

Now, to sort out the even size image grid issue.

stggh:
The conclusion from my analysis above is that the Hansen and Law algorithm works for odd size images, and requires  a small  grid size, or large array size, in order to correctly converge.

Attempts to transform a Gaussian, centred on a grid, i.e. even size, have failed, the inversion diverges.
hlgauss-grid


My second conclusion is that image centre must be on a pixel, not grid. i.e. the analysis can only handle an odd size image. The two algorithm image requirements are consistent.

Centring may be achieved through interpolation (, or re-binning the electron coordinates, in my experiment).

I will modify the Hansen and Law code to handle an image centred on a pixel.

-----------------------------------------------------------

Code, for the figure, above:

```
from abel.hansenlaw import iabel_hansenlaw_transform

import numpy as np
import abel.hansenlaw

Gauss = lambda a,sigma,r: a*np.exp(-r**2/sigma**2)

delta = 0.01   # small step size required for Hansen and Law algorithm
r = np.arange(0,25+delta,delta)  # odd size image
roff = r + 0.5 # 1/2 pixel offset corresponding to even size image

sigma = 10
A0=1

# Gaussian functions
orig  = Gauss(A0,sigma,r)
origo = Gauss(A0,sigma,roff)

proj  = sigma * np.sqrt(np.pi) * orig    # Abel transform is also a Gaussian
projo = sigma * np.sqrt(np.pi) * origo

# Hansen and Law inverse Abel transform, array flipped so outer edge
# is the first array element
recon  = iabel_hansenlaw_transform(proj[::-1])[0][::-1]  # flip/flip
recono = iabel_hansenlaw_transform(projo[::-1])[0][::-1]

# correct scaling
recon  *= np.pi/delta
recono *= np.pi/delta

# curves, with bar char plot to illustrate respective grids
ax=plt.subplot(121)
plt.plot(r,orig,'b',lw=2,label='Gaussian')
plt.plot(r,recon,'r--',lw=2,label='HL inv Abel')
plt.bar(r[::100],orig[::100],width=1,color='lightblue',align='center')
plt.title("centre is a pixel")
plt.legend()
plt.subplot(122,sharex=ax,sharey=ax)
plt.plot(roff,origo,'b-',lw=2)
plt.plot(roff,recono,'r--',lw=2)
plt.bar(roff[::100],origo[::100],width=1,color='lightblue',align='center')
plt.title("centre is the grid")
plt.axis(xmin=-1.5,xmax=15,ymin=0.2,ymax=1.2)


plt.savefig("HLGauss-grid.png",dpi=50)
plt.show()
```

rth:
This is very interesting.

Just a few question to better understand what's happening:
> The algorithm is very sensitive to Δ which definitely needs to be sub-pixel.
>   `rl = np.arange(0,0.5+delta,delta)   # 0 <= r <= 0.5`

So Δ is  the pixel size, if I understood correctly? How can it be sub-pixel then?

From skimming through the article:
> The methods proposed are recursive in form: Each succeeding point of the Abel inversion is computed from the previously computed point and a new data point. 

and
 
> The conclusion from my analysis above is that the Hansen and Law algorithm [..] requires a small grid size, or large array size, in order to correctly converge.

So as far as I understood, it's a recursive method not an iterative one (not that I know much about the former ones) , meaning that after a finite number of operations are performed it should yield to correct result, right? Unlike in a iterative method where we can increase the number of iterations to reach convergence. It that sens this would be worrisome as it would mean that for some random signal, if the grid size is large, we could run  HansenLaw, it would not produce any errors but the results would be a bit off (although it makes sens that the  numerical solution converges to the analytical one with the grid resolution, and this would be true for all implementations)

I think, at some later time, when we have more algorithms implemented, it could be useful to a) do some benchmarks c) write down what are the limitations and advantages of each if that can be clearly defined. 

Also at some future PR, could we add an optional`dr=delta` parameter (initialized at `dr=1`) to `iabel_hansenlaw`  (and possibly to `iabel_hansenlaw_transform`)   so that the scaling is done internally, as it is currently done in basex? Or some more verbose name if you prefer. 

>  the analysis can only handle an odd size image.
> Centring may be achieved through interpolation (, or re-binning the electron coordinates, in my experiment).

It is interesting that this is the case. Ok, this whole thing about odd/even pixel sized is annoying. I know we talked at length about it already. If the algorithm does not converge for even image sizes,  IMO the `iabel_hansenlaw` should raise an error if such an image is provided (instead of silently returning wrong results). On the other hand, the centering function could be made to accept both odd and even image sizes and return whatever is needed by the implementation later  (by adding fake rows etc as discussed earlier). Or maybe let's add a few more implementations, see what's the situation there, and decide on a global solution to this problem?

stggh:
@rth thanks for your clever thoughts/ideas.

1. &Delta; is the "function" sampling interval. This function is the pixel intensity along one row of the 1/2 image. In principle, we may sub-sample.

2. Iterative:  A large sampling interval may provide a poor (derivative <math>g'</math>) correction to the next inner pixel.

3. Raise error:  IMO not necessary.

  (a) My real examples, with even size images, have shown that the hansenlaw inversion does produce very close to the correct result.  This is no doubt due to the large pixel count, N>=1024.

 (b) The Gaussian test case for N even is a puzzle.  I think that there is some other factor at play here that I have missed. I will investigate further.

 (c) I will also shift the centre of an experimental velocity-map image to place it mid-pixel (odd size), to compare the effect.

4. Centering function:  should be able to shift the image by real <math>(&Delta;x,&Delta;y)</math> coordinates. In general, this is achieved using interpolation, but that may open a can of worms.

stggh:
For grid centred image, dropping some pixels, from left side centre, "fixes" the transform.
More later, off to Christmas lunch.


hloddeven

DanHickstein:
Another great discussion! We are all set to write a paper about the tiny details of the Abel transform :)

A few quick thoughts (sorry, did not have much free time this week):

I am confused about the &Delta; discussed above. @stggh, a few posts ago, you showed "curveA" with two different &Delta;s, and demonstrated that with the larger &Delta;, the HL transform did not agree well with the analytical result, but it did for the small &Delta;. You indicated in that plot that ``N=51`` in both cases. Was ``N`` indeed constant? It would make more sense if increasing ``N`` improved the behavior of the transform.

Fundamentally, the HL algorithm doesn't know about &Delta;, it only knows the number of pixels in the image. Basically, I think that we are discovering that HL doesn't behave perfectly for "undersampled" objects, where "undersampled" probably has something to do with how accurately the gradient can be calculated when the slope is changing significantly from pixel to pixel. This seems reasonable, and, as @rth points out, they say this in their paper. 

We could compare the output of np.gradient to the analytical slope of function and see how the error increases. [``np.gradient``](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.gradient.html) uses ["second order accurate central differences"](https://en.wikipedia.org/wiki/Finite_difference#Higher-order_differences
), is this what we want? There may be other ways to calculate the gradient that behave better for undersampled situations (spline interpolation?). 

But maybe we just need to print a warning when the number of points is very small. I'll but that many of the algorithms don't actually work when the curve is severely undersampled. Some of the methods that use large basis sets may be more robust, but in general, it seems like a difficult problem to solve. 

Oh, also, we should be careful to test the inverse abel algorithms with a function that goes very close to zero at the highest radius. I think that the inverse abel transform needs to contain the entire object in order to be completely accurate. (i.e., some of the gaussian functions that we were using are still maybe 1% at the maximum value at the edge of the plot, and I think that this will cause a small error in the transform.)

Regarding sub-pixel centering, there is a numpy function for shifting images with interpolation, so it should be pretty efficient:
http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.ndimage.interpolation.shift.html
The problem is that it will make centering even more difficult to think about :)

stggh:
@DanHickstein - as usual very insightful thoughts
  1. Paper - excellent idea. PyAbel should be the definitive go-to place for Abel transforms.
  2. &Delta; and N:  Good point! The code above, does change N = 50,  500, since the domain is fixed [0,1).
  3. Undersampling: Agreed, this is most likely the issue.
  4. Spline interpolation: May not be good choice for noisy images.
  5. <math>G(&infin;) = 0</math>  is also worth investigating. Some VMI measurement methods involve adding multiple camera images, accumulating background noise = bad.
  6. interplation.shift:  Great.  I will examine this method.
     For my velocity-map imaging analysis, the image is centred with respect to the object, irrespective of the pixel grid. It just happens to be dumped as an even image. 
I think, in the end, the HL algorithm will be independent of even/odd image size.

More investigation, and more careful programming required on my part.

stggh:
Following on from the tests above, I have changed  ```hansenlaw.py``` to process odd images, i.e. the image centre is mid-pixel. 

```scipy.ndimage.interpolation.shift(im,(-0.5,0.5))``` followed by im = im[:-1,1:] appears to recentre the ANU even images to odd (this needs further verification).  I have modified the example_hansenlaw....py files to suit.

I decided to split ```iabel_hansenlaw ()``` into ```iabel_hansenlaw(img, calc_speeds=True, verbose=True)``` and ```iabel_hansenlaw_symmetric (img, use_quadrants=(True,True,True,True), vertical_symmetry=True, horizontal_symmetry=True, calc_speeds=True, verbose=True)```. The former inverts an image as is. The latter retains the quadrant processing, with vertical and horizontal symmetry constrains.  This change makes the ```iabel_hansenlaw()``` helper function easier to understand.

Note, I have experienced big issues, rebasing my fork back to the PyAbel/master, and then pushing my changes to the fork. It would be nice to know that my changes are compatible with a future PR?

```test_hansenlaw_gaussian()``` fails - I will investigate.

```./example_hansenlaw_basex.py 
Loading data/O2-ANU1024.txt.bz2
Even pixel image cols={:d}, adjusting image centre
  shift(im,(-0.5,-0.5)
image size 1023x1023
Performing Hansen and Law inverse Abel transform:
HL: Calculating inverse Abel transform:  image size 1023x1023
0.68 seconds
Generating speed distribution ...
0.36 seconds
Performing basex inverse Abel transform:
Loading basis sets...           
0.09 seconds
Reconstructing image...         
0.82 seconds
Generating speed distribution...
0.36 seconds
```
o2-anu1024_comparison_hansenlaw

DanHickstein:
Nice work! 

What stops you from just combining the two functions into one? You could just add:

```python
if not vertical_symmetry and not horizontal_symmetry and use_quadrants.all():
    AQ0 = iabel_hansenlaw_transform(Q0)
    AQ1 = iabel_hansenlaw_transform(Q1)
    AQ2 = iabel_hansenlaw_transform(Q2)
    AQ3 = iabel_hansenlaw_transform(Q3)
```
to ``iabel_hansenlaw_symmetric`` and then it would cover all of the cases, correct?

Sorry about the trouble merging with PyAbel/master. I think that the only conflict is in tools.py, where I changed two lines in the index_coords function. The rest of the changes like like they can be merged just fine. 

Good to hear that the interpolation function works. I like the idea of incorporating an ``even2odd`` function in tools.py that provides a way for users to easily swap their images.

stggh:
Fixed ```test_hansenlaw.py```, by increasing N.

@DanHickstein - excellent point.  Will do.  I had not  yet convinced myself that quadrant division is equivalent.

Issue #39, opened 2015-12-01
Discretization of the grid for the Abel transform


rth:
While I think we all agree on the definition of the image center as discussed in issue #34 , illustrated on the figure below, 

![test](https://cloud.githubusercontent.com/assets/630936/11503290/1c7e06f4-9840-11e5-9791-76d18db76ca5.png)

the fact that the transform requires odd or even images, also has wider implications on the spatial discretization assumed by the algorithm (which I have personally never considered). For instance in 1D, the two possible discretization are illustrated on the figure below,

![abel_discretization](https://cloud.githubusercontent.com/assets/630936/11503199/90bb9a96-983f-11e5-88cc-e153f63f019b.png)
where the pixel size is `dr`. @stggh Regarding your earlier comment, both approaches are uniformly sampled, as far as I understand. 

Although, in practice, far away from the origin, I suppose it shouldn't really matter, the way the grid is discretized is probably a fundamental property of a given implementation. Therefore,  I'm not sure we can expect all implementations to support both odd and even images. For instance at present, it appears that there is the following separation,

 - **Odd images** : BASEX, BASEX_asym (originally)
 - **Even images**: HansenLaw, direct (I think), Fourier-Hankel (I think)
 
Another applications is, for instance, 2D cylindrical fluid dynamics simulation, that I project with the forward Abel transform, are also in the "Even images" case.  

In the particular case of the `direct` implementation, I'm not so inclined to redo all math to see what it should be if we change the discretization of the grid for odd images, nor do I think it would be that useful (more code => more possibilities for bugs, possible inconsistent behavior between odd and even images, and finally it shouldn't matter anyway for real data of > 1000px ).   

So the question is whether in general we should,

 a. Raise an error if e.g. an algorithm working with odd images is given an even image
 b. Hack the input/output image  as to work both with odd and even images, while actually providing the transform with the an image of the parity it supports. For that  we could use @DhrubajyotiDas proposition, i.e: 
   - *even -> odd*:  add a fake central row, as an average of the neighbor rows, then remove it after the processing
   - *odd -> even*: split the central row in 2, then take the average of the resulting 2 rows before outputting the image.
This would introduce some additional error, particularly near the origin, because of a shift by 1/2 pixel of the whole image when this hack is used, but I don't think it would be that bad, particularly as the data near the origin is noisy already in the majority of implementations. 

[...]

PR #46, opened 2015-12-03
Hansen and Law algorithm updates

stggh:
The main issue is that ```reproject_into_polar()``` does not appear to handle even pixel sized images, due to the use of (integer,integer) coordinate centre, and associated ```np.meshgrid()```, as discussed in issue #39.  Edit: I take it back ```index_coords()``` accepts origin=(x,y). Just need to pass origin=(x,y) via tools.calculate_speeds()```, I think.

NB ```iabel_hansenlaw()``` performs the correct inverse Abel transform, so the code is still good to merge, with a caveat on the speed distribution, that is nearly correct.

### DIFF
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -193,7 +193,8 @@ def iabel_hansenlaw (data,quad=(True,True,True,True),calc_speeds=True,verbose=Tr
         verboseprint('Generating speed distribution ...')
         t1 = time()
 
-        speeds = calculate_speeds(recon,(N2-0.5,N2-0.5))
+        image_centre = (N2,N2) if N2%2 else (N2-0.5,N2-0.5)
+        speeds = calculate_speeds(recon,origin=image_centre)
 
         verboseprint('{:.2f} seconds'.format(time() - t1))
         return recon, speeds
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -9,7 +9,8 @@ import numpy as np
 import multiprocessing as mp 
 from time import time
 from math import exp, log, pow, pi
-from abel.tools import calculate_speeds, get_image_quadrants
+from abel.tools import calculate_speeds, get_image_quadrants,\
+                       put_image_quadrants
 
 ###########################################################################
 # hasenlaw - a recursive method inverse Abel transformation algorithm 
@@ -157,7 +158,7 @@ def iabel_hansenlaw (data,quad=(True,True,True,True),calc_speeds=True,verbose=Tr
 
     t0=time()
     # split image into quadrants
-    Q = get_image_quadrants(data, reorientate=True)
+    Q = get_image_quadrants(data, reorient=True)
     (N2,M2) = Q[0].shape   # quadrant size
 
     AQ = []  # empty reconstructed image
@@ -184,15 +185,8 @@ def iabel_hansenlaw (data,quad=(True,True,True,True),calc_speeds=True,verbose=Tr
            AQ.append(pool.map(iabel_hansenlaw_transform,\
                      [Q[q][row] for row in range(N2)]))
 
-    # for odd-pixel image trim off 1-pixel from end (centre)
-    if N2%2:
-        for q in (1,2,3):
-            AQ[q] = AQ[q][:,:-1]
-
     # reform image
-    Top    = np.concatenate ((AQ[1],np.fliplr(AQ[0])),axis=1)
-    Bottom = np.flipud(np.concatenate ((AQ[2],np.fliplr(AQ[3])),axis=1))
-    recon  = np.concatenate ((Top,Bottom),axis=0)
+    recon = put_image_quadrants(AQ,odd_size=N%2)
             
     verboseprint ("{:.2f} seconds".format(time()-t0))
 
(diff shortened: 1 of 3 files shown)


rth:
This is great!

Just a suggestion that I forgot to mention in the orginal PR about the `freecpus`. I think a more conventional approach would be just to set `ncpus=`, as `freecpus` can be potentially confusing. For instance, say you have a dual core multithreaded laptop, so it has 2 CPU cores, but htop shows 4 CPUs. You won't get much speed improvement above 2 parallel processes, but I have a doubt whether `mp.cpu_count()` returns 4 or 2 (probably 4). The point is you aways have to do the arithmetics of how many cores you have vs how many cores you want to run this on. Also, I think that it's fine to run on all avalable cpu, as the OS should handle the load balancing, so there is no need to set `all_cpus  - 1` as the default, I think, no? 

Also allowing to fall back to a non parallel implementation would be nice (unless `mp.Pool(1).map` is equivalent to the built-in `map`), as the tracebacks from withing `Pool.map` can be a bit confusing. So what do you think about something like this (or replace `ncpus` by some name you find more intuitive)?

     iabel_hansenlaw (... ,   ncpus='all')

          if ncpus == 'all':
               pool = mp.Pool() # I think this uses all cpus (need to check)
               pmap =  pool.map
          elif ncpus > 1:
               pool = mp.Pool(processes=ncpus) 
               pmap  = pool.map
           elif ncpus == 1:
               pmap = map # fall back to a non parallel implementation with the built-in map
           else:
               raise ValueError('Wrong value for the input parameter ncpus={}'.format(ncpus))

           # then use pmap as before

what do you think about it?

Also as a side note, how much speed speed up did you get from using numba JIT on `iabel_hansenlaw_transform`? I agree with @DanHickstein earlier comments, because it has quite a few nested loop and essentially operates on indices, there should be some room for optimization (I was thinking `Cython` for `iabel_hansenlaw_transform`, as I'm already using it in my `direct` implementation, so that won't add dependencies unlike `numba`) But we can discuss this at some later point, after the direct implementation is merged.

Finally, you would have to rebase this on the current PyAbel master, since github shows that it can't be automatically merged right now :)

stggh:
Thanks, Roman.  I will try to implement your suggestions tomorrow. 

numa JIT gave only a fraction of a second "speed up", something like 14.5s cf 14.7s (for a particular test case) and so, is not a worthwhile change. Cython may be a possibility.

BTW, the speed calculation should really include \sin\theta in the sum over angle (if it is summed in angle. I have not yet fully understood ```np.sum(polarIM,axis=1)```).

DanHickstein:
@rth's intuition for parallel processing agrees with what I found testing Hansen–Law a few days ago. On my MacBook i7, ``mp.cpu_count()=4``, since it's dual core with hyperthreading. Setting ``free_cpus=3`` runs in 13 seconds. ``free_cpus=2`` runs in 7 seconds (nearly 2x speedup). Setting ``free_cpus=2``or anything lower (and I tried -99) have the same 7 sec runtime. 

Also, there is no noticeable delay in the responsiveness of the system, so it seems fine to run on all the available cores (though it would appear that hyperthreading means that many processors are lying about their true parallel processing capability).

Okay, maybe we should ignore all of that, because I think that I figured out how to just vectorize the outer loop (the loop over the rows) and now it runs in 0.2 seconds with no parallel processing calls (seems too good to be true...). I just passed the whole quadrant to ``iabel_hansenlaw_transform`` and iterated over each column at once. 

Here's the repo with the code: 
https://github.com/DanHickstein/PyAbel/tree/hacking-hansen-law
Let me know what you think...

--
And yes, you are correct that the calc_speeds should include a factor of sin(theta)*R in order to take into account that the object is going to have more volume as the distance from the central axis becomes larger. I think that what we are currently doing is a correct integration of the central slice, and I would like to keep this capability in the ``calc_speeds`` function, since I actually use this all of time in other aspects of my research. Possibly we could include a keyword argument ``use_jacobian=True`` or ``volume_element=True`` (I think that these are the correct terms, but my Calculus III class was a long time ago at this point). This way users could select if they are looking for a simple angle integration of a 2D image or if they want us to take into account the cylindrical symmetry to give them the distribution of the real 3D object. The difference is important, because information about theta is lost in the returned 1D array, so users cannot simply multiply or divide the array to obtain the different behavior. 

Also, as @rth pointed out a while back, the name ``calc_speeds`` is only meaningful to the photoelectron/photoion community, and we should change the name to something like ``integrate_angular`` to indicate an integration over all angular coordinates.

DanHickstein:
So, I think that I'm just going to merge this in order to preserve the improvements to the quadrants code in tools.py. 

Then I can include my changes in a separate PR.

stggh:
Wow! @DanHickstein an amazing code improvement. Well done! This is what open source is all about. 

The speed distribution is still an issue. On close inspection of the O2- speed spectrum, HL appears to have less resolution, which implies that the image centre is offset. The origin offset (-0.5,-0.5) applied to even images makes the HL speed align with basex speed, but basex is speed is (slightly) incorrect for even images. i.e. the (-1/2,-1/2) pixel offset made the HL speed align with an incorrect basex speed. More investigation required.

DanHickstein:
Yes, the calc_speeds is still not doing something quite right. We'll need to think about this more. 

Take a look at the vectorized code (and other changes) in https://github.com/PyAbel/PyAbel/pull/47 and let me know if you're okay to go ahead with the merge.

PR #47, opened 2015-12-03
Vectorized Hansen and Law transform

DanHickstein:
I removed the loop over the rows by just passing ``iabel_hansenlaw_transform`` the 2D arrays and having it work with 2D arrays instead of 1D. The rows are all independent, so there should be no difference in the results, it's just faster math to let numpy wrestle with all of the rows at once instead of one-by-one. 

I also removed the multiprocessing stuff, since it's not necessary now. 

I also changes ``np.all(quad)`` to ``np.any(quad)`` in line 180 (new line 185) or hansenlaw.py. I think this was the right thing to do. 

I also added a few things to .gitignore to avoid committing basis sets and example results.


stggh:
All looks very good. I suggest adding your name to the header block + comment on vectorization. This is a major achievement. Also, add Roman, as each code suggestion input has contributed.

Amazing! I am busy checking some big 8192x images to marvel at the processing speed ...

rth:
@stggh No it's ok, you can leave my name out of it :) @DanHickstein Awesome, vectorizing was the best solution! Great work!

DanHickstein:
I added a comment about the vectorization in the header. Did this before I saw Roman's comment, so his name is stuck there for now :). 

I'll go ahead and merge.

Issue #1, opened 2015-02-02
Problem loading example image


DanHickstein:
Chunlong reports the following error when running the example: 

Python:>>> recon,speeds = BASEX.center_and_transform(raw_data,center,median_size=2,gaussian_blur=0,post_median=0,verbose=True)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "BASEX.py", line 139, in center_and_transform
    image = center_image(data,center=center)
  File "BASEX.py", line 159, in center_image
    H,W = np.shape(data)
ValueError: too many values to unpack

This seems to indicate that the image is not being loaded correctly. Maybe it is being loaded as a color image?

DanHickstein:
The shape of the image was (1388,1038,4), so it looks like plt.imread was loading it in rbga color. I think that I solved this by loading the raw file instead. We'll see if it works for people.

PR #49, opened 2015-12-06
hansenlaw changes to process 1/2 image

stggh:
In my attempt process the 649x519 Xe photoelectron image, I realised that the HL quadrant processing assumes a square image, when it need not. The treatment of odd/even width sized images requires only an addition of a pixel column. Hence, I removed ```get/put_image_quadrants``` from abel.tools, replaced with ```add_image_col, delete_image_col```. Any quadrant stuff is, specific to and is, handled by hansenlaw. Hansenlaw now handles odd and even images. Basex should call add_image_col(image) for even images, and delete_image_col(recon) after reconstruction, before the speed calculation. (see the example_hansenlaw files).

Agreement, between HL and basex is now excellent:
example_hansenlaw_basex

The Xe spectrum also looks good:
example_hansenlaw_xe

I am having problems rebasing with git mergetool, will try at ANU tomorrow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,7 @@ build/
 *.egg-info/
 *.npy 
 examples/data/Xenon_800_nm*
+examples/*.pdf
+examples/*.png
+examples/*basis*
+examples/*.dat
--- a/README.md
+++ b/README.md
@@ -2,23 +2,36 @@
 
 [![Build Status](https://travis-ci.org/PyAbel/PyAbel.svg?branch=master)](https://travis-ci.org/PyAbel/PyAbel)
 
-PyAbel is a Python package for performing Abel and (primarily) inverse Abel transforms. The Abel transform takes a 3D object and finds the 2D projection of that object. The more difficult problem -- the inverse Abel transform -- takes the 2D projection and finds the central slice of the 3D object by assuming cylindrical symmetry in the vertical direction.
+PyAbel is a Python package for performing Abel and (primarily) inverse Abel transforms. The Abel transform takes a cylindrically symmetric 3D object and finds the 2D projection of that object. The more difficult problem -- the inverse Abel transform -- takes the 2D projection and finds the central slice of the 3D object by assuming cylindrical symmetry in the vertical direction.
 
-Currently, this package only uses the BASEX algorithm creared by Dribinski, Ossadtchi, Mandelshtam, and Reisler [[Rev. Sci. Instrum. 73 2634, (2002)](http://dx.doi.org/10.1063/1.1482156)], but we hope to include more algorithms for the inverse Abel transform in the future! 
+The PyAbel package offers several options for completing the inverse Abel transform:
 
-The BASEX implementation uses Gaussian basis functions to find the transform instead of analytically solving the inverse Abel transform or applying the Fourier-Hankel method, as both the analytical solution and the Fourier-Hankel methods provide lower quality transforms when applied to real-world datasets (see the RSI paper). The BASEX implementation is quick, robust, and is probably the most common method used to transform velocity-map-imaging (VMI) datasets.
+1) The BASEX algorithm creared by Dribinski, Ossadtchi, Mandelshtam, and Reisler [[Rev. Sci. Instrum. 73 2634, (2002)](http://dx.doi.org/10.1063/1.1482156)]. The BASEX implementation uses Gaussian basis functions to find the transform instead of analytically solving the inverse Abel transform.
 
-In this code, the axis of cylindrical symmetry is in assumed to be in the vertical direction. If this is not the case for your data, the `numpy.rot90` function may be useful.
+2) The "Hansen and Law" recursive method described in [[J. Opt. Soc. Am A 2 (4) 510 (1985)](dx.doi.org/10.1364/JOSAA.2.000510)]
+
+3) In the future, we hope to have more options for the forward and inverse abel transforms.
+
+### Symmetry
+
+In this code, the axis of cylindrical symmetry is in assumed to be in the vertical direction. If this is not the case for your data, the `numpy.rot90` function can be used to rotate your dataset.
 
 ### Installation notes
 
 This module requires Python 2.7 or 3.3-3.5. It can be installed with
 
     python setup.py install --user
-	
+
+Or, if you wish to edit the PyAbel code without re-installing each time (advanced users):
+
+    python setup.py develop
 
 ### Example of use
 
-See an  example in `examples/example_main.py`.
+See several example in `examples` folder.
+
+### Contributing
+
+We welcome new implementations of the inverse Abel transform or other code improvements. Please feel free to submit an issue or make a pull request.
 
 Have fun!
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -6,7 +6,6 @@ from __future__ import unicode_literals
 
 #from numba import jit
 import numpy as np
-import multiprocessing as mp 
 from time import time
 from math import exp, log, pow, pi
 from abel.tools import calculate_speeds, get_image_quadrants,\
@@ -31,10 +30,9 @@ from abel.tools import calculate_speeds, get_image_quadrants,\
 #
 ###########################################################################
 
-#@jit
-def iabel_hansenlaw_transform (ImgRow):
+def iabel_hansenlaw_transform(IM):
     """ Inverse Abel transformation using the algorithm of: 
-        Hansen and Law J. Opt. Soc. Am A2, 510-520 (1985).
+        Hansen and Law J. Opt. Soc. Am. A 2, 510-520 (1985).
                         
                        ∞                
                        ⌠                
@@ -58,8 +56,8 @@ def iabel_hansenlaw_transform (ImgRow):
 
         Parameters:
         ----------
-         - ImgRow: a n/2 numpy vector = one row of one quadrant of the image
-           |       orientated top/left
+         - Img: a n/2 x n/2 numpy array = one quadrant of the image
+           |    oriented top/left
            |     +--------+              --------+ 
            \=>   |      * |               *      |
                  |   *    |                  *   |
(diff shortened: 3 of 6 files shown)


DanHickstein:
Thanks for thinking about this! It's great to have Hansen–Law work for non-square images. 

I actually think that the previous code *almost* works or non-square images. When I try the current code with a non-square image, I get an error, which is due to a mistake that I made while vectorizing the second-to-last line in the ``iabel_hansenlaw_transform`` function. Currently it reads:

```python
AImg[ncols-1] = AImg[ncols-2]  # special case for the center pixel
```

But this is not correct! We are accidentally setting the last *row*, but we meant to set the last *column*. Changing this line to 
```python
AImg[:,ncols-1] = AImg[:,ncols-2]  # special case for the center pixel
```
allows the current code to work with non-square images.
___
Allowing Hansen–Law to work for odd-sized images is a great idea! Doesn't adding and removing the column slightly change the results? Not a big problem, but I think that we should present a real warning, rather than just print a message (not much practical difference, but a warning sounds more important! ). Users should be encouraged to think about how they want to crop and center their images depending on the transform.
___
Also, I'm confused why you switched the behavior of the quadrants functions. It seems like splitting the image in half and splitting into quadrants are general functions that we would want to have in tools.py, but maybe there is something that makes it specific to hansenlaw?
___
One more minor thing, I noticed that you changed some functions to follow the spacing format of:
```python
def function (arg1,arg2,arg3):
```
I think that the [usual style](https://www.python.org/dev/peps/pep-0008/#code-lay-out) would use the format
```python
def function(arg1, arg2, arg3):
```
i.e., no space between the function and the open-paren, and a space between each input argument.

stggh:
@DanHickstein 
No worries, I am easy with reverting the code, but let me explain my rationale, so we may reach a consensus on the best way forward:

(1) ```iabel_hansenlaw_transform()``` is unchanged. Thanks, for spotting the last-line bug. The function happily handles any sized "quadrant" N/2xN/2 up to the NxN/2 "half-image". 

Issue #1 is a matter of how best to describe this aspect in the function documentation. 

(2) ```iabel_hansenlaw()``` was rewritten (and I happy to revert back) to take a user image, and hand ```iabel_hansenlaw_transform()``` 2xhalf images. There is a special case, where image symmetry allows quadrants to be co-added (or excluded), in which case, one quadrant is transformed. I agree that this function may be somewhat confusing.

Issue #2: How best tidy this function.

(3) odd (hansenlaw) or even (basex) width images:  
    (a) The Hansen and Law algorithm transforms a pixel row, starting at the outer edge and ending at the image centre. ``assumption of uniform sampling grid is not necessary'' (p513 JOSA paper). In principle, last n=N/2-1, &Delta; should be 1/2-pixel size.  I should fix this. The work around is to add one whole column at the image centre, i.e. one extra pixel at the end of the row.  In the end this would make little or no difference, since all the image measurement noise ends up dumped into this pixel. I believe that he HL inverse Abel transformation is correct for odd and even images.

What is important, is to ensure that ```calculate_speeds()``` receives the correct, original, sized image, with the correct origin (centre) coordinate, (-1/2, -1/2) pixel, for even size images.

(b) Basex: The current basis set limitation requires even images to be expanded by one centre column. IMO, this should be handled within ```iabel_basex()```. The reconstructed image must be returned to the original shape before the speed calculation.

Issue #3: IMO image reshaping should be transparent to the user and occur within the ```iabel_hasenlaw/basex()``` functions.

(4) ```get/put_image_quadrants()```
   @rth original code is fine. The (my!) code that adds an extra column and row is incorrect, as only the additional column is required. I believe this work around code is unnecessary, since it should be handled by our respective inverse Abel transform functions.
The original quadrant code is fine, but IMO unecessary, since there is no reason for the user hand the transform functions a single quadrant. Except, perhaps, for the anisotropy parameter, it is useful to treat image quadrants, as <math>I &prop; &beta;P<sub>2</sub>(cos&theta;)</math> is linear in <math>P<sub>2</sub>(cos&theta;)</math>.

Issue 4: I think that the tools quadrant code with image size reshaping may generate confusion for the user.

(5) Thanks, for the coding style guidance. I need this information.

rth:
Extending  Hansen–Law to work with non square images would be indeed very useful! Besides of the obvious usage, it is also nice to have for efficient 1D transformations where one dimension can be set to 1.

I'm not competent with matters related to speed calculation and anisotropy (as my applications are quite different),  so I can't really comment on that. 

> @rth The original quadrant code is fine, but IMO unnecessary, since there is no reason for the user hand the transform functions a single quadrant.

I agree with @DanHickstein  that keeping a generic `get_image_quadrants`  in `tools.py` would be useful, as is also currently used by  `abel.benchmark.is_symmetric` which is used in unit tests to check that `center_image` works correctly. So if you remove `get_image_quadrants` function, unit tests will fail. You can actually see if your local changes break anything, even before making a pull request, with

     import abel.tests
     abel.tests.run_cli(coverage=True)

or alternatively from the command line,
    
     cd PyAbel/
     nosetests -s abel/tests/

So as long as we keep `get_image_quadrants`, alongside the new `add_image_col` and `remove_image_col` (which are very useful btw!), that sound good. Then again, I'm not exactly sure what quadrant function you would need in the HansenLaw function, so even some code duplication would be acceptable depending on your requirements (i.e. if the generic `get_image_quadrants` is not suitable for you). So we can keep things as they are now, just adding back the `get_image_quadrants` function?

As a last comment,  maybe we can move  `top_bottom_image` and `left_right_image` under `tools.py` as they can also be useful for other people? In the later case we could then also replace   `mid = Hdata.shape[0]//2` by  `mid = Hdata.shape[0]//2 + Hdata.shape[0]%2` etc. to work with both odd and even images, I think?

DanHickstein:
Great discussion guys!

@rth brings up an important issue: the Hansen-Law transform should work for 1D images (i.e., 1 row, n columns). This requires just a few changes (mainly adding ``np.atleast_2d(data)`` in a few places). I can make a PR for this, but maybe I should wait a bit until we make decisions on the other issues here to avoid too much Git craziness. 

I am also not too familiar with the anisotropy parameters either, but @rth, the speeds calculation is not complicated: it's just a spherical integration of the retrieved 3D object to yield the intensity as a function of radius. Not as applicable to your situation since you don't have spherical objects, but it's just geometry. 
_____

This talk of where to put the "quadrants" functions brings up a larger issue of what options we should provide for symmetrizing the images. Even with the vertical + horizontal symmetric implementation of basex that we are currently working with, we could imagine offering the user their choice of which quadrants to use. In the case the the user wants each quadrant transformed separately, we could simply pass the basex transform four different images (Q1, Q2, Q3, Q4), one after the other. We could then stitch the final image together from the four different transforms.

Currently, the way that Hansen-Law treats the quadrants is nice, but it occurs to me that it doesn't cover all possible scenarios. As it stands, the user can either transform all 4 quadrants separately, or they can combine any of the quadrants they want into single quadrant and obtain a final image with full symmetry (vertical + horizontal). But I think that there are actually two more scenarios: the user wants to assume only vertical symmetry or the user wants only horizontal symmetry. The case of horizontal symmetry can be rephrased as combining Q1 and Q2 on the top and combining Q1 and Q2 on the bottom. This behavior cannot currently be obtained with the options currently implemented in Hansen–Law. 

Potentially, I think that the following keyword arguments would cover all possible situations:
```use_quadrants=(True, True, True, True), vertical_symmetry=False, horizontal_sym=False```

The default behavior would be to transform each quadrant independently, but any situation could be achieved. For example, if you want to combine three quadrants into one, then you could use the arguments:
```use_quadrants=(True, True, True, False), vertical_symmetry=True, horizontal_symmetry=True```

Additionally, this would avoid the need to use ``quadrants=(False, False, False, False)``, which I think is a little bit confusing.

Anyway, the point is, dividing the images up into quadrants, symmetrizing images, and adding quadrants back into images will likely be a part of many/all of the transform methods in the future, and we should try to keep as much of that code in tools.py as practical in order to maximize the code reuse. Some of the specifics will change with each implementation, for example, basex cannot work on a single quadrant (as currently implemented), while Hansen-Law can.

_____

@stggh,

In response to the numbered issues that you listed:

Issues 1 and 2: I think that these are addressed above.

Issue 3: Yes, I agree that any image reshaping should happen in the ``iabel_hansenlaw/basex`` functions. The issue of how transparent it should be to the user depends on if we think that we have completely preserved the accuracy of the transform for the wrong-sized images. If we think that we've done a perfect job, then we can reshape the image silently. If we are not sure, then we should at least print a statement to the effect the the image is being reshaped. We can always give the user an option to suppress the warning/print statement.

Issue 4: You are right, we should try to keep the quadrants functions in tools.py easy to understand. The more confusing stuff should take place in the ``iabel_hansenlaw/basex`` functions. But, one point of tools.py is to hold functions that are common to all of the transform methods. Not all of them are necessarily intended for general consumption by the users. 

Let me know your thoughts and we can decide how to proceed.

stggh:
1-d: no problem for HL.

My life was much simpler in my closed world of only even image sizes, that nicely subdivide. 

Any talk of quadrants forces an even size image (in both axes). This is what I tried to wrestle with on the W/E (and failed). It does now appear what we have is ok. I do dislike that some quadrants are padded.

I now better comprehend your inputs:
```use_quadrants=(True, True, True, False), vertical_symmetry=True, horizontal_symmetry=True```
and I agree.

For my experiment the inverted image is only for show, the remaining analysis (PES and PAD) is done only with the single combined quadrant of the symmetric image.

DanHickstein:
Oh yes, the odd/even issue makes the whole symmetry discussion even more complicated. Sounds like we should just make basex work with even images and forget about these damn odd images! :)

Okay, so maybe I will make a separate PR for fixing the typo in the last line of ``iabel_hansenlaw_transform`` and make it work for 1D datasets (which it does on paper, but not quite in practice due to my poor coding). Then we can make additional changes from there.

stggh:
Closed. It is decided that the original quadrant approach is the way forward, as per the discussion.

DanHickstein:
Sounds good. Thanks for bringing up this discussion. I think that there were some other small changes that you made to the examples and to the documentation that looked good. Feel free to make another PR with these changes (or other changes).

PR #50, opened 2015-12-07
Making Hansen-Law transform work for non-square images

DanHickstein:
Following the discussion in https://github.com/PyAbel/PyAbel/pull/49, this PR fixes the second-to-last line in the ``iabel_hansenlaw_transform`` function in order to allow the transform to work with non-square images. 

I am actually not sure if this line is even necessary. The center few pixels are noisy in either case, and I cannot really tell if there is a difference by eye. @stggh, does this line serve a specific purpose?

___

Also, I added a ``np.atleast_2d()`` statement to make ``iabel_hansenlaw_transform`` work if it's passed an array that is 1D (also changed the ``np.gradient`` to work for 1D). 

Does it make sense for ``iabel_hansenlaw`` to also work with 1D images? It will require a little bit of extra hacking, since the up/down quadrants concept doesn't apply. Right now, ``iabel_hansenlaw`` does not produce the right result when passed a 1D array due to the quadrants, so I made it throw an error if it gets a 1D array.

### DIFF
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -72,9 +72,10 @@ def iabel_hansenlaw_transform(IM):
                  +--------+      --------+
                               
     """
+    IM   = np.atleast_2d(IM)
     N    = np.shape(IM)  # length of pixel row, note in this case N = n/2
-    AImg = np.zeros(N)   # the inverse Abel transformed pixel row
-    
+    AImg = np.zeros(N)   # the inverse Abel transformed image
+
     nrows,ncols = N      # number of rows, number of columns
 
     # constants listed in Table 1.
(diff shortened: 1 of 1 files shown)


stggh:
(1)`I am actually not sure if this line is even necessary. The center few pixels are noisy in either case, and I cannot really tell if there is a difference by eye. @stggh, does this line serve a specific purpose?`

``` AImg[:,ncols-1] = AImg[:,ncols-2]``` is to provide a similar (sensible) value for the innermost pixel, otherwise there would be a vertical zero pixel line visible through the centre of the image.

(2) Parameter comment ~line 66, may still refer to `one row' of the image

(3) Agreed, 1-d images may go straight to ```iabel_hansenlaw_transform```

(4) I will reset my fork once the updates are merged and look at implementing ```vertical/horizontal_symmetry``` parameters

DanHickstein:
1. Oh, yes, you are completely correct. I somehow fooled myself about this earlier. Were you mentioning earlier that there was a better long-term solution for this? 

2. Fixed mention of "one row".

Otherwise, are we ready to merge? @rth, should we merge this first, and then your other PR?

stggh:
1. `Better`, only in relation to a possible use of a non-uniform grid &Delta; for evaluating the last pixel of an odd size image. However, the HL algorithm stops at n=N-2 in any case, so nothing needs to be done.

DanHickstein:
Oh, it just stops at ``n=N-2``? Interesting. Seems like cheating! :)

PR #52, opened 2015-12-10
Direct iabel/fabel implementation

rth:
This PR contains a implementation for forward and inverse Abel transforms using direct numerical integrations, as discussed in issue #23 . It was moved from [`hedp.math.abel`](https://github.com/luli/hedp/blob/master/hedp/math/abel.py). 

I wrote this a few years ago and should have probably checked other more efficient methods back then. Here, the integration is carried out numerically in Cython,
  - with the Simpson rule     (both with `naive=True` and `naive=False` options) 
  - with an analytical correction in the pixel where singularity of the weighting function `1/sqrt(y^2 - r^2)` occurs (if `naive=False` ). In this case, I assume that the data is piecewise linear in that pixel, and this  correction was pre-calculated with `sympy`.

The results are reasonable enough with the correction ('direct' implementation, `naive=False`), except near the origin in the inverse abel implementation, although the  numerical solution does converge to the analytical solution for higher spatial resolutions (I should compare if this behavior is similar to HansenLaw).

 As to the 'direct-naive' implementation, `naive=True`, it is the simplest thing one could do when approaching Abel transform (and for a matter of fact I know a few people who did use this), however the results are off by ~30%, and I put it here only for illustration purposes. It should not be used in practical applications (maybe I should even remove it from the example, and print warnings when this is activated).  See the two plots below, from `examples/abel_direct_gaussian.py`,

figure_1

It is actually very helpful to have a baseline comparison with other methods, as I now realize that despite it's implementation in Cython to reach C like speeds (spend quite a bit of time optimizing this), it is still quite slow and is 2 to 5 times slower than HansenLaw (although I'm not sure that the benchmark is valid in this case as I will explain later). 

Indeed, this implementation assumes only cylindrical symmetry, and operates on 1 quadrant, where the input array is of the shape `(z_axis, r_axis)` and the `r_axis` goes from `0` to `r_max`, so in this sens it is more close to `iabel_hansenlaw_transform` (except that it works also with 2D arrays) than to `iabel_hansenlaw`. Some normalization to be consistent with other implementation would be necessary in the future (e.g. add something similar to what is done with quadrants in `iabel_hansenlaw`, except in cylindrical and not spherical symmetry.)

This also adds the Cython as a dependency to PyAbel, since I cannot do without it in this implementation for performance reasons (the code cannot be easily vectorized, or at least I couldn't come up with a way to do that). Besides, I think that having access to Cython can be potentially useful for other implementations in the future to optimize some critical parts of the code. Cython converts the `.pyx` to `.c`, which then need to be compiled with a C compiler. On Linux and Mac, it is not an issue, since one should be present by default on the system. On Windows,  when using conda, mingw (gcc) can be installed with `conda install libpython`. In any case if for some reason the compiler cannot be found, or there was an compilation error for some reason, the `setup.py` will produce a warning similar to,

     abel/lib/direct.c:1:2: error: #error Do not use this file, it is the result of a failed Cython compilation.
     #error Do not use this file, it is the result of a failed Cython compilation.
      ^
    **************************************************
    WARNING: Cython extensions failed to build (used in abel.direct).
    Typical reasons for this problem are:
      - complier is not found
      - issues using mingw compiler on Windows 64bit (experimental support)
    This only means that the abel.direct implementation will not be available.
    ************************************************** 
and continue with with the installation. So this means that the user won't be able to import the `abel.direct` implementation but the rest of PyAbel will work as before. 

To give some context, personally now I'm mostly using the forward Abel transform in my research to post-process 2D cylindrical fluid dynamics simulation (`r = [0, r_max]`). The typical image size is of the order of `(2000, 1500)` and since the simulation produces multiple frames, performance is important. Now I realize that this implementation may not be so fast, and I'll try to look into Fourier-Hankel next for comparison. If BASEX (without the up/down symmetry) could be somehow used to get a forward transform (without too much effort), I would be interested too..

~~**Edit**: the Windows build fails before the install phase because of a [`input line too long`](https://stackoverflow.com/questions/16821784/input-line-is-too-long-error-in-bat-file) error, it doesn't look like it's related to this PR, maybe I mis-configured something in `appveyor.yml` I will check it up.~~

**Edit2**: Not sure what happened with the error above, but it went away when I restarted the Appveyor build. There is another issue however related to Cython when using Anaconda on Windows 64bit, so the situation is as follows,
  - Linux, Mac: will work out of the box as long as cython dependency is installed
  - Windows, Anaconda, PY2 32/64 bits and PY3 32bit : will work as long as cython + libpython is installed with conda
  - Windows, Anaconda, PY3 64 bit: Cython extensions are installed without errors, but raise an exception when imported. This is a know issue in Anaconda related to the 64 bit mingw compiler,
          https://groups.google.com/a/continuum.io/forum/#!topic/anaconda/3ES7VyW4t3I
    So in this case, users won't be able to use `abel.direct` but all the rest of PyAbel will install and work fine. Hopefully somebody from the anaconda or cython team will fix this issue in the future.
The test suite also passes in this case on Appveyor, although tests corresponding to `abel.direct` are skipped. 
  - Windows, other distributions than Anaconda: should work


DanHickstein:
First of all - this is awesome! Having a method for the numerical forward abel transform is a crucial part of the PyAbel package.

I'm not wild about adding the Cython dependency, both because I don't want to add more dependencies, especially those that can pose problems during installation, and because I think that the non-vectorized code is much harder to read. But, after trying to write a vectorized version of the simple Abel transform and seeing super slooooooooow performance, I'm starting to come around to your viewpoint.

I can write the transform straight from the forward [Abel integral on wikipedia](https://en.wikipedia.org/wiki/Abel_transform) without using any loops, and I thought that this would mean that it was really fast. I don't really understand why the following is so slow. I realize that the 3D array is pretty huge, but in my (albeit limited) Python experience, if you can eliminate loops, things will almost always be pretty fast. 

```python
from abel.analytical import GaussianAnalytical
import numpy as np
import scipy.integrate

r = np.linspace(0,10,300)
height = 300 
sigma = 3
A0=1

orig = A0 * np.exp(-r**2/sigma**2)  # analytical gaussian
proj = sigma * np.sqrt(np.pi) * A0*np.exp(-r**2/sigma**2) # analytical projection

size = np.shape(r)[0]

# in the following, 
# first dimension is z
# second dimention is r
# third dimension is y

ORIG = np.zeros((height,size,size)) 
ORIG[:,None,None] = orig # Assign f(r) values for all z- and y-positions

Z,R,Y = np.meshgrid(np.ones(height),r,r,indexing='ij') # set up grids

dr = r[1]-r[0]

# set up the integral: https://en.wikipedia.org/wiki/Abel_transform
P = 2 * ORIG * R * dr/ np.sqrt(R**2-Y**2)
P[R<=Y] = 0   # Set the limits of r from y -> inf

F = np.sum(P,axis=1) # take the integral
result = np.mean(F,axis=0) # just plot a lineout

result = result/3.6 # we still need an empirical scaling factor for some reason...

plt.plot(r,orig,label='Original')
plt.plot(r,proj,label='Projection (analytical)',color='r',ls='dashed',lw=2,)
plt.plot(r,result,label='Projection (numerical)',color='g')

plt.legend(loc=0)
plt.show()
```

rth:
Thanks for the feedback!

Regarding Cython, I agree that it would be preferable not add other dependencies and that vectorized code is easier to read, but the thing is that scientific code in optimized C/Fortran/Cython is still heard to compete with using vectorized numpy operations (e.g. in core scipy functions etc). In most other cases, performance is not that critical and that's why vectorized numpy code works just fine. In addition, you kind of pay the convenience of numpy vectorized operations with RAM: for instance, in Abel we could need to create a 3D array `(n, n, n)`, if `n=1000`  that's already `8 GB` or RAM, which means that on my computer with 16GB or RAM I wouldn't  be able to run this transform for `1500x1500` images.

It's interesting, the vectorized example you tried can be actually speed up a bit with numpy broadcasting,

    from abel.analytical import GaussianAnalytical
    import numpy as np
    import scipy.integrate
    from abel.lib.direct import _cabel_integrate
    from abel.direct import fabel_direct

    np.seterr(all='ignore')

    N = 300
    sigma = 3
    A0 = 1
    r_max = 10

    ref = GaussianAnalytical(N, r_max, sigma, A0, symmetric=False)

    orig1d = ref.func
    orig2d = np.tile(ref.func, (N, 1))


    def fabel_v1(orig, r, dr, N, intfunc=np.sum):
        Z,R,Y = np.meshgrid(np.ones(N),r,r,indexing='ij') # set up grids
        # set up the integral: https://en.wikipedia.org/wiki/Abel_transform
        P = 2 * orig * R * dr/ np.sqrt(R**2-Y**2)
        P[R<=Y] = 0   # Set the limits of r from y -> inf
        return intfunc(P,axis=1)

    def fabel_v2(orig, r, dr, N, intfunc=np.sum):
        R, Y = np.meshgrid(r,r,indexing='ij')
        P = 2 * orig * R * dr/ np.sqrt(R**2-Y**2)[None,...]
        P[:,R<=Y] = 0
        return intfunc(P,axis=1)

where `fabel_v2` is 3 times faster than the original vectorized version `fabel_v1`,

    %timeit fabel_v1(orig1d[:,None, None], ref.r, ref.dr, N)
    %timeit fabel_v2(orig1d[:,None, None], ref.r, ref.dr, N)
    # 1 loops, best of 3: 1.27 s per loop
    # 1 loops, best of 3: 394 ms per loop

but if you compare it with the optimized Cython version it's still 25 times slower (using the equivalent `scipy.integrate.simps` integration here):

    %timeit fabel_v2(orig1d[:,None, None], ref.r, ref.dr, N, intfunc=scipy.integrate.simps)
    %timeit fabel_direct(orig2d, r=ref.r)
    # 1 loops, best of 3: 859 ms per loop
    # 10 loops, best of 3: 34.8 ms per loop

So I'm not saying that Cython is always much faster, but in this particular case it is, as it allows to a) avoid unnecessary operations for `r<=y` array elements b) spends less time for memory copy of huge arrays in RAM. We could add the vectorized version alongside the Cython version but I don't think it would be very useful as users would rapidly run out of RAM for large image sizes, and it's more code to maintain/test.

With respect to deployment, what is done for some packages is to upload to PyPi a wheel package with pre-compiled Cython extensions for different OS. This removes the Cython dependence, but it's a more work for the developers, so I don't think that it makes sens to do it in our case.

In conclusion, I think the best approach to simply make Cython an optional dependency, so that `abel.direct` will not be usable if it is not present, but the rest of PyAbel will work fine. This would not be a big deal as the forward transform should also computable with the Fourrier Hankel implementation (issue #24 ), although it looks like the latter would need the `hankel` package to work (i.e. another optional dependency). Would this be acceptable?

DanHickstein:
@rth Thanks for the analysis!

I don't quite understand how numpy broadcasting works in this case, but it seems that I can get a similar speedup by looping over the rows instead of trying to vectorize the entire thing. In fact, I can also loop over ``y`` instead of vectorizing it. So, I have three cases: ``zero_loop_abel``,  ``one_loop_abel``,  and  ``two_loop_abel``. Which is faster depends on the size of the image. In a small square image, say 200x200, the zero-loop is the fastest. For larger square size images, say 500x500, the one-loop method is the fastest. In the event that we just have a few rows and lots of r-values, say 2x5000, the two-loop method is the fastest. (I pasted the code at the bottom, because it's a little long.)

Also, I figured out the problem with the absolute scaling. It was just a typo and I was accidentally orienting the data along the y-dimension instead of the r-dimension. So, now there is good agreement between the numerical and analytical results, but it takes a very fine grid step. 10,000 r-values are required before it looks really good. Also, there is a small difference between the two-loop result and the other two results, and I'm not quit sure why.

In the end, I think that I am convinced that your Cython method is probably the best way to ensure good performance for the direct numerical transforms. I was thinking that it may also be nice to include a very short and readable pure-python implementation of the direct numerical transforms. This could be used if Cython is not available, but more importantly, it will provide users with an intuitive understanding of the Abel and inverse Abel transform, a feature that is lacking in all of the other transform methods, which introduce additional complexities. This could be called the "simple" method, and could be located at ``abel.abel_simple``. It might be really slow for large images, but depending on which algorithm we select, it's quite usable for large 1D datasets. 

It also serves as a nice benchmark for the rest of the transform algorithms. I.e., we can point out that all of the rest of our methods are orders of magnitude faster than this implementation.

As an aside, I wonder if there are even better pure-python methods of tackling this problem. Maybe it's worth a Stack Overflow post to see if some code ninjas can think of something that has not occurred to us.

So, I think that we should go ahead with this PR. Can you add a few more in-line comments to the ``.pyx`` file? For example, say which loop is over the rows, and which is over y, etc. Otherwise, it seems like it's good-to-go.

Code for simple, pure python forward abel transforms:

```python
from abel.analytical import GaussianAnalytical
import numpy as np
import scipy.integrate
import time
import warnings
warnings.filterwarnings("ignore")

r = np.linspace(0,10,800)
height = 10
sigma  = 3
A0     = 1

orig = A0 * np.exp(-r**2/sigma**2)  # analytical gaussian
proj = sigma * np.sqrt(np.pi) * A0*np.exp(-r**2/sigma**2) # analytical projection

orig2d = np.tile(orig, (height, 1))

def zero_loop_abel(IM,r):
    # in the following,
    # first dimension is z
    # second dimention is r
    # third dimension is y
    height,size = np.shape(IM)

    ORIG = np.zeros((height,size,size))
    ORIG = IM[:,:,None]

    plt.show()
    Z,R,Y = np.meshgrid(np.ones(height),r,r,indexing='ij') # set up grids

    dr = r[1]-r[0]

    # set up the integral: https://en.wikipedia.org/wiki/Abel_transform
    P = 2 * ORIG * R * dr/ np.sqrt(R**2-Y**2)
    P[R<=Y] = 0   # Set the limits of r from y -> inf

    F = scipy.integrate.simps(P,axis=1) # take the integral
    return F
    

def one_loop_abel(IM,r):
    size = np.shape(r)[0]
    dr = r[1]-r[0] # this assumes even spacing of r, but this is not strictly necessary
    res = np.zeros((height,size))

    for index,row in enumerate(IM): # loop over rows
        ORIG = row[:,None]
        R,Y = np.meshgrid(r,r,indexing='ij') # set up grids
        
        P = 2. * ORIG * R * dr / np.sqrt(R**2-Y**2) # set up the integral
    
        P[R<=Y] = 0   # set limits of integral
        F = scipy.integrate.simps(P,axis=0) # take the integral    
        res[index] = F
        
    return res
    

def two_loop_abel(IM,r):
    # start of Abel transform: 
    y_list = r # Here we use the same r-spacing for y, but this does not need to be the case
    dr = r[1]-r[0] # this assumes even spacing of r, but this is not strictly necessary
    res = np.zeros(np.shape(IM))

    for i,row in enumerate(IM): # loops over rows
        for j,y in enumerate(y_list):  # loop over y-values of final image
            if y != np.max(r):
                ind = (r>y)
                P = 2 * row[ind] * r[ind] * dr / np.sqrt(r[ind]**2-y**2)
                F = scipy.integrate.simps(P) # take the integral
                res[i,j] = F
            else:
                res[i,j] = 0

    return res


t = time.time()
zero = zero_loop_abel(orig2d,r)
zero = np.mean(zero,axis=0)
print 'zero loops: %.2f sec'%(time.time()-t)

t = time.time()
one = one_loop_abel(orig2d,r)
one = np.mean(one,axis=0)
print 'one loop: %.2f sec'%(time.time()-t)

t = time.time()
two = two_loop_abel(orig2d,r)
two = np.mean(two,axis=0)
print 'two loop: %.2f sec'%(time.time()-t)

plt.plot(r,orig,label='Original',color='k',alpha=0.5)
plt.plot(r,proj,label='Projection (analytical)',color='b',lw=2)
plt.plot(r,zero,label='Projection (zero-loop)',color='r',ls='dashed',lw=2)
plt.plot(r,one, label='Projection (one-loop)', color='m',ls='solid', lw=1)
plt.plot(r,two, label='Projection (two-loop)', color='r',ls='solid', lw=1)

plt.legend(loc=0)
plt.show()

```

DanHickstein:
Oh, are you including an option to fall back to pure-python if Cython is not present? You have awesome skills :)

rth:
> I was thinking that it may also be nice to include a very short and readable pure-python implementation of the direct numerical transforms. 

I agree that this would be nice, and I think that your `one_loop_abel` loop is the way to go. It can actually be speed up by another factor of 2 by factorizing the `1./sqrt(R**2-Y**2)` calculation outside of the loop,

    def one_loop_abel_v2(IM,r):
        size = np.shape(r)[0]
        res = np.zeros((height,size))
        R, Y = np.meshgrid(r,r,indexing='ij') # set up grids
        I_sqrt = np.zeros(R.shape)
        mask = R<Y
        I_sqrt[mask] = 1./np.sqrt(R**2-Y**2)[mask]
       
        for index,row in enumerate(IM): # loop over rows
            P = 2. * row[:, None] * R * I_sqrt # set up the integral
    
            F = scipy.integrate.simps(P, r, axis=0) # take the integral    
            res[index] = F
    
        return res

and it is then extremely close to what I do in Cython (at least for the first step of the calculation). So I have created a [`_pyabel_direct_integral`](https://github.com/PyAbel/PyAbel/pull/52/files#diff-58ebe14f53d4b01444a9ae9b4bc5432bR171) which is based on `one_loop_abel_v2` but 
   - generalizes it so the same integral can be used both for the direct and inverse transform
   - optionally adds a correction (what happened before for `naive=False` parameter, now happens for `correction=True`)
This function is then supposed to be the exact equivalent of the Cython function [`_cabel_direct_integral`](https://github.com/PyAbel/PyAbel/pull/52/files#diff-155c2187156fad19b103d739fb5de382R11). So one or the other can be used from `fabel_direct` and `iabel_direct` by passing a `backend='C'` or `backend='Python'` parameter. Furthermore if Cython extensions are not installed, we can automatically fall back to the Python implementation. 

Now, having theses 2 implementations is interesting as it actually allowed me to find several bugs in my Cython code, so code review is definitely useful. In particular,

1. It does not fully support non uniform sampled data. So in this case, we also fall back to the Python backend as it's easier to handle there;
2. The Cython and Python backend do not produce exactly the same output, because my Cython code has a bug in the Simpson rule integration (the second step where the correction is applied is identical). In particular, the Simson rule is only valid for an even `n` while I apply the formula in both odd and even cases. What is ridiculous however is that with this bug, the results are better than without it, both for forward and inverse transforms. Here is for instance a plot of the absolute error between the analytical Abel transform of a Gaussian and the numerical reconstruction  with `abel.direct.fabel_direct`, using [this script]()
test_convergence_v1
So not only the error for `abel.direct.fabel_direct(...,  correction=1, backend='C')` is inherently smaller, but it also converges to zero faster with `N`  (which means that  we can correctly reconstruct a Gaussian with a lesser number of points).

I am very perplexed as to why this happens, maybe there is a second error in the code that cancels outs somewhere, but in any case `abel.direct` implementation appears to work fine. Although I would need to look in detail why this is happening.  

All of the changes described in this comment were added in previous commits. I also commented more the Cython code and removed duplicate parts. So as far as I am concerned we could merge this, and maybe investigate this issues later on as they are not critical but rather have an impact as to how fast the reconstruction converges to the correct solution.

DanHickstein:
Great work! I like the idea of having the pure-python implementation as a fallback (if Cython is not present), for readability, and for catching bugs in the Cython code. 

Is the plot that you showed the current behavior, or did you fix the bug in the Cython implementation? As I recall, the numerical version with the correction tends to be a little bit too high near the center. If the bug somehow made those values artificially smaller, it could end up making the agreement better. It would be nice to have more-or-less identical results from the Cython versus Python implementations if we are going to wrap them up in the same function, which seem implies that users can expect identical results. 

Still, it's okay with me if you want to merge and open a separate issue to look into the source of the difference. Again, great work with this. It's so nice to finally have a numerical forward transform. You put the "Abel" in PyAbel!

rth:
@DanHickstein  The plot is the current behavior. I agree that Cython versus Python implementations should produce identical results, as otherwise this will be confusing for users. Yeah, I'll do another PR later to investigate and fix this later.

DanHickstein:
Sounds good, you are good to merge as far as I'm concerned.

Issue #51, opened 2015-12-08
Compatibility with numpy 1.10.1


DanHickstein:
Numpy 1.10 apparently has stricter rules for casting floats as integers:
https://github.com/scikit-learn/scikit-learn/issues/5397

This causes an error with PyAbel as well, specifically for the ``example_hansenlaw.py``.

I think that we just need to specifically cast the numbers to the correct types, but I think I recall there is something weird going on with a half-integer origin in the case of the hansen-law transform? @stggh 

Here is the output from trying to run the example:

```
Daniels-MacBook-Pro-3:examples danhickstein$ python example_hansenlaw.py
Loading data/O2-ANU1024.txt.bz2
image size 1024x1024
Performing Hansen and Law inverse Abel transform:
HL: Calculating inverse Abel transform:  image size 1024x1024
HL: Individual quadrants
HL: Calculating inverse Abel transform ...
0.29 seconds
Generating speed distribution ...
Traceback (most recent call last):
  File "example_hansenlaw.py", line 52, in <module>
    recon, speeds = iabel_hansenlaw (im,quad=(False,False,False,False),verbose=True)
  File "/Users/danhickstein/Documents/PyAbel-master/abel/hansenlaw.py", line 210, in iabel_hansenlaw
    speeds = calculate_speeds(recon,origin=image_centre)
  File "/Users/danhickstein/Documents/PyAbel-master/abel/tools.py", line 24, in calculate_speeds
    polarIM, ri, thetai = reproject_image_into_polar(IM,origin)
  File "/Users/danhickstein/Documents/PyAbel-master/abel/tools.py", line 145, in reproject_image_into_polar
    x, y = index_coords(data, origin=origin)
  File "/Users/danhickstein/Documents/PyAbel-master/abel/tools.py", line 179, in index_coords
    x -= origin_x
TypeError: Cannot cast ufunc subtract output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

Thanks to @ellisje624 for pointing this out.

stggh:
```np.meshgrid(np.arange(nx), np.arange(ny))``` returns integer numpy.arrays for nx, ny integer. 
A possible solution is to convert to float:
```
178,180c178
<     x, y = np.meshgrid(np.arange(nx), np.arange(ny))  # int numpy.array
<     x.astype(float)
<     y.astype(float)
---
>     x, y = np.meshgrid(np.arange(nx), np.arange(ny))
```

[...]

PR #53, opened 2015-12-12
Numpy compatibility

DanHickstein:
This allows compatibility with the latest version of numpy (1.10.1), as discussed in https://github.com/PyAbel/PyAbel/issues/51. Only change was to explicitly cast ``nx`` and ``ny`` to floats in ``index_coords``. I think that they were being automatically being cast to floats previously, so the behavior should be the same.

### DIFF
--- a/abel/tests/test_tools.py
+++ b/abel/tests/test_tools.py
@@ -67,8 +67,8 @@ def test_centering_function():
         assert_equal( is_symmetric(res), True,\
             'Validating the centering function for ni={}, n_c={}, n={}'.format(ni, n_c, n))
 
- def test_speeds_non_integer_center():  
-     # ensures that the rest speeds function can work with a non-integer center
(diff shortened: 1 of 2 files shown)


rth:
This looks good!

Also, I think it could be a good practice when we encounter a bug to add a small unit test that fails with this bug, and then passes with the fix. For instance, here, @DanHickstein would you mind adding 

     def test_speeds_non_integer_center():  
         n = 101
         IM = np.random.randn(n, n)
         calculate_speeds(IM, origin=(50.5, 50.5))

to `abel/tests/test_tools.py` (or something similar), before merging this ? This would make sure that if somebody were to edit the speed generation function at some later time, this bug would not be reintroduced again by accident...

DanHickstein:
Good idea to add the test. It's in there now.

PR #57, opened 2015-12-16
Tweak setup.py for future PyPi upload

DanHickstein:
I made a few small changes to the setup.py file. Most importantly,
changed the author from myself to “The PyAbel Team”, to acknowledge the
hard work by all of the current (and hopefully future!) contributors to
the project.

Also, changed the description to mention that we also do the forward
Abel transform. Finally, made it clear that the lack of Cython would
only doom the C implementation of abel.direct, since we now have the
pure-Python fallback, right @rth?


rth:
Everything looks great. We can merge this now, or if you intend to upload the package to PyPi, we can also use this PR to add a few more [arguments to `setup`](https://python-packaging-user-guide.readthedocs.org/en/latest/distributing/#setup-args) asked by PyPi, I think in particular `licence` and `classifiers` (not completely sure).

DanHickstein:
Okay, I added the license, url, and classifiers field. I wasn't quite sure what to put for some of these, so please review my choices:
```python 
license='GPLv3'
```
Looking on http://choosealicense.com/, GPL version 3 seems like the best license to me. It only allow open source use, which seems nice. The other option is the MIT license, which allows for all uses, including closed-source commercial uses. I'd like to see PyAbel widely used, but also like to support open source. Happy to hear thoughts from other contributors as well (@DhrubajyotiDas @stggh). Currently we have the GPLv2 license in the repo. According to the above link, "V3 is similar to V2, but further restricts use in hardware that forbids software alterations." So, V3 seems better, I suppose.

```python
classifiers=[
# How mature is this project? Common values are
#   3 - Alpha
#   4 - Beta
#   5 - Production/Stable
'Development Status :: 3 - Alpha',

# Indicate who your project is intended for
'Intended Audience :: Science/Research',
'Intended Audience :: Developers',
'Topic :: Scientific/Engineering :: Mathematics',
'Topic :: Scientific/Engineering :: Physics',
'Topic :: Scientific/Engineering :: Medical Science Apps.',
'Topic :: Software Development :: Libraries :: Python Modules'

# Pick your license as you wish (should match "license" above)
'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
```
The options for the ``Intended Audience`` and ``Topic`` are listed here. I took a few guesses. Let me know if you have suggestions. I included "Medical Science Apps." since I thought that perhaps the Abel transform would be used by people dealing with reconstructing objects from X-ray images. 

```python
# Specify the Python versions you support here. In particular, ensure
# that you indicate whether you support Python 2, Python 3 or both.
'Programming Language :: Python :: 2',
'Programming Language :: Python :: 2.7',
'Programming Language :: Python :: 3',
'Programming Language :: Python :: 3.3',
'Programming Language :: Python :: 3.4',
'Programming Language :: Python :: 3.5',
],
```
I included just the python versions that we have on Travis CI. So, we are saying that we officially support Python 2.7, 3.3, 3.4, 3.5, but not 2.6 or anything else. Seems reasonable enough.

rth:
Regarding the Python version, yes I think we can say that we support version that are listed in Travis CI setup file. Python 2.6 will probably produce errors  (e.g. it does not support dictionary generators); 3.1, 3.2 might work, but again since we don't tests for them it's hard to tell. So I think that we can say that Python 2.7 & 3.3-3.5 are officially supported.  The `classifiers` look good too.

Personally, I don't know much about licensing: both GPL v2/v3 or MIT type of license sound good. One point to consider I suppose what would happen if some other project would want to use PyAbel, and decided to distribute binaries of the executable they provide (including PyAbel). In the case of a GPL license, this would only be possible if the project itself is under an open-source license (GPL/MIT/etc).   In the case of an MIT license, they could do it  even if their project is not open source. In the latter case, there still could be contributions back to PyAbel, it just means that they are allowed to distribute PyAbel included in their project. I don't really have strong convictions one way or the other, probably a bit more in favor of the MIT license, but we can  keep GPL v3 for now and see if that's an issue later. @DhrubajyotiDas and @stggh would this work for you?

The only external project I'm aware of that could potentially use PyAbel in Python scripting is [Neutrino](https://github.com/aflux/neutrino) developed by  @aflux and @iltommi, but it has a GPL license, so it's not an issue either way.

DanHickstein:
I think you're right. We should let people incorporate PyAbel into binaries. And the more people using it the better. MIT license it is.

stggh:
My preference is for GPLv3, in support of true open source software. However, I have no great objection your decision to expand the application of PyAbel by using the MIT license.

DhrubajyotiDas:
Personally, I am not very knowledgeable about licenses either, but in the interests of building consensus, I support applying the MIT license as well.

DanHickstein:
Okay, I'll go ahead and merge this with the MIT license, but we can continue to discuss in the future.

Issue #61, opened 2015-12-28
Implement 3-point Abel inversion technique


DhrubajyotiDas:
[Dasch, 1992](https://www.osapublishing.org/ao/abstract.cfm?uri=ao-31-8-1146) (Applied Optics, Vol 31, No 8, March 1992, Pg 1146-1152) discusses several numerical techniques for performing Abel inversions of cylindrically symmetric line-of-sight images to extract the underlying radial distributions (see discussion in #56). One technique discussed in the paper is the 'Abel inversion using three-point interpolation' or '3-pt Abel' technique; this technique is fairly popular in the combustion diagnostics literature. 

It would be useful to incorporate this technique in PyAbel. 

(more details coming in 2016!)

DanHickstein:
Awesome! I'm excited to see how this method works!

DhrubajyotiDas:
Happy New Year!

Prof. Marshall Long's group at Yale uses [the linked matlab script](https://gist.github.com/DhrubajyotiDas/16d9381a78352f93bdfe) to implement the 3-point Abel deconvolution as per Dasch 1992.

[I've ported this to python](https://gist.github.com/DhrubajyotiDas/0159ccbc075821632e8c) and run some basic tests. This inversion is fairly quick and works on a half-image. As usual, the raw data must have cylindrical symmetry, and the output rows are independent (transform works on a row-by-row basis).

![dashc 3pt abel](https://cloud.githubusercontent.com/assets/6530840/12072553/dfbdbec8-b0b1-11e5-8e84-410800c3417a.png)

Future works include:
* Timed runs
* Tests with noisy raw data
* Tests with experimental raw data
* Comparison to BASEX and other approaches

I have not filed a PR for this because I'd like @DanHickstein, @rth, and @stggh's input on how to package this inversion with the proper wrapper functions, etc.

stggh:
@DhrubajyotiDas this algorithm is pretty neat.  Thanks, for taking the time to make it available. 

[...]

PR #64, opened 2016-01-06
Implementing Three-point Abel - another inverse Abel transformation algorithm

DhrubajyotiDas:
As discussed in #61, this pull request adds the 3-point Abel inversion technique to the repertoire of inversion algorithms available under PyAbel.

This technique was developed by [Dasch, 1992](https://www.osapublishing.org/ao/abstract.cfm?uri=ao-31-8-1146) (Applied Optics, Vol 31, No 8, March 1992, Pg 1146-1152). This algorithm exploits the observation that the value of the Abel inverted data at any radial position `r` is primarily determined from changes in the projection data in the neighborhood of `r`. 
From the paper: *the Abel integral is broken into segments around each point r<sub>j</sub>. In the neighborhood of each point,* the line-of-sight data *is expanded as a quadratic function of r - r<sub>j</sub>. Then each integral segment is integrated analytically. This method is called the 3-point Abel deconvolution since the quadratic expansion uses only the projections adjacent to r<sub>j</sub>.*

The original algorithm (with correction of typographical errors) was [implemented in Matlab](https://gist.github.com/DhrubajyotiDas/16d9381a78352f93bdfe) by Prof. Marshall Long's group at Yale. This PR is mostly a direct port of that code without any attempted optimizations (which are coming soon *hopefully*). 

In addition to the transform code, a new unit test checking the `three_point` transform against an analytical Gaussian is also included. The agreement between the two is essentially identical. 

    import numpy as np
    from abel.three_point import iabel_three_point

    n = 101
    r_max = 4
    r = np.linspace(-1*r_max, r_max, n)
    dr = r[1]-r[0]

    fig, axes = plt.subplots(ncols = 2)
    fig.set_size_inches(8,4)

    axes[0].set_xlabel("Lateral position, x")
    axes[0].set_ylabel("F(x)")
    axes[0].set_title("Original LOS signal")
    axes[1].set_xlabel("Radial position, r")
    axes[1].set_ylabel("f(r)")
    axes[1].set_title("Inverted radial signal")

    Mat = np.sqrt(np.pi)*np.exp(-1*r*r)
    AnalyticAbelMat = np.exp(-1*r*r)
    DaschAbelMat = iabel_three_point(Mat, 50, dr)

    axes[0].plot(r, Mat, 'r', label = r'$\sqrt{\pi}e^{-r^2}$')
    axes[0].legend()

    axes[1].plot(r, AnalyticAbelMat, 'k', lw = 1.5, alpha = 0.5, label = r'$e^{-r^2}$')
    axes[1].plot(r, DaschAbelMat, 'r--', lw = 1.5, label = '3-pt Abel')

    box = axes[1].get_position()
    axes[1].set_position([box.x0, box.y0, box.width * 0.8, box.height])
    axes[1].legend(loc='center left', bbox_to_anchor=(1, 0.5))
    fig.tight_layout()

dasch 3pt abel

### DIFF
--- a/abel/three_point.py
+++ b/abel/three_point.py
@@ -26,13 +26,7 @@ def iabel_three_point_transform(IM, basis_dir='.', verbose=False):
     inv_IM = np.zeros_like(IM)
 
     for i, P in enumerate(IM):
-        F = np.zeros(col)
-        for j,k in product(range(col), range(col)):
-            # Deconvoluting projection data P to give field distribution F
-            # See Eq (1) in Dasch 1992.
-            F[j] += D[j,k] * P[k] 
-
-        inv_IM[i] = F
+        inv_IM[i] = np.dot(D, P)
 
     return inv_IM
 
--- a/abel/three_point.py
+++ b/abel/three_point.py
@@ -83,3 +83,43 @@ def OP1(i,j):
         raise(ValueError)
         
     return I1
+
+def iabel_three_point(data, center, dr = 1.0):  
+    """ This function splits the image into two halves, 
+        sends each half to iabel_three_point_transform(), 
+        stitches the output back together,
+        and returns the full transform to the user.
+
+    Parameters:
+    -----------
+        - data:   NxM numpy array
+                  The raw data is presumed to be symmetric about the vertical axis. 
+        - center: * integer - 
+                    The location of the symmetry axis (center column index of the image). 
+                  * tuple (x,y) - 
+                    The center of the image in (x,y) format.
+        - dr:     float - 
+                  Size of one pixel in the radial direction
+    """
+
+    # sanity checks for center
+
+    # cut data in half
+    # each half has the center column at one edge
+    left_half, right_half = data[:,0:center+1], data[:,center:]
+
+    # mirror left half
+    left_half = np.fliplr(left_half)
+
+    # transform both halves
+    inv_left = iabel_three_point_transform(left_half)
+    inv_right = iabel_three_point_transform(right_half)
+
+    # undo mirroring of left half
+    inv_left = np.fliplr(inv_left)
+
+    # stitch both halves back together
+    # (extra) center column is excluded from left half
+    inv_IM = np.hstack((inv_left[:,:-1], inv_right))
+
+    return inv_IM
--- a/abel/three_point.py
+++ b/abel/three_point.py
@@ -20,7 +20,7 @@ def iabel_three_point_transform(IM):
     """
     IM = np.atleast_2d(IM)
     row, col = IM.shape
-    D = np.zeros((col, col))
+    D = np.zeros((col, col)) # Empty D_ij matrix
 
     for i,j in product(range(col), range(col)):
         D[i,j] = OP_D(i,j)
@@ -30,7 +30,9 @@ def iabel_three_point_transform(IM):
     for i, P in enumerate(IM):
         F = np.zeros(col)
         for j,k in product(range(col), range(col)):
-            F[j] += D[j,k] * P[k]
+            # Deconvoluting projection data P to give field distribution F
+            # See Eq (1) in Dasch 1992.
+            F[j] += D[j,k] * P[k] 
 
         inv_IM[i] = F
 
@@ -38,8 +40,9 @@ def iabel_three_point_transform(IM):
 
 def OP_D(i,j):
     """
-    Calculate three-point abel inversion operator Di,j
-    The formula followed Dasch 1992 (Applied Optics) which contains several typos.
+    Calculates three-point abel inversion operator D_ij, 
+    following Eq (6) in Dasch 1992 (Applied Optics).
+    The original reference contains several typos.
     One correction is done in function OP1 following Karl Martin's PhD thesis
     See here: https://www.lib.utexas.edu/etd/d/2002/martinkm07836/martinkm07836.pdf
     """
@@ -60,7 +63,9 @@ def OP_D(i,j):
     return D
 
 def OP0(i,j):
-    
+    """
+    This is the I_ij(0) function in Dasch 1992, pg 1147, Eq (7)
+    """
     if j < i or (j == i and i == 0):
         I0 = 0
     elif j == i and i != 0:
@@ -74,6 +79,9 @@ def OP0(i,j):
     return I0
 
 def OP1(i,j):
+    """
+    This is the I_ij(1) function in Dasch 1992, pg 1147, Eq (7)
+    """
     if j < i:
         I1 = 0
     elif j == i:
--- a/abel/three_point.py
+++ b/abel/three_point.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 import numpy as np
 from itertools import product
 
-def iable_dasch_transform(IM):
+def iabel_three_point_transform(IM):
     """ Inverse Abel transformation using the algorithm of: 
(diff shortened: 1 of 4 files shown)


DanHickstein:
@DhrubajyotiDas, thanks for implementing this! We really have a nice comparison of different methods now. 

The agreement with the analytical solution is achieved without a scaling factor?

DhrubajyotiDas:
Yes, the agreement is without any arbitrary scaling factor (apart from the necessary `dr` scaling)

rth:
Looks great! I think we can merge this!

abel/three_point.py
```diff
@@ -0,0 +1,146 @@
...
+
+        inv_IM[i] = F
+
+    return inv_IM
+
+def OP_D(i,j):
+    """
+    Calculate three-point abel inversion operator Di,j
+    The formula followed Dasch 1992 (Applied Optics) which contains several typos.
+    One correction is done in function OP1 follow Martin's PhD thesis
```
DanHickstein: Can you provide a link to this thesis somewhere?

DanHickstein:
I think that this looks pretty good as far as I can tell (though I can't say that I really understand exactly what's going on). 

While it's fresh in your mind, can you provide some comments in the code to explain what parts are doing which steps? If you want to merge this and add the comments in a separate PR, that's also fine with me.

It sounds like you have fixed some typos that occurred in the original paper. We should write down exactly what equations we are using. A good place for this might be the wiki? I have been thinking that we can include some long-form general explanations here, including the background of the transform method, the math, and any helpful advice that the user should know about the transform. I started a page for the Three Point transform here: https://github.com/PyAbel/PyAbel/wiki/Three-point-transform

abel/three_point.py
DhrubajyotiDas: https://www.lib.utexas.edu/etd/d/2002/martinkm07836/martinkm07836.pdf

Karl Matthew Martin, UT Austin, 2002.

He mentions the typo issue on Pg 114, and gives the corrected version of some of the equations on the following pages (presumably from Dasch's 1992 errata). I've pasted an example below. Left: Martin's thesis, Right: Dasch's paper.

image

DhrubajyotiDas:
I am not too familiar with this algorithm either, but from what I can surmise, the real magic is happening in the`D` matrix, which is encoding the arithmetic that needs to be performed on the line-of-sight data to extract the inverted data. In general, such arithmetic cannot be encoded as a linear operator, but in the simplified case of approximating the raw projection data as a quadratic fit in the neighborhood of every data point, Dasch performed the algebra and solved the inverse Abel integral
image
analytically. The results of this algebra are Eqs 6 and 7 in the 1992 paper, and the actual math is discussed in Appendix A. 

For the purposes of the three_point technique, there were two typos in the original paper which were corrected by Dasch/Martin. I've added links to Martin's thesis in the documentation since it is the "correct" source of some of the equations used. I believe the paper also had other typos, but they were regarding other transform techniques outside the scope of this code. 

I think the wiki is a great idea. I'll add that to my to-do list!

stggh:
My interpretation is that the derivative function <math>P<sup>'</sup>(r) = dF/dy</math>, common to all inverse transform methods is estimated using a 3-point approximation (to the derivative), Eq. (A2). This results in 3 integrals (trapezoid sum) for the inverse Abel transform.

The `hansenlaw` and `direct` methods perform a crude, direct, evaluation of the derivative (@rth has his own derivative function, see `abel.math`). There may be something to be learned here.

DhrubajyotiDas:
@stggh Thanks for the clarification. I've added your interpretation to the wiki page on this technique.

stggh:
One minor downer, speed optimizations may be required, the O2- VMI test case fails to finish (within the realm of operator patience) - could be operator error

rth:
Regarding derivatives, it is true that the derivative functions for F could be common to most implementations that do not hard-code it (in particular `direct` and `hansenlaw`).   In fact, `direct`  implementation takes in a `gradient` function for that effect. By default it is set to `abel.math.derivative`  which is identical to `np.gradient` (i.e. a 2nd order centred differences) except that it also works for non uniform sampling. 

With the direct implementation I have also been using the [Savitzky-Golay](https://github.com/luli/hedp/blob/master/hedp/math/derivative.py#L62) differentiator for noisy data (that has been added to scipy of late). In  Dasch (Eq. A2) the derivative seems to be calculated as a average of centered 1st and 2d order differences, which might also have to do something with filtering noise (not sure).  BTW, if someone is interested,  here is some more reading about noise robust differentiators: 
http://www.holoborodko.com/pavel/numerical-methods/numerical-derivative/smooth-low-noise-differentiators/

If execution time is a problem, @DhrubajyotiDas would you mind just adding a line

     res_iabel['Three_point']['tr'].append(
                Timer(lambda: iabel_three_point(x)).timeit(number=NREPEAT)/NREPEAT)

 in [`abel.benchmark` at line 60](https://github.com/PyAbel/PyAbel/blob/master/abel/benchmark.py#L60) with the corresponding import of `iabel_three_point` at the top of the `__init__` function, so it is easier to benchmark the run time when optimizing this code?

DanHickstein:
@stggh, I like your idea of including all of the methods in a single example. If the three-point method is too slow to be usable for your 1000x1000 example image, you could downsample the image to provide a reasonable run-time. Obviously it won't be a completely accurate comparison of the methods...

@rth, very good point about the derivative being a common feature of many of the transform methods. I like how you've allowed the direct methods to take any differentiation function - cool idea! Can the user simple call ``iabel.direct(IM, derivaive=scipy.signal.savgol_filter(deriv=1))`` to obtain the transform with the Savitzky-Golay differentiation? This should be mentioned in the docstring, since it seems like a feature that many users will like.

DhrubajyotiDas:
@rth I've added the line you requested to `abel.benchmark`. I've also added the ability to cache the `D` matrix (called the "three_point operator matrix" for now...) so it can be reused if a batch of images of the same width are being processed. I will also be looking into vectorizing some of the loops to see if that makes things faster.

@stggh I can confirm that trying to generating a 1001x1001 operator matrix takes an unacceptable amount of time right now, so it's not operator error.

DhrubajyotiDas:
@stggh With the latest commit, you ought to be able to transform a 1001x1001 image fairly quickly.

stggh:
@DhrubajyotiDas    looks very good.  

examples/example_all_O2.py

Unfortunately, no free quadrant for the three_point transformed image (I will need to change example to smaller image segments)

all

direct inverse ...
                    0.4 sec
onion inverse ...
                   28.7 sec
hansen law inverse ...
                   1.0 sec
basex inverse ...
                   1.3 sec
three_point inverse ...
                   0.5 sec

stggh:
@rth I also like  Savitzky-Golay and agree the method is worth investigating further

DanHickstein:
@DhrubajyotiDas, great idea to add the caching. Is this PR ready to merge then?

rth:
@DanHickstein Yes, I think we should merge this before Steve's PR as there could potentially be some conflicts in `three_point.py`. 
For Savitzky-Golay, it could be something like,

     abel.iabel_direct(IM, dr=dr, 
             derivative=lambda x: scipy.signal.savgol_filter(x,
                                       window_length=9, polyorder=3,  deriv=1)/dr)
I will check it up further (the interface changed slightly when it was added to Scipy with respect to the option I copied from somewhere) and add an example to the docstring.

DhrubajyotiDas:
Yes, this PR is ready to be merged. Thanks for all your helpful comments!

DanHickstein:
Great work!

Issue #60, opened 2015-12-27
anisotropy beta parameter


stggh:
I have written some rough code to evaluate the anisotropy parameter &beta;  ```tools.anisotropy()```, in my fork. Example, ```examples/example_anisotropy()```

![example_anisotropy](https://cloud.githubusercontent.com/assets/10932229/12009554/809b71d6-acd0-11e5-8bdf-e006b0db20f8.png)

Notes/issues:
(1) Code is rough, needs further testing (by me), exception checks, unit tests.
(2) Uses ```scipy.optmize.curve_fit()```.  Currently ```center_image_by_slice()``` uses ```scipy.optimize.minimize()``` - I should make them consistent.
(3) ```tools.reproject_image_into_polar()``` returns polar angle reference to the horizontal direction (? I think). Our &theta; should be with respect to the vertical direction. I am not sure whether the correction should be applied within the ```reproject...``` function?
(3) Both ```calculate_speeds()``` and now ```anisotropy()``` reproject the image coordinates. This may be wasteful, when only one evaluation is required.

DanHickstein:
@stgg, great work! The anisotropy calculation is one of fundamental tools for many VMI spectroscopy labs, and it's a central feature of the BASEX program, so I think that many users will be looking for it. 

It would be helpful to see a definition of the anisotropy parameters, since I have never made use of them. But I am on vacation and don't have my books in front of me, and I can't find a solid definition in a paper. Do you have a textbook that has a good definition of the anisotropy parameters that you can take a photo of and post here?

What you are calculating is Beta_2, right? I know that people often like to look at higher order beta parameters, typically even ones. But in some cases odd beta parameters may be useful as well. We should probably have an option to include higher order terms.

Regarding #2: The use of ``scipy.optimize.curve_fit()`` has me a little bit worried. I'll bet that a nonlinear fit is the best way to extract beta from a nice dataset (like the beautiful ones that you collect). But nonlinear fits do not always converge, and they are very dependent on the starting conditions, often finding false minima. It seems like if we are just extracting one number from a curve, there should be a simpler way. ``scipy.optimize.minimize()`` at least lets you set the bounds on the anisotropy parameter. It seems like this should be beneficial. We should also return an estimate of the error of the anisotropy parameter. 

[...]

Issue #56, opened 2015-12-15
Implement "Onion Peeling" algorithm


DanHickstein:
Professor Eric Wells from Augustana University has kindly provided a Matlab implementation of the Onion Peeling (also known as "back projection") algorithm described in his recent paper: 

http://scitation.aip.org/content/aip/journal/rsi/85/11/10.1063/1.4899267 

The algorithm actually originates from this [1996 RSI paper by Bordas et al.](http://dx.doi.org/10.1063/1.1147044), so we should cite both papers in the code and the documentation. 

He said that we could port his algorithm to Python and incorporate it into the PyAbel package if we wish. I tested the Matlab implementation, and it provided sub-second transform times for 1000x1000 pixel images and the results seem to be high-quality. 

I uploaded the matlab code here: https://github.com/PyAbel/PyAbelLegacy/tree/master/onion_peeling

In the future, we should incorporate this into the PyAbel package. 

EDIT: This "onion peeling" method is distinct from the "polar onion peeling" method discussed in https://github.com/PyAbel/PyAbel/issues/30. Polar onion peeling uses a basis set of fairly broad legendre polynomials and cannot resolve very small angular features (I think), whereas the onion peeling algorithm does not require a basis set.

DanHickstein:
I got a preliminary port of this working. The results look pretty reasonable for the O2 test image. 

Inverse abel transform on left, original image on right
![image](https://cloud.githubusercontent.com/assets/1107796/11916956/7198671c-a694-11e5-8c1f-2c0e726b87ab.png)


The code is fairly short. Here is the full working example. I'm not 100% sure what's going on, but basically it's iterating over the column from outside to inside and subtracting the projection of each column from the inner columns. 

Here is the code:
```python
import numpy as np
import matplotlib.pyplot as plt
import time

def init_abel(xc,yc):
    # this seems like it could be vectorized pretty easily
    val1 = np.zeros((xc+1, xc+1))
    val2 = np.zeros((xc+1, yc+1))
    
    for ii in range(0,xc+1): 
        for jj in range(ii, xc+1):
            val1[ii,jj] = np.arcsin((ii+1)/float(jj+1))-np.arcsin((ii)/float(jj+1)) 

    for idist in range(0,xc+1):
        for jdist in range (0,yc+1): 
            val2[idist,jdist] = np.sqrt((idist+1)**2+(jdist+1)**2)/(idist+1)

    return val1,val2


[...]

Issue #55, opened 2015-12-15
Implement Hansen-Law forward transform?


DanHickstein:
Equation 15 of the Hansen and Law paper is the forward Abel transform, and is the analog to the inverse transform in equation 17, which we have implemented in ``abel.iabel_hansenlaw_transform``. The forward version has the potential to be a very efficient forward abel transform, so we should implement it when we get a chance. 

What do you think @stggh?

![image](https://cloud.githubusercontent.com/assets/1107796/11816563/324b8c26-a30e-11e5-9adf-0a68d45a2e46.png)
![image](https://cloud.githubusercontent.com/assets/1107796/11816570/39a37970-a30e-11e5-834c-e19706955f06.png)

stggh:
pushed ```fabel_hasenlaw_transform``` and unit test.  The change from ```iabel...()``` is trivial, to ```Gamma``` and ```gp``` is now just the raw image.  The two functions could easily be combined into one. However, I have not decided on the best way to do this.  The unit tests need tightening.

stggh:
@DanHickstein now combined into a single function/method. Unfortunately, this leaves psuedo ```f/iabel_ ``` function names in order to comply with the naming convention.  It would be nice to employ some sort of function decorator. Not yet  fully tested.

DanHickstein:
Great work! That happened much faster than I expected! :)

If the algorithms are basically identical except for a minor switch, I think that it's fine to have one of the functions, say ``abel.iabel_hansenlaw`` be the "real function" and then have ``abel.fabel_hansenlaw`` just call ``abel.iabel_hansenlaw`` and pass along all of the arguments as well as switching the keyword ``transfrom_type`` from ``transform_type='inverse'`` to ``transform_type='forward'``. Alternatively, we could just call the main function ``abel.hansenlaw`` and have the iabel and fabel methods just call hansenlaw with either the forward or reverse keywords. 

Maybe @rth has an opinion on this.

stggh:
@DanHickstein  I have examined the beautiful ```abel.direct.py``` python code of @rth and will implement the same naming scheme and boolean flag inverse=True/False.

At the moment I am revisiting my persistent nemesis, the output forward (inverse) amplitudes from HL. For the Gaussian function, using  adapted  code from ```test_direct.py/test_forward_direct_gaussian()``` the forward transform agreement is good, apart from scaling factors (in this case N=101, ref. Gauss. is ref.fun x &sigma;&Sqrt;&pi;):

[...]

PR #67, opened 2016-01-08
Updated hansenlaw, tools, tests, examples

stggh:
I have been on a summer break (part 1 break finishes this Monday) and have made a few code changes, as discussed issue #60 (anisotropy parameter calculation), #55 (forward Abel transform using hansenlaw algorithm), issue #39 (polar grid coordinate sizes), issue #36 (consistent naming, in particular, `_abel_hansenlaw_transform_core()`, and `inverse=True/False` parameter). Top-right quadrant, `Q0` is now the orientation for the `transform` code (issue #56). I have updated the HL examples, and tools. In the `example_all_O2.py` the `three_point_transform()` did not finish in a sensible time and so has been commented out.

I think it would be a good time to incorporate the changes since they provide new functionality.

**NB  these changes may break legacy code** 
  -  calculate_speeds returns speeds, r, theta        
  -  cart2polar   - angle now referenced to vertical  
  -  polar2cart                                       

**Summary** (in plain text - any markup is accidental!)
(1)  abel/

   (a) abel.hansenlaw.py

   _abel_hansenlaw_transform_core  = forward and backward Abel transforms

   Q0 top-right quadrant is now the default (cf the old Q1 top-left)

   (b) abel.analytica.py

    added sample_image_dribinski(n=361, origin=None)

      to generate the BASEX model image

(2) abel.tools.py

    (a) calculate_speeds(IM, origin=None, Jacobian=False, dr=1, dt=None)

      New parameters:
       -Jacobian: multiply intensity by    r**2 sin theta

       -dr: radial coordinate spacing
       -dt: theta coordinate spacing (degrees) - best left as default
 --------

    (b) calculate_angular_distributions(IM, radial_ranges=None)

     This function mirrors calculate_speeds with sum along axis=0 (radius)
      parameters:
        - radial_ranges: tuple list [(r0, r1), (r2, r3), ...]
                      Evaluate the intensity vs angle for the radial
                      ranges r0_r1, r2_r3, etc.
     returns
        - theta, intensity-vs-theta for each radial range (ri, rj)
 --------

    (c) center_image_by_slice(IM, slice_width=10, r_range=(0,-1),
                          pixel_center=True):

       Center image by comparing opposite side, vertical and
       horizontal slice profiles

       Parameters
       ----------
         - IM : rows x cols numpy array
         - slice_width : add together this number of rows (cols) to improve signal

         - r_range(rmin,rmax): radial range [rmin,rmax] for slide profile comparison
         - pixel_center: boolean, make center of image within a pixel
 --------

    (d) anisotropy_parameter(theta, intensity, theta_ranges=None)

       Evaluates beta using least-squares fitting
      (NB may be a bug with the theta range evaluation)

 --------

    (e) reproject_image_into_polar(data, origin=None, rmax='corner', dr=1, dt=None)

      new parameters
            - rmax:   radius maximum, beyond which no projection is done
               'corner' => sqrt((rows//2)**2+(cols//2)**2)
                  - used to convert the whole cartesian image to polar
               otherwise the minimum of the 1/2 width or 1/2 height of the image
               = min((rows//2, cols//2)
                  - used to convert only the circular region that fits within
                    the image
     - dr: radial coordinate spacing for the grid interpolation
             tests show that there is not much point in going below 0.5
     - dt: angular coordinate spacing (in degrees)

 --------

(3) abel.tests.test_hansenlaw.py

   added  "Curve A" functions f(r), g(R)  - these really belong in
   abel/analytical.py

  Modified tests to pass ;-)
   hansenlaw performs better with a larger size image (pixels)
   boosted n=51 -> 1001

   added forward tests

   Note, the comparision ratios are somewhat relaxed and could
   possibly be improved. In general hansenlaw misbehaves near
   the image center. This may be associated with residual origin center
   issues.

   To do - add further tests for anisotropy calculations

(4) abel.examples

     Modified hansenlaw tests to work with new calculate_speeds O/P format
~


abel/analytical.py
```diff
@@ -1,11 +1,70 @@
...
+        return 2000*(t0+t1+t2) + 200*(t3+t4+t5) + 50*(t6+t7+t8) + t9
+
+    # set up an dummy image 2D n x n numpy array shape 
+    im = np.arange(n*n).reshape((n,n))
+
+    if origin is None:
+        origin = (n//2+n%2, n//2+n%2)
+
+    # determine cartesian coordinates (x,y), for each pixel
+    x, y = index_coords(im, origin=origin)
```
DanHickstein: It might be more straightforward to use meshgrid, since that's basically what you are setting up. (Also, I'm not sure if this is as real python style, but I tend to use lower-case ``x`` for variables or 1D arrays, and ``X`` for 2D arrays.) You could use:

```python
x = np.arange(n)
X,Y = np.meshgrid(x,x)
R,THETA = cart2polar(X,Y) 
IM = I(R,THETA)

File: abel/tools.py [hunk 2]
```diff
@@ -7,89 +7,196 @@
...
+    if Jacobian:  #  x r^2 sinθ    
+        sintheta = np.abs(np.sin(theta))
+        polarIM = polarIM*sintheta[np.newaxis, :]
+        r2 = r**2
+        polarIM = polarIM*r2[:,np.newaxis]
+    
     speeds = np.sum(polarIM, axis=1)
     
-    # Clip the data corresponding to the corners, since these pixels contain incomplete information
+    # Clip the data corresponding to the corners, since
```
DanHickstein: Since you are using ``rmax='circle``, then you should not need to crop the image, right?

DanHickstein:
Strong work @stggh! What a monster of a PR! So much good stuff. 

A few comments:

1. Thanks for adding the preliminary onion peeling code! I was not totally satisfied with the way that it worked for the gaussian example image, so I do not really trust it yet. Please include in the docstring that this function is in alpha testing and unreliable results may be expected. Actually, it should probably print a warning to the effect of "This function is in early testing stages and may not produce reasonable results."

2. Currently, you are including the ``r**2 * sin(theta)`` Jacobian in the calc_speeds function. However, I think that the calc_speeds function only incurs a ``r*sin(theta)`` Jacobian. The other factor of ``r`` arises from the reprojection into polar coords, and it should be incorporated into the ``reproject_image_into_polar()`` function. Can you give it a similar` `jacobian=True`` parameter?

3. On a related note, your code for applying the Jacobian is a little convoluted because we are returning 1D arrays for r and theta from the ``reproject_image_into_polar()`` function. Instead, we could return the 2D meshgrid arrays from the ``reproject...`` fuinction: ``return output, r_grid, theta_grid``. In general, this seems like the better option. If people want the 1D arrays, they can just take the first element (either row or column) from those 2D arrays. It probably makes the algebra easier for everyone using the ``reproject_image_into_polar()`` function.

4. I think that it's potentially confusing to have ``calc_speeds`` return theta. The x-axis for the returned data is ``r``, so what relevance does theta have? (Yes, of course the theta bin size may affect the results, but after the integration has taken place, theta has not importance.) Is there a reason that people may really want the theta array from ``calc_speeds``?

5. That's awesome that you have implemented an automated centering function! People are going to love this. However, some users (like @DhrubajyotiDas and @rth) have data that only has horizontal symmetry, and not vertical. Can you easily modify your function to only center in either the horizontal or vertical. Perhaps something like ``center_image_by_slice(IM, center_vertical=True, center_horizontal=True, ...)``.

Thanks again for these great improvements!

stggh:
I have addressed all of @DanHickstein's excellent suggestions.   Now I know where `IM` comes from ;-) 

I am a `meshgrid` noob and am not confident that I have the right or nice python expression to obtain the `r` and `theta` coordinates:
```
    polarIM, r_grid, theta_grid = reproject_image_into_polar(IM, origin,
                                       Jacobian=Jacobian, rmax='circle',
                                                           dr=dr, dt=dt)
    theta = theta_grid[0,:]  # theta coordinates
    r = r_grid[:,0]          # radial coordinates
```
However, the tests appear fine:
jacobian

I will, later, double check the PES spectral intensities.

A couple of queries:
(1) `center_vertical/_horizontal` - would it be better to use `axis=0/1` for Python consistency?
(2) The "Curve A" functions, that I have in `abel/tests/test_hansenlaw.py` are also used by @DhrubajyotiDas

DanHickstein:
Awesome @stggh! Thanks for making those changes.

Your understanding of the meshgrid appears to be fine.

Regarding the use of ``axis=1/0`` versus having two arguments ``center_horizontal=True, center_vertical=True``. Are you saying that we should use ``axis=...`` in agreement with something like [``np.mean()``](http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.mean.html)?

I was about to dismiss this suggestion, because I thought that the ``axis`` argument in ``np.mean`` could only have an integer value to specify the axis where the mean was being taken. Now that I look at the documentation, I see that ``axis`` can be a tuple of integers to specify multiple axes. So, it would work in this situation.  I think that I still prefer the ``center_vertical/horizontal`` language since I think that it's more clear, but I think that we could also use ``axis=(0,1)``.

stggh:
More in terms of array slicing [axis definition] (http://pages.physics.cornell.edu/~myers/teaching/ComputationalMethods/python/arrays.html), which is fundamental to numpy/scipy arrays.

stggh:
**Problem**

The new look `iabel_<method>_transforms()` will all operate on the Q0 quadrant, top-right orientation.

I have modified `examples/example_all_O2_v2.py` (note, additional `_v2`) to process each transform using only the top-right quadrant.

 It fails:
(1) `reproject_image_into_polar()` appears to require  `origin=(0,0)`, rather than the correct `(512,0)` for a numpy.array quadrant ? I checked the theta_grid for `(0,0)` vs `(512,0)`

(2) The/my modification concerning `rmax = 'corner'/'circle' that sets `rmax ~ nx//2` is **incorrect**, because a quadrant is already 1/2-image. I tried to *fix* it, as below, but there are still issues:
```python
if rmax == 'corner':
       rmax = r.max()
       nr = rmax
    else:
       min_nxy = np.min((nx,ny))
       # Fix me!  not 1/2 image if dealing with quadrant
       if origin[0] > 1:
           rmax = min_nxy//2 + min_nxy%2 - 1
           nr = int((min_nxy//2 + min_nxy%2)/dr+0.5)
       else:
           nr = rmax = min_nxy
```

Here is a minimal example:
```python
from abel.hansenlaw import *
from abel.tools import *
from matplotlib.pyplot import *
IM = np.loadtxt("data/O2-ANU1024.txt.bz2") # or select your own image
Q = get_image_quadrants(IM)
AQ0 = iabel_hansenlaw_transform(Q[0])   # or use your preferred method
imshow(AQ0,vmin=0)
show()   # looks fine
speeds, r = calculate_speeds(AQ0, origin=(0,0))
plot(r,speeds)  # looks bad
show()
```
I think the problem sits in `reproject_image_into_polar()`.  Does it behave correctly when given a Q0 quadrant image?

abel/tools.py
```diff
@@ -195,50 +301,163 @@ def center_image_asym(data, center_column, n_vert, n_horz, verbose=False):
...
     # Determine that the min and max r and theta coords will be...
     x, y = index_coords(data, origin=origin)
-    r, theta = cart2polar(x, y)
+    r, theta = cart2polar(x, y)   # convert (x,y) -> (r,θ)
 
-    nr = r.max()
-    nt = ny//2
+    if rmax == 'corner':
+       rmax = r.max()
+       nr = rmax
```
DanHickstein: ``nr`` needs to take into account ``nr``. At a minimum, this should be 
``nr = rmax/dr``,
but perhaps more complicated as it is below?

abel/tools.py
```diff
@@ -195,50 +301,163 @@ def center_image_asym(data, center_column, n_vert, n_horz, verbose=False):
...
+       if origin[0] > 1:
+           rmax = min_nxy//2 + min_nxy%2 - 1 
+           nr = int((min_nxy//2 + min_nxy%2)/dr+0.5)
+       else:
+           nr = rmax = min_nxy
+         
+    if dt is not None:
+       nt = int((theta.max()-theta.min())/(np.pi*dt/180)+0.5)
+    else:
+       nt = ny//2
```
DanHickstein: We shouldn't set ``nt`` in terms of just ``ny`` since this provides inconsistent behavior for rectangular images if they are rotated. Can we just have the default be 
``nt = nr`,
or multiplied by some factor?

DanHickstein:
The ``reproject_image_into_polar()`` function seems to take into account the effect of the origin on x,y, and therefore it should get r, theta correct. It's probably ``rmax`` that is incorrect. I think that it makes sense that the ``rmax='circle'`` mode of  does not work, because there is no complete circle (centered at the origin) that can fit within a single-quadrant image. We would either need to just use the ``rmax='corner'`` mode, or add an additional mode of operation (I think this is easiest).

So, if we have 
```
x, y = index_coords(data, origin=origin)
left = -np.min(x)
right = np.max(x)
top = np.max(y) # controversial convention...
bot = -np.min(y)
```
then the ``corner`` mode uses ``rmax = np.max((left,right,top,bot)).
The ``circle`` mode uses ``rmax = np.min((left,right,top,bot)).
and we could define a ``side`` mode that uses ``rmax = np.min( (np.max(left,right), np.max(bot,top) )``.

So, basically it's finding the ``rmax`` that provides values over at least one quadrant. Alternatively, this mode could be called ``quarter_circle``.

Note: we don't actually need to use the ``left, right, bottom, top``, they just help us think about the coordinate system. But maybe they will help prevent  confusion in the future and should be used...

stggh:
@DanHickstein thanks for taking time to look at this. 
1) `reproject_image_into_polar()` is not quite off the hook, because of the quadrant image requirement that `origin=(0,0)`. For our Q0 numpy.array quadrant the origin is `(512,0)`.
```python
[ipython session]
polarAQ0, r_grid, t_grid = reproject_image_into_polar(AQ0, origin=(0,0))

t_grid.min(), t_grid.max()

(0.0, 1.5707963267948966)        # 0 -> &pi;/2   correct

polarAQ0, r_grid, t_grid = reproject_image_into_polar(AQ0)

 t_grid.min(), t_grid.max()

(-3.1376864234578261, 3.1415926535897931)    # -&pi; -> &pi;   incorrect
```

2) Good find  `nr/dr1`

3) `rmax='corner'` must always be   `np.sqrt(nx**2 + ny**2)`
 
In the end I don't think the value of `rmax` is critical, provided the returned  reprojected image corresponds to part of the quadrant, since `calculate_speeds()` simply sums the image along one axis.

DanHickstein:
@stggh 

1. I'm confused what you are saying here. Is Q0 the upper right or upper left quadrant? Why does the origin need to be ``0,0``, and not ``0,512``? The example behavior that you demonstrated seems to make sense. In the first case it is being treated as a quadrant, in the second case ``reproject_image_into_polar()`` is defaulting to putting the origin in the center.

3) Yeah, you are totally right. My brain was turned off this morning apparently :)

And you are right that ``rmax`` is not especially important in principle as long as the user knows what it is, can specify it, and the behavior is predictable.

stggh:
@DanHickstein  Q0 is the upper right quadrant. 

Ahh... I now see my issue. `origin` coordinate is relative to the bottom-left corner.  I had originally thought that this should be a numpy.array coordinate (relative to the top-left corner).

Thanks.

stggh:
I have not resolved the quadrant `reproject_image_into_polar()` issue.

If I reverse the order of theta, r, for the function and tuple in `theta_grid, r_grid = np.meshgrid(theta_i, r_i)`
then the polar image almost looks correct, but noisy, and the same for the speed distribution. However, this then breaks the fully image reprojection. Perhaps, some array sorting is required?

I dropped the `rmax` parameter and code from `reproject_image_into_polar()`. IMO the code complexity is not justified, the reprojection appears to be quite fast in any case. `rmax` can be applied within `calculate_speeds()`.

DanHickstein:
@stggh, the meshgrid looks correct to me. Can you provide and example of what's going wrong? 

I suppose it's fine to ignore ``rmax`` and deal with it in calculate_speeds...

rth:
@stggh Thanks you for this very extensive work! Everything looks good! I have been using the `hansenlaw.py` from your fork and can confirm that it works nicely (well, better than the direct implementation at this point).  Just a few minor comments:

1. We should merge @DhrubajyotiDas's PR first as there is some redundant commits here.
2. Nice work for the  `center_image_by_slice` function! Thought can we rename it to something more verbose? Right now given  `center_image` and `center_image_by_slice` function names, it is not clear that one finds the center automatically and the other does not. Maybe `find_center_image` or `find_image_center`' for the latter one (or something similar)? 
3. Both `examples/example_all_O2.py` and `examples/example_all_O2_v2.py` are included. I looks like they would not produce exactly the same image with comments in `_v2`, however the latter code is much more structured, readable and extendable IMO. Can we just keep the `_v2`, rename it to `example_all_O2.py`, or is there a reason to keep the original file?
4. After this PR `tools.py` will include ~12 functions, ranging from speed and anisotropy calculations to centring and mathematical operations in polar coordinates. Maybe we could move the speed and anisotropy calculation to `vmi.py`? As I suppose in future, there could be even more tools related to velocity calculations  and `tools.py`, and it will become harder to find functions in there. @DanHickstein what do you think?
5. Regarding earlier discussion, I agree on the `axis=1/0` notation. 
6. Finally, and more importantly, right now, unless I missed them, there are no unit tests for `calculate_angular_distributions`, `anisotropy_parameter`, `put_image_quadrants`, `center_image_by_slice` in `test_tools.py` (or `test_vsi.py` if the point above is implemented). Would it be possible to add even some very basic unit tests (even just executing them with some random data in a test function would be a start) for those functions to ensure that they work on all supported Python versions, (on Windows and Linux); and would not be broken if someone were to edit it in the future?  The [current code coverage](https://travis-ci.org/PyAbel/PyAbel/jobs/101566997) can be seen at the end of the Travis CI build (clicking on the 'Details' button in the "All checks have passed" section of the Pull request, then by selecting a Python version). At present the test coverage is of 54% for PyAbel (i.e. fraction of  lines that were executed during unit tests) and of 32% for `abel.tools` . If we can increase the latter a bit, that would be great!

stggh:
@DanHickstein, @rth, @DhrubajyotiDas 

Here is the issue with `reproject_image_into_polar()` for a whole image cf the top-right quadrant (Q0):
quadtest

The inverse Abel transforms are fine. The reprojection for the quadrant is incorrect. I think the issue has something to do with the ordering of the coordinates.

Here is the code used to produce the diagram:
```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

# demonstrate an issue with reproject_image_into_polar()
# for top-quadrant cf whole image

from abel.hansenlaw import *
from abel.tools import *

# O2- VMI example image
IM = np.loadtxt("data/O2-ANU1024.txt.bz2") # or select your own image

# split into quadrants (only want Q0)
Q0, Q1, Q2, Q3 = get_image_quadrants(IM)


# inverse Abel transform, for Q0 quadrant, and the whole image
AQ0 = iabel_hansenlaw_transform(Q0)   # single quadrant transform  

AIM = iabel_hansenlaw(IM)  # whole image transform

# reprojected images
polarAQ0, AQ0r, AQ0t = reproject_image_into_polar(AQ0, origin=(0,0))

polarIM, AIMr, AIMt = reproject_image_into_polar(AIM)


# visual


vmax=6
plt.subplot(221)

plt.imshow(AIM, vmin=0, vmax=vmax)
plt.title("inverse Abel whole image", fontsize=12)

plt.subplot(222)
plt.imshow(polarIM, vmin=0, vmax=vmax)
plt.title("polar projection whole image", fontsize=12)
plt.xlabel("$r-$axis", fontsize=12)
plt.ylabel("$\\theta-$axis", fontsize=12)

plt.subplot(223)
plt.imshow(AQ0, vmin=0, vmax=vmax)
plt.title("inverse Abel Q0 quadrant", fontsize=12)

plt.subplot(224)
plt.imshow(polarAQ0, vmin=0, vmax=vmax)
plt.title("polar projection quadrant image", fontsize=12)

plt.subplots_adjust(hspace=.5)
plt.savefig("quadtest.png", dpi=150)
plt.show()
```

1. Yes, definitely, go ahead and merge @DhrubajyotiDas's PR first
2. `find_image_center()` - ok
3. `example_all_O2_v2.py` is to replace `...O2.py`, but currently the quadrant speed distribution is not working, due to the `reproject_image_into_polar()` issue, above.
4. `vmi.py` - agree,  good idea.
5. `axis=0/1` - I also favour.
6. Unit tests - yes, very important. I will provide them.

DhrubajyotiDas:
I just merged the three_point PR so we can move ahead with this one.

stggh:
Hmm, tried to rebase, may have made a mess of it

DhrubajyotiDas:
[I think your version of the code inside `basex.py` might have been outdated...](https://github.com/stggh/PyAbel/commit/9cf1a97477489c51006191df87237971b9699521#diff-cfaa2abc5c27b88c982cc40aeb483449R44)

stggh:
The rebase did not complain about `basex.py`, just `three_point.py`, tests, and examples. I don't understand why it re-listed all the commits.  I will take a look at the tests.


@DhrubajyotiDas , @DanHickstein, @rth 
re `basex.py` it still has the old interface. Will it be updated to (f/i)_abel_basex()`?

rth:
Aww, the joys of git... sometimes it feels like it's doing things on it's own in a semi-deterministic way...

So as far  as I have seen there were 3 states for this PR with respect to the git history:
  * State 1: before the final rebase following @DhrubajyotiDas's PR => tests passed, ~40 commits
  * State 2: just after the final rebase => tests failed and the `basex.py` corresponded to an outdated version ~40 commits
  * State 3: trying to fix the rebase => tests passed but multiple occurrences of the same commits (~140 new commits in total). 

It is annoying that a rebase changes past git history, but here is a version of what could have happened.

I'm not  sure whether the last rebase (up to the state 2) actually broke anything, as I remember when reviewing the code in State 1, wondering about  `basex.py` that had significant changes. I did not pay too much attention at the time, but with @DhrubajyotiDas's comment it did look exactly as the PR to add the [asym BASEX](https://github.com/PyAbel/PyAbel/pull/54/files#diff-cfaa2abc5c27b88c982cc40aeb483449L43), which means that something was already wrong there in State 1.  I am not sure why unit tests still passed at the time, as can be seen [here](https://travis-ci.org/PyAbel/PyAbel/pull_requests), maybe there was a glitch with Travis CI or with rendering of the diff on github (something doesn't seem to be consistent there). 

Then after the rebase in State 2, the git tree actually looked good, although the tests failed with an outdated version of `basex.py`. Finally, State 3 fixed the unit tests but there is multiple repetitive commits that add and remove the same code, and I don't think we should merge it. 

Bottom line, I don't know why this issue happened. It is somewhat similar to the issue @DhrubajyotiDas  had a while back and is possible that this was due to [this commit](https://github.com/stggh/PyAbel/commit/8281fe9da0dac5b4f976b07a02d585f8ccb4efe6) from a different email (" Stephen Gibson committed with stggh")  which for git would be a commit by a different user. Therefore when rebasing git history as stggh, things would go bad with the same commits being added and removed as we have seen with @DhrubajyotiDas . 

Anyway, to get out of here and keep a reasonable git history, I propose to go to the last point where [things looked good]( https://github.com/PyAbel/PyAbel/network)(i.e. before the last rebase) which would be commit d3c512aadf2ca834642eb40a197f25e0e5649eb6, then create a new branch from here and simply make a new PR from that branch (as I don't think we can recover your master branch at this point). It should include all of the commits, no? It some commits are missing they could probably be added one by one. So the commands would be something like, 

      git checkout d3c512aadf2ca834642eb40a197f25e0e5649eb6
      git checkout -b new-branch
      # if this was not already done
      # git remote add upstream git@github.com:PyAbel/PyAbel.git 
     git fetch upstream
     git merge upstream/master
     # fix some conflicts in 3-point test
     # git push ( the command you would use to push normally)
 
after which we can close this PR from the master branch without merging it as it won't be necessary anymore.

I tried to run the commands above and pushed it to the [luli fork](https://github.com/luli/PyAbel) which gives the following [diff](https://github.com/PyAbel/PyAbel/compare/master...luli:new-branch?expand=1#diff-cfaa2abc5c27b88c982cc40aeb483449), so at a first glance it seems to work. I will remove this fork and branch as it is just for testing...

@stggh What do you think? Would this work for you?

stggh:
@rth should probably take my git drivers licence away ...

The only issue is:
```
git add abel/tests/test_three_point.py
[vuvpc5:.+thon/VMI/PyAbel]$ git push
fatal: The current branch new-branch has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin new-branch
```

Which I duly copied, and now I have a branch "new-branch" !

How may I reconcile back to just one branch?

rth:
@stggh Yeah, git does weird things sometimes...

Well I don't think there is a reason to reconcile to just one branch just now, since from the moment you will try to merge this with the master of your fork all those issues with duplicates commits will come back, as far as I understand. I was thinking, more like just working on this `new_branch` for the moment, and make a new independent PR request from it (in addition to the current one). Then once these commits are merged to PyAbel, we close this PR, and you could just delete your fork on github, then refork to get yourself a single clean master branch.

Yes, I think 
   
    git push  origin new-branch
should work with or without the `--set-upstream` flag. The only issue is that `new-branch` does not seem to appear in the [git tree](https://github.com/PyAbel/PyAbel/network) under your fork after you pushed it.

Well these are all suggestions anyway, maybe there is a better way of doing it..

stggh:
Thanks @rth git guru.  I think that I am back.  I  had already deleted the redundant branches.  If there are any issues I can always start again from the PyAbel master ...

rth:
Happy that it worked...

stggh:
@rth not quite!    git push creates  'new-branch'

rth:
Yes, and now if you open another PR from this `new-branch` on your fork [here](https://github.com/PyAbel/PyAbel/compare/master...stggh:new-branch?expand=1) it should (hopefully) work.

stggh:
PR moved to #68

Issue #65, opened 2016-01-07
Include only the asymmetric transform in BASEX?


DhrubajyotiDas:
With PR #54, BASEX can invert line-of-sight images which do not possess up/down symmetry. As @DanHickstein mentioned [here](https://github.com/PyAbel/PyAbel/pull/54#issuecomment-167258811), this makes the code for the (original) symmetric BASEX transform somewhat redundant. An up/down symmetric inversion ought to be equivalent to taking an image, averaging the top and bottom halves, and passing it through the asymmetric inversion routine. It turns out that this is true, at least for a Gaussian function with some random noise, as can be seen here.

(VS and no VS refer to Vertical Symmetry)

```python

import numpy as np
import matplotlib.pyplot as plt
from abel.basex import BASEX
from abel.analytical import GaussianAnalytical

fig, axes = plt.subplots(nrows = 3, ncols = 3)
fig.set_size_inches(9,9)

axes[0,0].set_title('Original LOS signal')
axes[0,1].set_title('Noisy LOS signal')
axes[0,2].set_title("Up/down averaged LOS signal")
axes[1,0].set_title('Abel inverted signal')
axes[1,1].set_title('With VS')
axes[1,2].set_title('Without VS')

n = 101
n_vert = 51
r_max = 4
sigma = 1

ref = GaussianAnalytical(n, r_max, sigma)
raw_data = np.tile(ref.abel, (n_vert, 1))
noisy_raw_data = np.zeros_like(raw_data)
averaged_raw_data = np.zeros_like(raw_data)
for idx, row in enumerate(raw_data):
    noise = np.random.normal(0,1,n)
    noisy_raw_data[idx] = row + noise/100.0
    
rows = noisy_raw_data.shape[0]

for idx in range(rows):
    averaged_raw_data[idx] = 0.5*(noisy_raw_data[idx] + noisy_raw_data[rows - idx - 1])

recon_vs = BASEX(raw_data, center=(n//2,n_vert//2), n=[n_vert, n], 
                 basis_dir = './', dr = ref.dr, vertical_symmetry=True, verbose=False)

recon_no_vs = BASEX(raw_data, center=(n//2,n_vert//2), n=[n_vert, n], 
                    basis_dir = './', dr = ref.dr, vertical_symmetry=False, verbose=False)

noisyrecon_vs = BASEX(noisy_raw_data, center=(n//2,n_vert//2), n=[n_vert, n], 
[...]

PR #68, opened 2016-01-12
Anisotropy analysis, update hansenlaw, rationalization of transform method interface

stggh:
Continued from #67 following PR #61


abel/hansenlaw.py
```diff
@@ -4,15 +4,14 @@
...
 import numpy as np
 from time import time
 from math import exp, log, pow, pi
 from abel.tools import calculate_speeds, get_image_quadrants,\
                        put_image_quadrants
 
-###########################################################################
-# hasenlaw - a recursive method inverse Abel transformation algorithm 
+################################################################################
+# hasenlaw - a recursive method forwrd/inverse Abel transform algorithm
```
DhrubajyotiDas: Typos - `hansenlaw - a recursive method forward/inverse Abel transform algorithm `

stggh:
@DhrubajyotiDas 'Typos - hansenlaw - a recursive method forward/inverse Abel transform algorithm'

Thanks. 

Q: Will the basex interface be changed to the 'standard' (f/i)abel_basex form?

DhrubajyotiDas:
@stggh There's another typo (`hasenlaw` --> `hansenlaw`) 

>Q: Will the basex interface be changed to the 'standard' (f/i)abel_basex form?

Yes, it will be changed (hopefully very soon). I'm going to start working on testing the forward BASEX implementation this week. If things go smoothly, that will be a good time to standardize the basex interface, unless @DanHickstein and @rth have other plans.

stggh:
I have tidied up `example_all_O2.py`, which now processes only the image top-right quadrant Q0, and hence, uses the core function of each method.  Method plot labels are not very well placed (to do).
example_all_o2

>./example_all_O2.py
>centered image 'data/O2-ANU2048.txt' shape = 1023x1023
>quadrant shape (512, 512)
>
>------- basex inverse ...
>Loading basis sets...           
>0.64 seconds
>Reconstructing image...         
>0.97 seconds
>                   1.7 sec
>
>------- direct inverse ...
>                   0.2 sec
>
>------- hansenlaw inverse ...
>                    0.1 sec
>
>------- onion inverse ...
>  Warning: iabel_onion_peeling() is in early testing and
>            may not produce reasonable results
>Image cannot have negative values, setting negative values to zero   # [ not sure what causes this ]
>                    5.5 sec
>
>------- three_point inverse ...
>                    0.1 sec

Issues:
 1) The `onion_peeling()` which now generates image noise at the center = puzzle.  
 2)  The `basex` speed distribution (broader peaks) indicated the image (reconstituted from Q0) is not correctly centered/made.  This issue may be eliminated once `basex` can handle a single Q0 quadrant.

Note also, as mentioned previously #67, in order to obtain the correct  polar image it is necessary to `np.flipud()` the Q0 quadrant image.  This flip is now done within reproject_image_into_polar(), in order to make (0,0) correspond to the bottom-left of the image. Here is the test case, for Q0, that I ran in #67:
quadtest

DanHickstein:
Thanks for all of the great work guys! I've been busy for the past few days starting a new postdoc job.

@DhrubajyotiDas should go ahead and standardize the BASEX interface if he's got time. We've just been lazy about this :)

1) The onion peeling method still produces crazy results (see https://github.com/PyAbel/PyAbel/issues/56#issuecomment-166974406), so I think that there is a serious bug with the code, which may be causing the centerline noise and affecting the scaling.

2) BASEX needs odd images, so I wonder how things are getting stitched back together in order to provide this. Intuitively it seems like this could cause a slight problem.

3) Glad that you solved the reproject_into_polar issue. I think that this brings up a broader issue about what we think of as the top and bottom of the image and explicitly state this in the documentation. I typically think about 0 as being the bottom of the image, but the typical image/computer-science definition is for 0 to be the top. PyAbel is exclusively using the "zero-top" definition, correct? We should state this somewhere in the wiki.

Also, to make this clear in the code, when we say things in the doc-string like ``"origin" is a tuple of (x0, y0) relative to the bottom-left image corner``, we should probably append ``(where the image is oriented such that IM[0] corresponds to the top row)``.

stggh:
The issue with `onion_peeling()` center noise been resolved.  I (stupidly) missed the `[:,` from the Q0 flip! `example_all_O2.py` updated. Note, all of this PR is in `new-branch` of my fork.

stggh:
I am attempting to tidy up this PR and would like some advice on how to split `tools.py` as per @rth's suggestion discussed in #67:
>After this PR tools.py will include ~12 functions, ranging from speed and anisotropy calculations to centring and mathematical operations in polar coordinates. Maybe we could move the speed and anisotropy calculation to vmi.py? As I suppose in future, there could be even more tools related to velocity calculations and tools.py, and it will become harder to find functions in there.

My preference is to make a `tools` directory containing:
  -  `vmi.py` with `calculate_speeds()`, `calculate_angular_distributions()`, `anisotropy_parameter()`, `find_image_center_by_slice()`,
  - `symmetry.py` with `get_image_quadrants()`, `put_image_quadrants()`, `center_image()`, `center_image_asym()`,  and
  - `polar.py` with `reproject_image_into_polar()` with its helper functions `index_coords()`, `cart2polar()`, `polar2cart()`.

This leaves `CythonExtensionsNotBuilt()` without a specific themed home file.

There is a conflict between what function will live in `vmi.py` vs `symmetry.py`. eg. `find_image_by_slice()` could move to `symmetry.py`, just that it is used in VMI analysis.

In the future I will add `circularize_image_by_slice()`, that may equally live in either file.

Since, this will be a major structural change, what are your thoughts?
@DanHickstein, @rth, @DhrubajyotiDas

DhrubajyotiDas:
@stggh I think splitting the tools into separate files inside a `tools` folder is a good idea. I am currently working on writing a function to average the top and bottom halves of an image for use in asymmetric BASEX (as per #65) and I think `symmetry.py` would be the natural home for it. 

I expect that, inevitably, some functions will seem appropriate for multiple `xxx.py` files - Some options for handling these cases include deciding their fate on a case-by-case basis or dropping them in `miscellaneous.py`.

stggh:
Hmm, apparently a subdirectory structure is not recommended, nor is the use of underscores for import, so `tools_vmi.py` would be out. 

For the moment I will go with `polar.py`, `symmetry.py` and `vmi.py`.

rth:
@stggh +1 for splitting `vmi.py`, `symmetry.py`, `polar.py`. 
Hmm, IMO using subdirectories is fine (the whole `scipy` package is built with subdirectories) and it is the natural way of organizing code when things become complicated. So I think using the `tools/` folder is a good idea. 

There is also `abel.math.py` that we can put some functions in and/or move into the `tools` folder (and which can be used for functions that don't fit any directory for instance). I'm not so keen on `miscellaneous.py` as that word is rather hard to spell right for non native English users :)

DanHickstein:
I would rather look at fewer files, even if they contain more functions, but if the consensus is to break up tools.py into several files, that is fine with me. 

Does using a subdirectory necessarily mean that the functions would have the form 
``abel.tools.vmi.calculate_angular_distribution()``?

Or can we make decisions about how we import the functions separately from the actual file structure? To me, having just three items (i.e., ``abel.xxxx.xxxx``) is preferable to four (i.e., ``abel.xxxx.xxxx.xxxx``). And four should probably be the maximum. 

Now that I look at it, we are a bit inconsistent about what is imported when we import the abel package. See the following ipython autofill:

```python
In [1]: import abel

In [2]: abel.
abel.absolute_import   abel.io                abel.tools
abel.basex             abel.math              abel.unicode_literals
abel.division          abel.print_function
```

So, we are importing BASEX but not any of the other abel transform methods. The package is small enough so that we should probably just import everything.

stggh:
Update: I have been restructuring abel/tools, and adding unit tests (test_tools_vmi.py). The latter has revealed some bugs that I am investigating. e.g the new look `find_image_center_by_slice()` appears to center the image correctly (from the plots of the slice profiles), but the speed distribution for the centered image is broadened. Perhaps, the `scipy.ndimage.interpolation.shift(IM, shift) is in the wrong direction. Checking ...

I have pushed the changes to my fork/master until fixed, so to not generate too much noise on the PR my fork/new-branch from failed builds.

stggh:
Sigh. Unit tests pass for me ...
```
nosetests -s  abel/tests/ --verbosity=2  --with-coverage --cover-package=abel
:
Ran 33 tests in 2.656s

OK (SKIP=6)
```
I will take another look tomorrow

rth:
Merging this PR following comments [1](https://github.com/PyAbel/PyAbel/pull/69#issuecomment-172970661) and [2](https://github.com/PyAbel/PyAbel/pull/69#issuecomment-172984845) in PR #69.

stggh:
Thanks @rth

DanHickstein:
Awesome! Great work on this @stggh!

PR #72, opened 2016-01-26
abel/tools/analytical.py sample_image_dribinski() code correction as per #70

stggh:
Code correction to the BASEX paper sample image function `abel.tools.analytical.sample_image_dribinski()` as per discussion issue #70.

*to do: add parameters to set image center, and provide for even image size*

PR #80, opened 2016-02-11
Clean up basex

DhrubajyotiDas:
In this PR, 

- @stggh [raised](https://github.com/PyAbel/PyAbel/pull/72#issuecomment-181306117) the issue of core functions within `basex.py` not following the naming conventions of the other transform algorithm functions. To resolve this, I've renamed the main functions within `basex.py` according to the agreed conventions. I tried to follow @stggh's approach with `hansenlaw.py`, but I wasn't completely sure if the "wrapper" function should be named `iabel_basex(...)` (current) or `iabel_basex_transform(...)`. Any advice would be much appreciated.

- The `center` and `n` parameters for the `_abel_basex_core(...)` function (previously `BASEX(...)`) now have sensible default values. Only mandatory parameter is the image to be transformed. Docstrings updated to reflect appropriate changes.

- As discussed in the issue triage meeting, I've removed a legacy basis set test which checked to make sure the PyAbel generated basis sets matched the sets generated using the original matlab code.

### DIFF
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -43,8 +43,15 @@ from .tools.symmetry import center_image, center_image_asym
 #
 #############################################################################
 
+# functions to conform to naming conventions: contributing.md ----------
 
-def BASEX(data, center, n, 
+def iabel_basex(IM, dr=1.0, **kwargs):
+    """
+    Inverse Abel transform for one-quadrant
+    """
+    return _abel_basex_core(IM, dr=dr, **kwargs)
+
+def _abel_basex_core(data, center='auto', n='auto', 
         nbf='auto',  basis_dir='./', calc_speeds=False, vertical_symmetry=True, dr=1.0, verbose=True):
 
     """ This function that centers the image, performs the BASEX transform (loads or generates basis sets), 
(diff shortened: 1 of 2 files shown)


stggh:
@DhrubajyotiDas    `iabel_<method>(IM)` transforms a whole image, `iabel_<method>_transform(Q0)` transforms a single quadrant (top-right, usually labelled `Q0`). basex is currently limited to processing the whole image, so your wrapper function will be `iabel_basex()`.

DanHickstein:
I think that there is some confusion here. Here is what I was thinking (which I think is also what @rth had in mind):

1) The individual `<method>.py` files (basex, hansenlaw, etc.) should provide only a 1-quadrant transform. These function should be called, for example, `iabel_basex/fabel_basex`.

2) All of the transforms should operate identically, so any code dealing with flipping quadrants, etc. belongs in the `abel.iabel/fabel` class.

3) Internal to each `<method>.py` file, there can (optionally) be additional functions like `iabel_basex_transform` to distinguish the functions completing the actual transform from those that will be called by the user. In the case of `basex` the `iabel_basex` function decides whether to call the functions to load the basis set, or generates the basis set, and then finally calls the actual transform function. The structure of these internal functions is flexible, and the only thing that is required is that each `<method>.py` file provides two functions: `iabel_<method>` and/or `fabel.<method>`.

The logic here is to try to minimize code repetition. And some feature sar

stggh:
Oh, ok, that makes sense.

DanHickstein:
This is okay to merge then, right?

PR #82, opened 2016-02-12
Preliminary format for abel.iabel and abel.fabel

DanHickstein:
Here is a rough sketch of what the `abel.iabel` and `abel.fabel` classes could look like. They are both currently at `abel.transform.iabel` and `abel.transform.fabel`, but we could import them to `abel.iabel` in `abel/__init__.py`. 

Both the `abel.iabel` and `abel.fabel` classes are just child classes of the `abel.transform.abel_transform` class, so there is minimal code repetition. As you can see, all of the quadrants code has been shipped into this class, so it no longer needs to appear in the individual transform functions. 

@PyAbel/developers, have a look and see if this works for you. Let me know your suggestions.

@rth, I imagine that you may have already come up with something better, so I have no problem with you re-writing or replacing this PR. I just wanted to get the conversation started with something concrete.

(Sorry for the undocumented code - I need to head to work, so I didn't finish polishing this.) 

Also, I found that the `put_` and `get_ image_quadrants()` did not work correctly when the `vertical_symmetry` and `horizontal_symmetry` keywords were used. I think that I made these work. I think that it might be less confusing if we just made the `get_image_quadrants()` function always re-orient the images. The `put_image_quadrants()` function implicitly assumes this anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ examples/*.pdf
 examples/*.png
 examples/*basex_basis*
 examples/*.dat
-*basex_basis*
+*basis*
 *.coverage
 *.c
 *.so
--- a/abel/tools/__init__.py
+++ b/abel/tools/__init__.py
@@ -4,3 +4,4 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 from . import io
+from . import analytical
--- a/abel/tools/symmetry.py
+++ b/abel/tools/symmetry.py
@@ -54,6 +54,15 @@ def get_image_quadrants(IM, reorient=False, vertical_symmetry=False,
     Q1 = IM[:n_c, :m_c]
     Q2 = IM[-n_c:, :m_c]
     Q3 = IM[-n_c:, -m_c:]
+    
+    if reorient:
+        Q1 = np.fliplr(Q1)
+        Q3 = np.flipud(Q3)
+        Q2 = np.fliplr(np.flipud(Q2))
+    
+    if vertical_symmetry and horizontal_symmetry and not reorient:
+        raise ValueError('In order to add quadrants (i.e., to apply horizontal or vertical symmetry),' 
+                         'you must reorient the image.')
 
     if vertical_symmetry:   # co-add quadrants
         Q0 = Q1 = Q0*use_quadrants[0]+Q1*use_quadrants[1]
@@ -63,11 +72,6 @@ def get_image_quadrants(IM, reorient=False, vertical_symmetry=False,
         Q1 = Q2 = Q1*use_quadrants[1]+Q2*use_quadrants[2]
         Q0 = Q3 = Q0*use_quadrants[0]+Q3*use_quadrants[3]
 
-    if reorient:
-        Q1 = np.fliplr(Q1)
-        Q3 = np.flipud(Q3)
-        Q2 = np.fliplr(np.flipud(Q2))
-
     return Q0, Q1, Q2, Q3
 
 
@@ -75,8 +79,10 @@ def put_image_quadrants (Q, odd_size=True, vertical_symmetry=False,
                          horizontal_symmetry=False):
     """
     Reassemble image from 4 quadrants Q = (Q0, Q1, Q2, Q3)
-    The reverse process to get_image_quadrants()
+    The reverse process to get_image_quadrants(reorient=True)
     Qi defined in abel/hansenlaw.py
+    
+    Note: the quadrants should all be oriented as Q0, the upper right quadrant
     
     Parameters
     ----------
@@ -101,6 +107,8 @@ def put_image_quadrants (Q, odd_size=True, vertical_symmetry=False,
         Reassembled image of shape (rows, cols) 
     """
 
+    Q0, Q1, Q2, Q3 = Q
+    
     if vertical_symmetry:
         Q0 = Q1
         Q3 = Q2 
@@ -110,12 +118,12 @@ def put_image_quadrants (Q, odd_size=True, vertical_symmetry=False,
         Q3 = Q0 
 
     if not odd_size:
-        Top    = np.concatenate((np.fliplr(Q[1]), Q[0]), axis=1)
-        Bottom = np.flipud(np.concatenate((np.fliplr(Q[2]), Q[3]), axis=1))
+        Top    = np.concatenate((np.fliplr(Q0), Q1), axis=1)
+        Bottom = np.flipud(np.concatenate((np.fliplr(Q2), Q3), axis=1))
     else:
         # odd size image remove extra row/column added in get_image_quadrant()
-        Top    = np.concatenate((np.fliplr(Q[1][:-1,:-1]), Q[0][:-1,:]), axis=1)
-        Bottom = np.flipud(np.concatenate((np.fliplr(Q[2][:,:-1]), Q[3]), axis=1))
+        Top    = np.concatenate((np.fliplr(Q1[:-1,:-1]), Q0[:-1,:]), axis=1)
+        Bottom = np.flipud(np.concatenate((np.fliplr(Q2[:,:-1]), Q3), axis=1))
 
     IM = np.concatenate((Top,Bottom), axis=0)
 
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -12,7 +12,7 @@ from scipy.ndimage.interpolation import shift
 from scipy.optimize import curve_fit, minimize
 
 
-def calculate_speeds(IM, origin=None, Jacobian=False, dr=1, dt=None):
+def calculate_speeds(IM, origin=None, Jacobian=True, dr=1, dt=None):
     """ Angular integration of the image.
 
         Returning the one-dimentional intensity profile as a function of the 
--- a/abel/transform.py
+++ b/abel/transform.py
@@ -9,7 +9,7 @@ import abel
 import time
 
 class abel_transform(object):
-    def __init__(self,IM,direction=None,method='hansenlaw',integrate=True,verbose=True,
+    def __init__(self,IM,direction=None,method='three_point',integrate=True,verbose=True,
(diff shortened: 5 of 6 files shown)


DanHickstein:
Oh, and here is the result of the fun example contained in `abel.transform.main()`, which shows how easy it is to forward transform with hansenlaw and inverse transform with three-point, using identical syntax. 

image

DanHickstein:
Oh, and I also need to add the image centering to the abel_transform.__init__ function...

rth:
Nice work @DanHickstein ! I was a bit slow to look into it, and it is a good thing that you did. I think after everybody interested reviews it, we can merge this.  If I think of some improvements, I will propose another PR on top of this one. 

Just a few minor comments:
  - by convention class names are capitalized in Python, so it should be `AbelTransform`. This also applies to `iabel` and `fabel`, if you want them to be a class. But actually they don't  need to be, as they are simply a function that returns a class object. This could be achieved with, for instance,

        class AbelTransform(object):
            def __init__(self, ...): 
                   # do something 

            @classmethod
             def iabel(cls, *args, **kwargs):
                  return cls(*args,direction='inverse', **kwargs)
   
            # same for fabel

        iabel = AbelTransform.iabel
        fabel = AbelTransform.fabel

   - replace `if method == 'three_point':` etc.  by `elif method..` . And  `if direction ==` by `elif`.

Otherwise looks great!

DanHickstein:
Thanks for checking this @rth! 

What are the advantages/disadvantages of the functions versus separate child class approach? I suspect in both cases we can achieve the desired goal of having these live at `abel.iabel/abel.fabel` by simply adding
```python
from . import transform.iabel as iabel
```
(as the code is currently structured) or
``` python
from . import transform.AbelTransform.iabel as iabel
```
(in your proposed change).

rth:
Well one disadvantage that if it is a class it would have to be called `Iabel`  or `IAbel`. It's more a style preference, one would use a subclass to change existing (or add new) methods of the parent class, which is not the case here (apart from the `__init__` which is a special case). Also  I'm just not so keen on the `super` call unless one cannot do without it. 

 In both the current and proposed format one could import it with `from .transform import iabel`  (that's why I added the `iabel = AbelTransform.iabel` lines at the end). ))

DanHickstein:
That makes sense. I'll make the change!

stggh:
Nice **class** work @DanHickstein, @rth.   

I find the class instance parameters a nightmare, can they be set within the relevant methods?
The result of the transform will need to be fetched, so it is appropriate to set parameters, and then call functions as required. I realise the aim is to provide a single transform call, but in that case a separate class method may do.

`AbelTransform` also includes speed and anisotropy analysis, so perhaps it could include I/O (and then be renamed 'PyAbel'?).

I am a `class` noob, so take my comments with a grain of salt.

DanHickstein:
@stggh, which part is the nightmare? Maybe you can provide an example of the concern...

stggh:
This aspect may just be my way of viewing classes, which is as more of a container.  

At the moment there are 10 instance parameters, which are aligned to the Abel transform: 
```python
def __init__(self,IM,direction=None,method='three_point',center=(None,None),verbose=True,
                 vertical_symmetry=True,horizontal_symmetry=True,use_quadrants=(True,True,True,True),
                 integrate=True,transform_options=())
```
No doubt more parameters will need to be added, `center_image=True, center_image_method="help", circularize_image`, etc.

I would rename `__init__`, say to `transform`  and then have the class constructor `__init__(self, IM):`
```python
IMob = AbelTransform(IM)
IMob.transform(direction="inverse", method="hansenlaw", ...)
IMob.calculate_speeds()
radial, speed = IMob.get_speed_data()
```
In principle, the parameters in `transform` should be set using class methods. Unfortunately, this moves away from your nice concept of a single method call.

This discussion is more for my own education.

DanHickstein:
Yes, there is much to ponder here. I am not a professional at object-oriented programming either, so I'm not sure what the conventions are. 

But I think that this comes back to some basic questions about how we want the users to approach the PyAbel program. I have always been a fan on the "one line" Abel transform, since I think that is easier for first-time users. They can get started quickly and start modifying the parameters one-by-one to suite their needs. In the end, they may wind up with a function call with lots of parameters, but I think that is okay. 

But, I am also okay with having the centering function be separated from the transform function, since I think that the two functions are completely independent. 

In contrast, some of the transform methods that we might implement in the future have better ways of extracting the radial distribution and beta (I think that POP and pBASEX do this). So, it's probably a good idea to wrap the anisotropy and speeds calculations in with the transform functions to allow for this functionality in the future. 

Basically, I would just have the `abel.iabel` and `abel.fabel` just be functions instead of classes. The only reason that I wrote them as classes is so that we can use the `object.item` syntax to return multiple items (i.e., transform, speeds, anisotropy) in an elegant way.

And I guess I am a little confused by the last code-block in your post. Was that intended as an example of use, or the actual code in the `AbelTransform.__init__.py` function?

stggh:
Example of use ;-)    My fundamental objection is the number of parameters for the class instance which are method variables. The class object is the image.  At the moment the class constructor will change as the methods change.  If the current `__init__` is renamed as a method, then new methods/parameters may be more easily added.

However, not a big deal, I am happy to try your approach.

DanHickstein:
Hmmm, I'm still not quite sure if I understand your terminology of concerns. When you say that you don't like using too many "method variables", are you just concerned with having a messy function, or some difference in scope between the method variables and instance variables? I think that you are right to say that the variable `method='baesx'` would die after the `__init__` function finishes, whereas if we set `self.method = method`, then we would preserve this as an instance variable. However, I'm not quite sure why this would be useful.

Also, I think that each instance of a class is considered an object. So it's more than the image. For example:
```python
anObject = abel.AbelTransform()
```

And you are correct that, as we have currently written the class, the constructor function (`__init__()`) will change (trivially) if we add a new transform function. But why is modifying a `AbelTransform.transform` function and harder or easier than modifying the `__init__` function.

I think that in terms of real object-oriented programming, your idea that the __init__ function should do very little is probably the normal mode of operation. I suppose it would not be too crazy to have the user initialize the object as you describe and then call all of the sub-functions, like center, transform, and integrate. 

It could look something like:

```python
transformObject = abel.iabel(IM)
centeredImage = transformObject.center('by_slice')
transformedImage = transformObject.transform(method='basex')
speeds = transformObject.angular_integrate()
```

I think that I still prefer the 
```python
results = abel.iabel(IM,center='by_slice', angular_integrate=True)
```
where we can get the things that we want from `results.transform` and `results.radial_intensity`.

DanHickstein:
Oh, heck, there are so many different ways that this could be implemented. I keep going back and forth. In any case, I think that by moving all of the symmetrizing business out of the individual transform functions, we are going to make things a lot simpler regardless of exactly how we chose to implement this

DanHickstein:
And I think that I'm just going to merge this and let @rth clean it up this weekend if he has time.

stggh:
>It could look something like:
```python
transformObject = abel.iabel(IM)
centeredImage = transformObject.center('by_slice')
transformedImage = transformObject.transform(method='basex')
speeds = transformObject.angular_integrate()
```
>I think that I still prefer the
```python
results = abel.iabel(IM,center='by_slice', angular_integrate=True)
```
>where we can get the things that we want from results.transform and results.radial_intensity.

Yes, the top example is how I think the class structure should look. It nicely, holds the intermediate results (centered image, transformed image etc) for future use. The bottom method is a kludge (IMO), but it does achieve your desired aim.  I claim the two may live together, in the same class, but without the multiparameter `__init__`.

DanHickstein:
Your arguments make sense to me. As long as we move all of the centering and "quadranting" out of the individual transform functions, I think that I'm happy. 

I will make the changes that you suggest. (And @rth and @DhrubajyotiDas can chime in if they have opinions about how the interface should operate.)

rth:
Thanks for working on this! 

@stggh I would tend to agree that something like this 
>     IMob = AbelTransform(IM)
>     IMob.transform(direction="inverse", method="hansenlaw", ...)
>     IMob.calculate_speeds()
>     radial, speed = IMob.get_speed_data()

would be more easy to use in the long run (as  more and more possible functions are added by different contributors). We already had a [long discussion](https://github.com/PyAbel/PyAbel/issues/14#issuecomment-157776749) about it a while back.  Still I think this PR is a good thing, and we can incrementally change the interface to get to a compromise point that works for everybody )) I will look into it.

stggh:
Aha, I now understand that @DanHickstein and perhaps, @rth as demonstrated in `analytical.py`(?) make use of classes as function result containers.

I believe the idea is have class `do_all_function()` with the results available in the class structure. A solution may be to use a "standard" constructor, with the current `__init__` renamed as separate`do_all_function()`:
```python
def __init__ (self, IM):
    self.IM = IM

def do_all((self, IM, direction=None, method='three_point', center=(None, None), verbose=True,
                 vertical_symmetry=True, horizontal_symmetry=True, use_quadrants=(True,True,True,True), integrate=True, transform_options=()):
```
nothing else need change, and then point `iabel =` and `fabel =` as appropriate.

I also believe that the `selected_transform(IM)` function may be replaced with:
`selected_transform = "abel.{:s}.{:s}abel_{:s}_transform()".format(method, direction[0], method)`.

Later, some of the functions in `do_all()` should be made available as class methods, and called from within `do_all`

The end result will be "standard" python class, with a convenient one call function.

I am tied up with my day job activities during today :-(, but if need be can put the class together tonight, although, this is @DanHickstein baby.

DanHickstein:
@stggh I think that I see what you are saying here. Basically, you don't mind the idea of a function that does a lot of different things, you would just prefer that I didn't stuff and __init__ function full of all kinds of crazy stuff, because that doesn't look professional. I think that the solution that you describe here is a good one. I'm making a separate PR and you can see what you think.

stggh:
@DanHickstein, yes, we are on the same page. Thanks.

PR #83, opened 2016-02-12
fix the bug of center='auto' and add a option to determine the center…

ldes89150:
… by calculating the center of mass

### DIFF
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -14,7 +14,7 @@ import sys
 import numpy as np
 from scipy.special import gammaln
 from numpy.linalg import inv
-from scipy.ndimage import median_filter, gaussian_filter
+from scipy.ndimage import median_filter, gaussian_filter, center_of_mass
 
 from ._version import __version__
 from .tools.vmi import calculate_speeds
@@ -65,7 +65,8 @@ def _abel_basex_core(data, center='auto', n='auto',
                 Abel inverse transform will be performed on a `n x n` area of the image
             * list in format [n_vert, n_horz] - 
                 Abel inverse transform will be performed on a `n[0] x n[1]` area of the image
-            * if n='auto', it is set to data.shape 
+            * if n='auto', it is set to data.shape
+
       - nbf: * integer - 
                 number of basis functions. If nbf='auto', it is set to (n//2 + 1).
              * list in format [nbf_vert, nbf_horz] -
@@ -74,7 +75,8 @@ def _abel_basex_core(data, center='auto', n='auto',
                     the center column of the image
                 * tuple (x,y) -
                     the center of the image in (x,y) format
-                * If center='auto', it is set to (data.shape[0]//2, data.shape[1]//2)
+                * If center='auto', it is set to (data.shape[1]//2, data.shape[0]//2)
+                * if n='com', it is set to the center of mass of data
       - basis_dir: string
             path to the directory for saving / loading the basis set coefficients.
             If None, the basis set will not be saved to disk. 
@@ -103,7 +105,13 @@ def _abel_basex_core(data, center='auto', n='auto',
         data_ndim = 2
 
     if n == 'auto': n = list(data.shape)
-    if center =='auto': center = (data.shape[0]//2, data.shape[1]//2)
+    if center =='auto':
+        center = (data.shape[1]//2, data.shape[0]//2)
+    elif center == 'com':
+        com = np.around(center_of_mass(data))
+        center = (int(com[1]), int(com[0]))
+        if verbose:
+            print("Center of mass at ({0}, {1})".format(center[1], center[0]))
 
     # make dimension-of-rawdata into list to account for rectangular n
     # Format of n -> n = [n_vert, n_horz]
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -77,7 +77,7 @@ def _abel_basex_core(data, center='auto', n='auto',
                 * tuple (x,y) -
                     the center of the image in (x,y) format
                 * If center='auto', it is set to (data.shape[1]//2, data.shape[0]//2)
-                * if n='com', it is set to the center of mass of data
+                * if center='com', it is set to the center of mass of data
       - basis_dir: string
             path to the directory for saving / loading the basis set coefficients.
             If None, the basis set will not be saved to disk. 
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -19,6 +19,7 @@ from scipy.ndimage import median_filter, gaussian_filter, center_of_mass
 from ._version import __version__
 from .tools.vmi import calculate_speeds
 from .tools.symmetry import center_image, center_image_asym
+from .tools.center import find_center
 
 #############################################################################
 # This is adapted from the BASEX Matlab code provided by the Reisler group.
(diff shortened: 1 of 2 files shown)


DanHickstein:
Hi @ldes89150!

Thanks for catching that bug! I like the idea to automatically set the center to the center of mass. We are in the middle of a code-restructuring, whereby all of the centering, symmetrizing, and integrating operations will be moved out of the individual transform functions and there will be a new, higher level set of functions (`abel.iabel` and `abel.fabel`) that will call the centering, transforming, and integrating functions. 

So, basically, what I am saying is that, in our new way of thinking, a function that centers the image by CoM should be located in a `abel.tools.center.center_by_com()`. We should probably move @stggh 's  [`abel.tools.vmi.find_image_center_by_slice()`](https://github.com/PyAbel/PyAbel/blob/master/abel/tools/vmi.py#L206) out of the tools/vmi.py file and into a new file tools/center.py.

By the way, do you work with cold atoms? So far, the @PyAbel/developers has @stggh and I who work with photoelectron/photoion spectra, @DhrubajyotiDas who works on flames, @rth who looks at laser plasma plumes. So, I think that if somebody from the cold-atom world is using PyAbel, then that is most of the major research applications covered! :)

ldes89150:
Hi @DanHickstein 

Yes, I'm working with cold atoms at Prof. Randall G. Hulet's research group at Rice University now. We are planning to use PyAbel to reconstruct the 3D structure of the atomic cloud in order to retrieve the information from the center of the cloud. I would be glad to contribute to this project :D

DhrubajyotiDas:
Hi @ldes89150 

Good catch on the `center = "auto"` bug! My apologies for that :D

ldes89150:
hi @DanHickstein ,

I create a find_center function at tools/center.py

DanHickstein:
Perhaps it can be called something like `find_center_com()`, or something to distinguish it from other methods of finding the center.

ldes89150:
It's a wrapper function which makes ppl to implement new find_center method more easily.

```python
func_method = {
    "auto": _find_center_by_center_of_image,
    "com": _find_center_by_center_of_mass,
}


def find_center(data, method="auto", verbose=True, **kwargs):
    return func_method[method](data, verbose=verbose, **kwargs)
```

DanHickstein:
That works. I would probably omit the leading underscore for the `find_center_by_center_of_mass` methods and let the users use them directly if they want.

DhrubajyotiDas:
I like the use of a dictionary to collect functions that do a similar thing.

ldes89150:
@DanHickstein sure

abel/basex.py
```diff
@@ -74,7 +76,8 @@ def _abel_basex_core(data, center='auto', n='auto',
                     the center column of the image
                 * tuple (x,y) -
                     the center of the image in (x,y) format
-                * If center='auto', it is set to (data.shape[0]//2, data.shape[1]//2)
+                * If center='auto', it is set to (data.shape[1]//2, data.shape[0]//2)
+                * if n='com', it is set to the center of mass of data
```
DhrubajyotiDas: Typo here, `* if n="com" it is set...` should be `* if center="com" it is set...`

ldes89150:
@DhrubajyotiDas  Thanks!

DanHickstein:
If there are no objections, I will merge this. 

@ldes89150, I am checking out your group's webpage - very cool work! (With very cool atoms!)

stggh:
@ldes89150 thanks for your contribution, the scipy...center_of_mass() is a handy function.
@PyAbel/developers

(Edit: This rant concerns `abel.tools.find_center()`, I dumped it in the wrong thread before)

I mean no offence, but in the interest of minimizing code bloat, why the double wrapper functions, why not (have the user) call scipy.ndimage.center_of_mass() directly?

find_center() also reverses the python axes order.

What is the purpose of **kwargs?

Sorry, the 34C Canberra summer heat, may be affecting my thought processes.

DanHickstein:
@stggh, I agree that it seems a little bit silly to wrap the scipy function, but I suppose the advantage of the find_center_by_center_of_mass function is a) that is rounds the output and (correctly) flips the order and b) that is suggests to the users that this might be a good option for centering their images. 

The find_center function provides a convenient way to flip though all of the options with the same interface. I think that we should move your find_image_center_by_slice function from vmi.py to center.py (and probably make it just find the center and not center the image). We should likely move the other centering functions from symmetry.py to center.py as well.

Flipping the CoM is because the abel.symmetry.center_image accepts the center in x,y format, instead of the Numpy convention of y,x. There are probably arguments to be made each way on this one. 

In the case of find_center, the **kwargs is used to accept additional keyword arguments and then pass them to find_center_by_center_of_image and find_center_by_center_of_mass. However, I'm not sure why we would want to make those functions accept any number of keyword arguments. (I.e., I don't know why this line contains **kwargs
```
def find_center_by_center_of_image(data, verbose=True, **kwargs):
```
Maybe @ldes89150 can let us know the motivation for this.

stggh:
The example from http://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.ndimage.measurements.center_of_mass.html appears to return, correctly, `(axis=0, axis=1)`:
```python
>>> a = np.array(([0,0,0,0],
                  [0,1,1,0],
                  [0,1,1,0],
                  [0,1,1,0]))
>>> from scipy import ndimage
>>> ndimage.measurements.center_of_mass(a)
(2.0, 1.5)
```
Perhaps, `abel.tools.symmetry.center_image()`, requires the coordinate flip? 

I am happy with the ideal of moving `abel.vmi.find_image_center_by_slice` to `abel.tools.center'.  This function currently returns the tuple `IMcentered, (y-shift, x-shift)` [The function doc may be incorrect - grr! I must check this function in any case].

Note, the origin/center should be a sub-pixel coordinate, and any image re-centering algorithm should handle a float shift. This is not the case at the moment.

DhrubajyotiDas:
>Note, the origin/center should be a sub-pixel coordinate, and any image re-centering algorithm should handle a float shift. 

This is only the case for even-shaped images, no? A regular 101 by 101 image would have it's center at (50,50)...

DanHickstein:
Well, @stggh's find_image_center_by_slice code is meant to operate on real data, where the center is likely not at an exact integer pixel. So, it finds the subpixel center. 

I think he's correct that the methods that are finding the center should just find the subpixel center and then the `center_image` function should re-interpolate to produce an image where the center is precisely in the middle of the center pixel (and the image has an odd number of pixels).

DanHickstein:
Oh, and yes, @stggh, the reason for the coordinate flip is that is what the `abel.tools.symmetry.center_image()` function wants.

DhrubajyotiDas:
Ah, I see. That's really neat! I'd like to see the results of the reinterpolation for some real data sets, maybe once I get around to tidying up and uploading a flame image.

stggh:
@DanHickstein  - perfectly stated.  An image should first be centered, before transforming. There *should* not be a need to pass `origin/center=(r, c)` to any Abel transform, or distribution function. This is not the case at the moment.

ldes89150:
@DanHickstein You answered most of my motivation :)

As for `**kwargs`, my idea is that maybe in the future, there will be another `find_center` method requiring additional parameters. To avoid exception happens at
```python
return func_method[method](data, verbose=verbose, **kwargs)
```
, it's better to add `**kwargs` at each method.

DanHickstein:
That makes sense to me that we would want the find_center function to accept kwargs, and pass them on to the other functions. However, if we are passing kwargs to a function that doesn't accept any additional kwargs, then wouldn't we want that function to throw an error, instead of happily accepting kwargs that don't apply to it?

I think that you might want the exception to raised in that case...

PR #85, opened 2016-02-13
add find_center_by_fit_gaussian

ldes89150:


DhrubajyotiDas:
@ldes89150 Great work with the Gaussian center-search code! 

1. Would it be possible to show an example of how this works, maybe for a contrived case?

2. Could you also add an unit test so we can make sure that changes to the code base in the future stay compatible with these centering methods?

Thanks again for your contributions!

stggh:
@ldes89150 thanks for your contribution, the `scipy...center_of_mass()`  is a handy function.
@PyAbel/developers 

I mean no offence, but in the interest of minimizing code bloat, why the double wrapper functions, why not (have the user) call  `scipy.ndimage.center_of_mass()` directly?

`find_center()` also reverses the python  `axes` order. 

What is the purpose of `**kwargs`?

Sorry, the 34C Canberra summer heat, may be affecting my thought processes.

DanHickstein:
This looks really good! Just add a few docstrings and comments to the code and it should be good-to-go.

stggh:
@ldes89150, @PyAbel/developers 
A few comments, that may help to improve the code, and the structure of PyAbel. NB I am not intending to be super critical, more open a discussion concerning PyAbel choices, and structure:
1. As mentioned (#83) the returned coordinates are reverse order `(axis=1, axis=0)`. This may be due to   an **incorrect** definition for:
>Flipping the CoM is because the abel.symmetry.center_image accepts the center in x,y format, instead of the Numpy convention of y,x. There are probably arguments to be made each way on this one. 

I think that `abel.symmetry.center_image()` should therefore be corrected. Two wrongs do not make a right ;-)
2.`center.py:  int(round)`.  IMO the float coordinate `(axis=0, axis=1)` should be returned, as discussed in #83. Note, also, `scipy.special.round()` does some interesting rounding:
```python
In [29]: from scipy.special import round
In [30]: round?
:
round(x)
Round to nearest integer
Returns the nearest integer to x as a double precision floating
point result.  If x ends in 0.5 exactly, the nearest even integer
is chosen.
```
i.e. `2.5 returns 2`, while `3.5 returns 4`?  Another reason to return the float values.
3. Code:
```python
def find_center_by_fit_gaussian(data, verbose=True, **kwargs):
    x = np.sum(data, axis=0)
    y = np.sum(data, axis=1)
    xc = fit_gaussian(x)[1]
    yc = fit_gaussian(y)[1]
    center = (int(round(xc)), int(round(yc)))
```
I find the array `sum` interesting. Q: Is the sum of Gaussian profiles a Gaussian profile? Would it be more accurate to use a slice? (I don't know).
Coming from a velocity-map imaging field, I find the concept of fitting a Gaussian profile opposite to the sorts of intensity profiles that I see. However, my BEC colleagues do detect Gaussian shaped blobs, so this function does make some sense (to me).
4. `abel.tools.math.py`: Gaussian functions are defined in `abel.tools.analytical.py`, can they be used instead? (Note, I am also guilty of duplicating `gauss` myself, in the same file!).
5. The dictionary for selecting methods is convenient, and I have used it myself (e.g. `examples/example_all_O2.py`), but I wonder, for the long term whether this will cause problems, because of the additional checks required, i.e. code to verify whether the function name is listed in the dictionary, and how to handle the error.  Our functions are nicely partitioned into `abel.tools.<files>` and so, can they not be called, using their real names, without the wrapper function?

Anyway, @ldes89150, thanks for the code, that drives this discussion.

ldes89150:
hi @stggh @PyAbel/developers ,

Thanks for your suggestions. Below are some of my opinions:

1. I think maybe we can add a flag called `round` in `find_center` functions, so we can determine to round it or not

2. Integral of a [2D Gaussian](https://en.wikipedia.org/wiki/Gaussian_function#Two-dimensional_Gaussian_function) along one axis would return since 2D Gaussian can be written in the f(x)g(y) form. (x and y can be completely separated.) It would be more accurate to fit 2 1D Gaussian instead of fitting 2D Gaussian directly if the noise of the image is large.

3. Before seeing your review, i didnt notice that there is an existing Gaussian in `analytical.py`. I would suggest to use my Gaussian function since it has more parameter, or we can just leave both of them there.

rth:
Nice work, @ldes89150 !

Just a minor comment, `round` is a builtin Python function, that already does what you want (and returns an int). So you don't actually need `scipy.special.round`  (which is an alias  for `numpy.round` BTW). Therefore I would suggest to just remove the `scipy.special` import, using the builtin function (i.e. replace `int(scipy.special.round` by `round`), and possibly avoid redefining  builtin's  (so use e.g. `round_func = .. ` instead of `round = .. `). 

Also I don't think that the  `_` function name follows usual Python naming conventions, maybe it could be replaced by `_f` to be more explicit?  Feel free to edit the Gaussian definition in  `analytical.py` to add  the missing constant parameter (with a default to 0, so it does not break other tests) if you want, or we can also keep both versions for now.
  
Otherwise this looks good, thanks a lot for contributing!

DanHickstein:
Good discussion here! 

@stggh, 

1) it's fine with me to switch the center_image function from the (x,y) to the (y,x) format. I thought that the (x,y) format would be more intuitive for most users, but if we think that the "row-column" indexing makes more sense, than that works too.

3) You are right, the Gaussian fit centering only makes sense if your data somewhat resembles a Gaussian. Our data does not, so it would likely not work well for us, but I can imaging many examples where the data has a peaked structure that might be adequately fit with a gaussian.

ldes89150:
@rth Thanks for your comment :D

stggh:
I think that we should stick with the standard numpy/scipy `axes`, which may in the end allow us to drop "vertical", and "horizontal" named directions, @rth's concern.

On the Gaussian shaped data:  I am not convinced that `data.sum(axis=0)` is a Gaussian (*) and that is why it may be better to use a slice `data[row//2+row%2-10: row//2+row%2+10, :].sum(axis=0)` and similarly for `axis=1`.  This breaks if the initial image is a long way off from center. Which is why I do like @ldes89150's `center_of_mass()`. Probably, simpler to stick with what is coded.

(*) `data.sum == a0*G(x0, y0) + a1*G(x0, y1) + a2*G(x0, y2) + ...` = too hard to picture

DanHickstein:
@stggh, while I don't really care what order we specify the center-coordinates (I think most users will find xy most intuitive when dealing with images, but don't mind the opposite way), I think that when we are talking about the symmetry, it will be very confusing to move away from calling it vertical/horizontal. The vertical/horizontal terminology is used throughout Numpy (for example, [numpy.flipud](http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.flipud.html)). And I think that most people will understand that vertical symmetry refers to top/bottom symmetry, where the row index represents the vertical direction. If we change those labels to "axis0symmetry", I'm not sure if people will know how to interpret this.

As far as the sum of gaussian goes, I think that as long as you are summing gaussian functions (summing in y) with the same width in the x-direction, then all you are doing is changing the amplitude. So, I think that if we are dealing with an elliptical gaussian (or just 2D gaussian with the same width in each x and y), then summing over one dimension should leave you with a gaussian. There is probably a simple way to show this analytically, but due to my aversion to analytical math, I've posted the code here that let's you play with this numerically:
https://gist.github.com/DanHickstein/824fb759fc8f97ec9b9e

stggh:
@DanHickstein excellent!  Thanks, for going to so much trouble.

ldes89150:
Hi I also found it's more reasonable to use
```python
(data.shape[1] // 2 + data.shape[1]%2, data.shape[0] // 2 + data.shape[0]%2)
```
rather than
```python
(data.shape[1] // 2, data.shape[0] // 2)
```
in finding center

DanHickstein:
Sorry, I caused some conflicts here with my latest merge :(. 

If you can add some documentation and take care of the conflicts, then we can go ahead and merge this. Thanks!

Issue #76, opened 2016-02-11
whole image vs single quadrant transform


stggh:
This *issue* follows on from a previous # that may have been closed, but I have further information. This is FYI, I will investigate further:

(1) I constructed a function to generate a pseudo O<sup>-</sup> photoelectron spectrum. O<sup>-</sup> photodetachment has well defined fine-structure transition positions and intensity ratios. The spectrum consists of 6 Gaussian transitions, very similar to `abel.tools.analytical.sample_image_dribinski()`. In my new-branch fork, the "_dribinski" name has been dropped, and added as a parameter, `sample_image(n=361, name="dribinski", sigma=2)` and
for O<sup>-</sup> `sample_image(n=1001, name="Ominus", sigma=3, temperature=200)`.

(2) The speed distribution of the raw simulated spectrum agrees pretty well with an experimental measurement, for both the whole image, and single quadrant, from `calculate_speeds()`. Middle panel of figure.

(3) The speed distribution of raw `Q0` quadrant (green line, thickened) -> `fabel_hansenlaw_transform()` -> `iabel_hansenlaw_transform()` (red line) agrees exactly. Right side panel.

(4) There is a persistent issue with the full quadrant analysis which slightly broadens the speed distribution. Blue line right panel.

![simulated_o-pes](https://cloud.githubusercontent.com/assets/10932229/12969462/e76e58c4-d0d4-11e5-93ac-2d11f5a6da4d.png)

DanHickstein:
@stgg, Thanks for making the new sample image! That looks really cool. 

Are you using the `get_` and `put_image_quadrants()` functions? I think that they may cause a one-pixel shift for the left quadrants in the case of a odd-sized image:

```python
In [54]: z = np.zeros((5,5)); z[1,1]=1; z[1,-2]=1; z[-2,1]=1; z[-2,-2]=1; print z
[[ 0.  0.  0.  0.  0.]
 [ 0.  1.  0.  1.  0.]
 [ 0.  0.  0.  0.  0.]
 [ 0.  1.  0.  1.  0.]
 [ 0.  0.  0.  0.  0.]]

In [55]: q = abel.tools.symmetry.get_image_quadrants(z, reorient=True)

In [56]: for quad in q: print quad
[[ 0.  0.  0.]
 [ 0.  1.  0.]
 [ 0.  0.  0.]]
[[ 0.  0.  0.]
 [ 0.  1.  0.]
 [ 0.  0.  0.]]
[...]

PR #87, opened 2016-02-14
Changes to the AbelTransform class

DanHickstein:
This is just a few-line change to the AbelTransform class to address @stggh's concern about having too much in the __init__ function. The functionality is the same.

### DIFF
--- a/abel/__init__.py
+++ b/abel/__init__.py
@@ -11,4 +11,5 @@ from . import direct
 from . import hansenlaw
 from . import three_point
 from . import tools
+from . import transform
 from .transform import iabel, fabel
--- a/abel/__init__.py
+++ b/abel/__init__.py
@@ -11,3 +11,4 @@ from . import direct
 from . import hansenlaw
 from . import three_point
 from . import tools
+from .transform import iabel, fabel
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -53,7 +53,7 @@ def iabel_basex(IM, dr=1.0, **kwargs):
     return _abel_basex_core(IM, dr=dr, **kwargs)
 
 def _abel_basex_core(data, center='auto', n='auto', 
-        nbf='auto',  basis_dir='./', calc_speeds=False, vertical_symmetry=True, dr=1.0, verbose=True):
+        nbf='auto',  basis_dir='./', calc_speeds=False, vertical_symmetry=False, dr=1.0, verbose=True):
 
     """ This function that centers the image, performs the BASEX transform (loads or generates basis sets), 
         and (optionally) calculates the radial integration of the image (calc_speeds)
@@ -107,7 +107,7 @@ def _abel_basex_core(data, center='auto', n='auto',
 
     if n == 'auto': n = list(data.shape)
 
-    if isinstance(center, str):
+    if type(center) == str or type(center) == unicode:
         center = find_center(data, method=center, verbose=verbose)
 
     # make dimension-of-rawdata into list to account for rectangular n
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -52,7 +52,7 @@ def iabel_basex(IM, dr=1.0, **kwargs):
     """
     return _abel_basex_core(IM, dr=dr, **kwargs)
 
-def _abel_basex_core(data, center='auto', n='auto', 
+def _abel_basex_core(data, center='image_center', n='auto', 
         nbf='auto',  basis_dir='./', calc_speeds=False, vertical_symmetry=False, dr=1.0, verbose=True):
 
     """ This function that centers the image, performs the BASEX transform (loads or generates basis sets), 
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -8,7 +8,7 @@ import numpy as np
 from time import time
 from math import exp, log, pow, pi
 from abel.tools.vmi import calculate_speeds
-from abel.tools.symmetry import  get_image_quadrants, put_image_quadrants
+from abel.tools.symmetry import get_image_quadrants, put_image_quadrants
 
 ###############################################################################
 # hansenlaw - a recursive method forward/inverse Abel transform algorithm 
@@ -108,6 +108,7 @@ _hansenlaw_transform_docstring = \
 
 _hansenlaw_docstring = \
     """ 
+    # THIS DOCSTRING IS NO LONGER USED. BUT, IT HAS A NICE IMAGE OF THE QUADRANTS AND SHOULD BE MOVED TO SOMEWHERE ELSE
     inverse Abel transform image
 
     options to exploit image symmetry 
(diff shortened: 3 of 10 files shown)


stggh:
Hmm, the solution is not quite as slick as I imagined, since the class object must be created, and then the `doTransform()` function called. However, I think this is the correct approach.

ldes89150:
@DanHickstein Thanks for your work on the AbelTransform class.

1. How do you thinks split forward and inverse transform into two classes (inherited  from the same base class) and use `__call__()` to do the transform. I believe people do transform in one direction at most of time.

2. My preference  is to put parameters like `method`, `center` ... at `__init__()` since the transformation with the same configuration may be applied to more than one images. It's a little annoying to copy lots of code while doing analysis in ipython notebook. How do you think?

stggh:
1. The forward and reverse transforms are the same function for `hansenlaw_transform` and `direct_transform`, with only the `(inverse=True/False)` parameter distinguishing the two function calls. IMO it is a pity to hide this fact behind `f/iabel...()` calls. It would make the situation worse having two separate classes.

2. @ldes89150, your preference is in favour of @DanHickstein's original class instance function structure. I believe (imagine!) each function call is a new class instance, so all the non default parameters must be still passed in each case?  This is no different in using the new class structure.
You have a good point that it would be nice to slot in another image, without creating a new class and resetting all processing options. This requires some more thought on the class structure.

DhrubajyotiDas:
>You have a good point that it would be nice to slot in another image, without creating a new class and resetting all processing options. 

@stggh I'd very much like this approach as well. My usual workflow involves doing the same pre, transform, and post-processing steps on batches of upto 1000 images, and this sort of structuring will simplify a lot of this code.

DanHickstein:
@ldes89150, we are currently splitting the transform into two classes, abel.iabel and abel.fabel, which are child classes to the abel.transform.AbelTransform. It's a little confusing, but it's in there... Does that work for you?

I am also fan on the one-line (or few line) Abel transform in order to minimize the size of the call to PyAbel when it's in the middle of a lot of other code. 

I like the idea that I could just type:

```transform = abel.iabel(IM, center='com', method='basex')```

We work with huge datasets as well, and so this is the kind of call that I want.

We could pretty easily provide the ability to switch the self.IM and run the transform again within a class, but I'm not quite sure that I see to situation where this would be easier than just calling the function again.

**Comment by DhrubajyotiDas** [comment]
@DanHickstein, just for clarity sake (and apologies in case this has already been mentioned before or in a different issue thread), could you provide an usage example of each of the two methods and what the structure of the source code looks like for each one. I'm not very familiar with the generic structure of these sorts of class-and-method relationships and I'm kinda lost here (purely due to my lack of python knowledge).

**Comment by DanHickstein** [comment]
@DhrubajyotiDas, The current source code is reasonably close to what I am proposing (there is a usage example in the `main()` function of the new `transform.py` file). I'd like to keep the number of lines that the user needs to type for a basic transform short. If they are concerned with the details, they are welcome to use all of the sub-functions that we have developed.

So, I would like a syntax that is just
```results = abel.iabel(image, center='auto', method='hansenlaw')
transformed_image = results.transform
```

Maybe I confused the issue by using a Python class in the first place. Really, I just want a function that bundles together the breaking up into quadrants, the abel transform, and the reassembling the image (and perhaps the centering which could be disabled by default). 

I am not exactly sure what @stggh is proposing, but I think that it's something with a more object-oriented flavor
```
transformObject = abel.AbelTransform(direction='inverse')
transformObject.IM = data
transformObject.method = 'hansenlaw'
transformObject.center = 'com'
transformObject.quadrants = (True, True, True, True)
transformObject.doTransform()
transformed_image = transformObject.transformData
```

But perhaps we can all agree on something in the middle:

```
transformObject = abel.iabel(IM)
transformObject.center_image('com')
transformed_image = transformObject.transform(method='hansenlaw')
```

That's only 2 lines for the basic transform. But, in my opinion, the first line is basically wasting the users time...

stggh:
@DanHickstein, nicely explained (as usual).  I see your point of view. 

I agree with the "middle" until @DanHickstein and @rth genius kicks in to identify a better solution.

stggh:
@PyAbel/developers, @ldes89150, actually, the solution is simple (*), use another class, with the "kitchen sink" instance, and inherit from the base class.

Something like:
```python
class PyAbel:
    def __init__ (self, IM):
      etc.

class AbelTransform(PyAbel):
     def __init__(self, IM, many other parameters):
```

Not tested.  Will try this later tonight.

(*) inspiration came while riding my push bike up a hill, into the western sun, against a strong westerly breeze.

ldes89150:
@DanHickstein It's my comment on having a class for both forward and inverse transform. Sorry for making you confused.

I prefer the functional object style. I think it can take advantage of oop and keep the code in a few lines

It looks like
```python
transform = abel.iabel(center='com', method='basex')  # Initialize the functional object with parameters
IM_transformed = transform(IM) #__call__() method
```

To process more than one image
```python
transform = abel.iabel(center='com', method='basex')  # Initialize the functional object with parameters
dataset_transformed = map(transform, dataset)
```

If someone wants to change the parameters in the `transform` object, just type
```python
transform.center = (30,10)
```

stggh:
I think that this issue may have created a mountain out of a molehill. Our efforts might better be spent working on the core PyAbel code.

Now that I understand:
>a function that bundles together the breaking up into quadrants, the abel transform, and the reassembling the image 

I am ok with the single call instance function, with bonus OO-bells. 
Sorry, for the noise.

PS: inheritance works, but adds another layer of indirection.

DanHickstein:
> I think that this issue may have created a mountain out of a molehill. Our efforts might better be spent working on the core PyAbel code.

Yes, I agree with this. The point of this function was to take some of the repeated code out of the transform functions, so we should now focus on streamlining the transform functions and making the documentation good. 

@ldes89150, point out an interesting option using the __call__ function. 

I will go ahead and finish this version of the AbelTransform class, which will continue to work in a similar way to how it works now. It should give us a good idea of how we can best structure the rest of the PyAbel code. In the future, it should be pretty easy for someone to re-design this top-level interface, or to build an alternate interface to the lower-level PyAbel functions.

DanHickstein:
I made some big changes to the structure of hansenlaw.py since the quadrants code is no longer needed. Now it's down to 200 lines. I may have botched the docstrings a little bit - sorry @stggh! 

I am going to go ahead and merge this, not because I think that it's perfect, but because I probably won't have much time to work in it for the next few days, and I'd like to open it up for the rest of you to make changes. 

@stggh - Can you check that I have not thrashed HansenLaw too badly?

stggh:
@DanHickstein  - a mammoth effort! Your are a machine.Thanks, for taking the trouble to transfer code for the new look PyAbel.  I like the new trim `hansenlaw.py`, which now cleanly demonstrates a powerful compact algorithm.

My cursory tests indicate that you have done no damage, and the code works as intended. 

***NB the bug fix #76 `put_image_quadrants()` has not made it into your branch****
```python
 -        Top    = np.concatenate((np.fliplr(Q[1][:-1,:-1]), Q[0][:-1,:]), axis=1)
-        Bottom = np.flipud(np.concatenate((np.fliplr(Q[2][:,:-1]), Q[3]), axis=1))
+        Top    = np.concatenate((np.fliplr(Q[1])[:-1,:-1], Q[0][:-1,:]), axis=1)
+        Bottom = np.flipud(np.concatenate((np.fliplr(Q[2])[:,:-1], Q[3]), axis=1))
```
ie. for an odd-size image the quadrant trim should occur after reorientation.

The defaults for the `AbelTransform` instance `vertical_symmetry=True, horizontal_symmetry=True, integrate=True` may cause a few surprises. They may better be left as `False`.

While I think of it, your comment (which I can't find due to the bombardment of changes ;-)) about averaging quadrants, rather than summing, is a very good point. After some thought, I think that averaging is the correct method.

The examples and unit tests will need to be changed. I can look at this tonight (in ~10 hours time).

PS:
My `new-branch` `abel.tools.analytical.py` has the modified `sample_image()` that includes O-, replacing `sample_image_dribinski()`. However, this code may be incorporated at a later merge.

rth:
@DanHickstein  Impressive work on `hansenlaw.py`! Good thing that you merged all of it, so we can all now work with this updated version.

Issue #89, opened 2016-02-15
Remove vertically symmetric BASEX


DanHickstein:
@DhrubajyotiDas, 

Where are you at with cleaning up basex.py? I am playing with it, and I am not sure if it returns a one-quadrant transform. Also, I think that the symmetric transform code is still there. 

We can get rid of the centering stuff as well.

I will merge a good draft of the `abel.iabel` function here in about an hour, so that should allow you to prune down the basex.py file.

DhrubajyotiDas:
@DanHickstein 

I am getting rid of the symmetric transform code today, and I expect the transition to one-quadrant transform to not take very long as well. I will work on it independently, and make sure it plays well with the `abel.iabel` function when you put it up.

DanHickstein:
Sounds good! Thanks @DhrubajyotiDas!

DhrubajyotiDas:
PR #94 is mostly cleanup of the symmetric code. I am currently working on enabling the single-quadrant functionality.

DhrubajyotiDas:
#94 takes care of this issue.

DanHickstein:
Great work @DhrubajyotiDas!

PR #94, opened 2016-02-16
Remove symmetric basex

DhrubajyotiDas:
This PR is related to #65 and #89. 

Briefly, the PyAbel codebase has a version of BASEX (call it **symmetric**) which accepts raw data with up/down symmetry (think of the letter B) and another version (**asymmetric**) which accepts raw data without this symmetry (think of the letter P). In the BASEX algorithm, raw data with up/down symmetry is *essentially* folded in half (i.e., the top and bottom halves are averaged), while raw data without up/down symmetry skips this step. Therefore, an equivalent result can be obtained by simply keeping the **asymmetric** code, and introducing a function to average the top and bottom halves when the user specifies that the raw data has up/down (or vertical) symmetry. This has the benefit of simplifying the codebase. 

In this PR, I've added a function to do the up/down averaging (I couldn't find an equivalent ready-made Numpy or Scipy function), and added a condition to the BASEX code which invokes this function when the `vertical_symmetry=True` parameter is passed to the basex function call. The rest is mostly cleanup of `basex.py` and `test_basex.py`. All code changes are on the backend; there should not be any UX changes.


DanHickstein:
Thanks for doing this @DhrubajyotiDas! 

I think that we already have the up/down averaging in the `tools.symmetry.get_image_quadrants` function. Are you planning (in this PR) to also modify basex to just accept a single quadrant?

DhrubajyotiDas:
@DanHickstein I just pushed #95 for the single quadrant code.

DhrubajyotiDas:
>I think that we already have the up/down averaging in the tools.symmetry.get_image_quadrants function. 

I hadn't realized that. I'll look to see if it is does what I intended with [updown_symmetry_rawdata()](https://github.com/DhrubajyotiDas/PyAbel/blob/remove-symmetric-basex/abel/tools/symmetry.py#L280), and if so, I'll modify this PR to reflect that.

If you think it's better to merge this first and file a separate PR for that, I'm okay with that as well, since I am fairly sure the code as-is is definitely functional (if a bit redundant).

rth:
@DhrubajyotiDas  This looks good, and would definitely simplify things. 

So now there are this PR and PR #94  which modify pretty much the same `basex.py` file. Should we merge one and then look at the other?

PR #95, opened 2016-02-17
One quadrant basex

DhrubajyotiDas:
This PR is a continuation of #80. The single-quadrant BASEX transform was not functional yet, but (hopefully) now is. 
```python
iabel_basex(IM, dr=1.0, **kwargs)
```
now takes a single (top-left) quadrant as `IM`, and returns the basex reconstruction of that quadrant. The cylindrical axis of symmetry is assumed to be the left-most column of `IM`.


DanHickstein:
This is a step in the right direction!

We still need to move the actual "basex transform" stuff from `_abel_basex_core` into `iabel_basex` and get rid of the centering function within basex. 

I will get to work on the "dictionary solution" for abel.iabel/fabel and hopefully everything will make more sense with that in place. 

I suppose that it makes sense to merge what you have already done here first...

DhrubajyotiDas:
> get rid of the centering function within basex

I am assuming you are referring to these lines in `basex.py`,
```python
image = center_image_asym(data, center_column = cx, 
                                n_vert = n[0], n_horz = n[1], verbose = verbose)
```
and
```python
if vertical_symmetry: # Average top and bottom halves of rawdata
        data = updown_symmetry_rawdata(data, center_row = cy)
``` 
?

DanHickstein:
Yeah, that line. And the fact that the function call still accepts "calc_speeds".

DanHickstein:
@DhrubajyotiDas, I'd like to merge these PRs so that I can start modifying the big "abel.transform" function. Can I just merge this one and forget 94?

DhrubajyotiDas:
@DanHickstein I'm not sure how the PR's would stack -- maybe you can merge #94 first, and then #95? Or is that the same as just merging #95?

DanHickstein:
If all of the changes in #94 are incorporated into #95, then I think we can just merge #95 and be fine (this looks to be the case). I guess to be safe I'll merge 94 first...

DanHickstein:
Okay to start merging, or are there more changes you'd like to make?

DhrubajyotiDas:
Not at the moment. Only thing outstanding is possibly getting rid of my vertical-averaging function and using the get_quadrant code. I'll file another PR for that.

PR #101, opened 2016-02-19
set_center

ldes89150:
example:
https://gist.github.com/ldes89150/eedb4088d31c7ba46afa

I'll add doc string maybe tomorrow. Before doing so, I would like to hear your opinions.


stggh:
Good work @ldes89150. I have a couple of points (then I may shut up ;-)  (*)):

1. (minor, but in case I forget) run `pep8` on the code - actually, very good, just a few compacted commas

2. `def set_center(data, center, crop='maintain_size', verbose=True):`
The code may be achieving the same as this [example](http://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.geometric_transform.html#scipy.ndimage.geometric_transform)
```python
>>> from scipy import ndimage
>>> a = np.arange(12.).reshape((4, 3))
>>> def shift_func(output_coords):
...     return (output_coords[0] - 0.5, output_coords[1] - 0.5)
...
>>> ndimage.geometric_transform(a, shift_func)
array([[ 0.   ,  0.   ,  0.   ],
       [ 0.   ,  1.362,  2.738],
       [ 0.   ,  4.812,  6.187],
       [ 0.   ,  8.263,  9.637]])
```
So perhaps (or similar):
```python
def set_center(data, center, crop='maintain_size', verbose=True):
     # delta stuff
     centered_data = ndimage.geometric_transform(data, shift_func)
    # crop stuff
```

In the end if your code works, then everything is fine. IMO it is a pity not to re-use available code, but we have to find it. See (*).

DanHickstein:
@ldes89150, thanks for doing this!! 

@stggh, I think that the output of the `set_center()` function with the `crop='maintain_size'` keyword can also be easily replicated by using the `scipy.ndimage.interpolation.shift` function, with a few extra lines to find the center. But the `set_center()` function supports several other cropping functionalities that would are a little bit tricker to figure out. So, I think that users will find this current `set_center` and `center_image` functions really useful.

DanHickstein:
I think that we should be able to merge this after you add the docstrings and fix the conflicts.

DanHickstein:
@ldes89150, this looks good! Can you modify to take care of the conflicts and then I can merge this?

ldes89150:
@DanHickstein sure :D

DanHickstein:
Looks good! I will merge this now!

Issue #96, opened 2016-02-17
iabel/fabel versus "transform"


DanHickstein:
Following up on https://github.com/PyAbel/PyAbel/issues/93: do people like the idea of having just one function?
```python
abel.transform(IM,direction='inverse')
abel.transform(IM,direction='forward')

```
rather than
```python
abel.iabel(IM)
abel.fabel(IM)
````

Should we make the same change for the individual transform functions?

It seems like this option would make the code slightly simpler. Of course, we would need to decide what to call the individual transform functions: `abel.basex.basex()` sounds a little bit funny.

@rth, @DhrubajyotiDas, @stggh, @ldes89150

**Comment by stggh** [comment]
[catching up on open issues, will @DhrubajyotiDas synchronize with me on this one, as per #99 ?]

I have no strong feelings either way, provided `function?` returns a parameter list.

I prefer the code/wrapper simplification of the one `abel.transform()` and this simplification should be continued to the individual transform functions. It would save many code lines.

**Comment by DhrubajyotiDas** [comment]
[Synchronization off by a few minutes, @stggh]

So, if I understand the difference between the two options,

```python
abel.transform(IM,direction='inverse')
abel.transform(IM,direction='forward')
```
would specify the transform algorithm (`basex` vs `three_point` vs `hansenlaw` etc) in the function call, whereas with
```python
abel.iabel(IM)
abel.fabel(IM)
```
would have `iabel` and `fabel` instantiated as objects of some sort of `AbelTransform` class, and therefore the transform algorithm and other relevant parameters would be specified when `iabel` and `fabel` are first set up in the user's code?

DanHickstein:
@DhrubajyotiDas, here I was asking about the general interface, if you all preferred a single function or to have two functions, one for the forward transform and one for the reverse.

In reality, we would basically have the same `abel.transform()` function under the hood, but then we would provide the abel.fabel() and abel.iabel() functions that will call the main transform function with `direction='inverse'/'forward'`.

So, either approach would be without classes.

Now that I describe it, it sounds like we are just putting wrappers around our actual code to try to make things slightly easier for the user, but making the code less readable. So, maybe I've finally convinced myself that we should do away with the `iabel/fabel` nomenclature (at least for the overarching `abel.transfom()` function.

[...]

PR #105, opened 2016-02-23
Added dummy direction parameter to basex and three_point

DhrubajyotiDas:
This PR adds the `direction` parameter to the function calls for `basex` and `three_point` as per the discussion in #96. 

@DanHickstein I hope I added the parameters in the right place.


DanHickstein:
Looks good to me!

Issue #103, opened 2016-02-23
Other improvements for Version 0.7


DanHickstein:
Hi @rth @stggh @rth @ldes89150,

Do you have any more suggestions for other improvements that we should make to PyAbel for version 0.7? There are a few issues still open, but I think that we are finishing those up. Other things that I would like to do:

1. Add links to the readthedocs.org documentation from the README file on the github main page.

2. Add links to the wiki from the README and give more information in the README in general.

3. Finish the wiki as well as we can. @DhrubajyotiDas did a great job on Three Point and Basex, and we should try to get HansenLaw and Direct up to a similar standard. 

4. Make sure that the docs look okay on Readthedocs.org. I know that @DhrubajyotiDas and @stggh fixed the errors, but we should proofread for other typos and for functions missing meaningful docstrings.

Any other suggestions? I'd like Version 0.7 to be fit for public consumption.

stggh:
from issue #96 legacy use of `f/i` in `f/iabel_<method>_transform()`    
>Question 2:
What do you guys think about the individual transform methods? Should we get rid of (for example) abel.hansenlaw.iabel_hansenlaw/fabel_hansenlaw and replace with a single method (perhaps called abel.hansenlaw.hansenlaw_transform() that accepts a direction='forward'/'inverse' keyword argument?
A while ago, @rth and I had decided that the iabel/fabel nomenclature was a good idea, but I think that I'm drifting away from this view. Largely, my opinion has changed based on the fact that the two methods that support the forward transform (direct and hansenlaw) are both calling the same function, just with an inverse=True/False keyword.
Thoughts on this?

I am in favour of dropping the `f/i` prefix, deleting the wrapper functions, leaving only the core transform routines:

`abel.<method>.<method>_transform(direction="forward/inverse")`

e.g.  `abel.hansenlaw.hasenlaw_transform()` for the half-image (or quadrant)  transform

NB 2nd related issue is whether the core (half-image) transforms should have another suffix rather than  `_transform()` which is too similar in name to the whole image Abel transform `abel.transform()`

DanHickstein:
@stggh: 

1. Seems like there is no objection to changing the function names to drop the i/fabel, so I think that we can go ahead with this. 

2. I think that `_transform` would be fine for naming the individual transform functions. Otherwise I think that `_half` or `_right` would better communicate that it's for a right half/quarter of an image. 

[...]

PR #108, opened 2016-02-25
Removal of f/iabel_<method> wrapper functions

stggh:
#96 and #103 removed `f/iabel_<method>` wrapper functions from all methods (except onion).

Repaired all `f/iabel` calls in `abel.<method>.py`, `abel.transform.py`, including doc strings, all unit tests, and all examples (phew!). 

All core functions are now `abel.<method>.<method>_transform()` with `direction="inverse/forward"` to transform right-side 1/2 or 1/4 image, and `abel.transform(method="<method>")` transforms the whole image.

IFAICT all unit tests and all examples run without error. 

NB:
 @rth `test_direct.py` required a scaling factor to pass the unit test `test_inverse_direct_gaussian()`
@DhrubajyotiDas `example_basex_step.py`  the curves do not overlap

@DanHickstein, @ldes89150 please check the examples. I am not happy with Xe `hansenlaw` vs `basex`

### DIFF
--- a/abel/direct.py
+++ b/abel/direct.py
@@ -29,46 +29,6 @@ except (ImportError, UnicodeDecodeError):
 ###########################################################################
 
 
-_direct_doctsting = \
-    """
-    This algorithm does a direct computation of the Abel transform
-
-      * integration near the singular value is done analytically
-      * integration further from the singular value with the Simpson
-        rule.
-
-    Parameters
-    ----------
-
-    fr : 1d or 2d numpy array
-        input array to which direct/inversed Abel transform will be applied.
-        For a 2d array, the first dimension is assumed to be the z axis and
-        the second the r axis.
-    dr : float
-        spatial mesh resolution           (optional, default to 1.0)
-    f : 1D ndarray
-        the spatial mesh (optional)
-    derivative : callable
-        a function that can return the derivative of the fr array
-        with respect to r
-        (only used in the inverse Abel transform).
-    inverse : boolean
-        If True inverse Abel transform is applied,
-        otherwise use a forward Abel transform.
-    correction : boolean
-        if False integration is performed with the Simpson rule,
-        the pixel where the weighting function has a singular value is ignored
-        if True in addition to the integration with the Simpson rule,
-        integration near the singular value is done analytically,
-        assuming a piecewise linear data.
-
-    Returns
-    -------
-    out : 1d or 2d numpy array of the same shape as fr
-        with either the direct or the inverse abel transform.
-    """
-
-
 def simpson_rule_wrong(f, x=None, dx=None, axis=1, **args):
     """
     This function computes the Simpson rule
@@ -116,10 +76,43 @@ def direct_transform(fr, dr=None, r=None, direction='inverse',
                      int_func=simpson_rule_wrong,
                      correction=True, backend='C'):
     """
-    Returns the forward or the inverse Abel transform of a function
-    sampled using direct integration.
+    This algorithm does a direct computation of the Abel transform
 
+      * integration near the singular value is done analytically
+      * integration further from the singular value with the Simpson
+        rule.
+
+    Parameters
+    ----------
+
+    fr : 1d or 2d numpy array
+        input array to which direct/inversed Abel transform will be applied.
+        For a 2d array, the first dimension is assumed to be the z axis and
+        the second the r axis.
+    dr : float
+        spatial mesh resolution           (optional, default to 1.0)
+    f : 1D ndarray
+        the spatial mesh (optional)
+    derivative : callable
+        a function that can return the derivative of the fr array
+        with respect to r
+        (only used in the inverse Abel transform).
+    inverse : boolean
+        If True inverse Abel transform is applied,
+        otherwise use a forward Abel transform.
+    correction : boolean
+        if False integration is performed with the Simpson rule,
+        the pixel where the weighting function has a singular value is ignored
+        if True in addition to the integration with the Simpson rule,
+        integration near the singular value is done analytically,
+        assuming a piecewise linear data.
+
+    Returns
+    -------
+    out : 1d or 2d numpy array of the same shape as fr
+        with either the direct or the inverse abel transform.
     """
+
     backend = backend.lower()
     if backend not in ['c', 'python']:
         raise ValueError
@@ -220,12 +213,6 @@ def _pyabel_direct_integral(f, r, correction, int_func=simpson_rule_wrong):
                     + np.arccosh(r[1:]/r[:-1])*(row[:-1] - f_r[i]*r[:-1])
 
     return out
-
-
-# append the same docstring to all functions
-#iabel_direct.__doc__ += _direct_doctsting
-#fabel_direct.__doc__ += _direct_doctsting
-direct_transform.__doc__ += _direct_doctsting
 
 
 def is_uniform_sampling(r):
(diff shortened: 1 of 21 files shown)


DanHickstein:
Great work @stggh!!

What exactly don't you like about the examples? Different output? or just the code is messy/different? I haven't looked at the examples in a while, so I suspect they are probably a bit out of date...

stggh:
All of the above!  ;-)    

I am not 100% certain about the outputs produced. I imaging that some of the `basex` curves should overlap.  `example_basex_gassian.py`:
example_basex_gaussian

 I am not sure what the output for `example_basex_step.py` should look like?
example_basex_step

I am not sure that the outputs of `hansenlaw` and `basex` agree for the Xe photoelectron data. I will check.

stggh:
@DanHickstein, @rth  AppVeyor build fail. I am not sure what this error means. Some server down?
```
Fetching package metadata: ...Error: HTTPError: 504 Server Error: GATEWAY_TIMEOUT for url: https://repo.continuum.io/pkgs/pro/win-32/repodata.json.bz2: https://repo.continuum.io/pkgs/pro/win-32/
 
Command exited with code 1
```

DanHickstein:
I restarted the AppVeyor build. Probably some glitch with their servers since one passed and one failed.

DanHickstein:
The basex_step looks crazy for sure. Probably it's passing the full double-step function instead of just the right side.

The different methods in `example_all` look remarkably similar! Interesting that the various methods would produce such comparable results...

stggh:
Hi @DanHickstein, welcome back.
`example_hansenlaw_basex.py` uses the full image transform `abel.transform()` methods, while `example_all_O2.py` uses the core routines `abel.<method>.<method>_transform()`, so, with the O2 data they provide a good test of whole-image vs single-quadrant analysis.

The agreement between methods is very-good, except on the spectroscopic scale where the transition peaks do not perfectly align. `abel.hansenlaw.py` has a subtle fudge, line 89:
```python
    #for some reason shift by 1 pixel aligns better? - FIX ME!
    if direction == "inverse":
        AIM = np.c_[AIM[:, 1:],AIM[:, -1]]
```
but, in reality the offset/shift may be a fraction of a pixel. I will look into this later.  

The Xe data analysis could be affected by the image center?

stggh:
The appveyor build appears to be asking for `https://repo.continuum.io/pkgs/free/win-64/setuptools-20.1.1-py34_0.tar.bz2` which is not present (at the URL).

stggh:
The results for the Ominus sample image is equally impressive `example_all_Ominus.py`:
example_all_ominus



This is interesting, the same (similar) program with dribinski sample image, has an odd background:
example_all_dribinski
The same issue may be the source of the Xe problem. 

Note: I have added these `example_all_Ominus.py` and `example_all_dribinski.py` and will remove them before the merge. I think the number and type of examples will need to be rationalized in any case.

stggh:
Today's issue: I have not correctly implemented passing transform args via `transform_options={}` dictionary parameter. Perhaps, **kwargs, may be better?

DanHickstein:
@stggh, you might have a point that we should switch to **kwargs. The dict is actually really nice, but the syntax is clunky and will probably confuse people. Though, if we provide examples, maybe it's not so bad. 

The syntax for the transform arguments is something like:
```python
abel.transform(data, method='basex', 
             transform_options={ 'basis_dir' : 'mydir', 'dr ': 0.1 } 
             centering_options={ 'center' : (250, 264) }
```
See if something like that works.

abel/direct.py
```diff
@@ -240,9 +223,9 @@ def _pyabel_direct_integral(f, r, correction, int_func=simpson_rule_wrong):
 
 
 # append the same docstring to all functions
-iabel_direct.__doc__ += _direct_doctsting
-fabel_direct.__doc__ += _direct_doctsting
-_abel_transform_wrapper.__doc__ += _direct_doctsting
+#iabel_direct.__doc__ += _direct_doctsting
+#fabel_direct.__doc__ += _direct_doctsting
+direct_transform.__doc__ += _direct_doctsting
```
DanHickstein: Is it now just one function? So should the docstring just live at the top of the function as normal?

abel/tests/test_direct.py
```diff
@@ -1,64 +1,64 @@
+#!/usr/bin/env python
```
DanHickstein: I think that we should not add this line to the tests. They should not be made executable or nosetests will not run them. So just run then with `python <test_file.py>` instead of executing directly from bash.

abel/tests/test_hansenlaw.py
```diff
@@ -1,20 +1,12 @@
+#!/usr/bin/env python
```
DanHickstein: same here.

abel/transform.py
```diff
@@ -201,6 +202,13 @@ def transform(
                 is not currently implemented
     """
 
+    abel_transform = {\
```
DanHickstein: This seems fine. You are just leaving the raising of exceptions to the individual transform functions. Do all of the functions raise an exception if the wrong direction is passed?

abel/transform.py
```diff
@@ -235,46 +243,26 @@ def transform(
...
-            if direction == 'forward':
-                raise ValueError('Coming soon...')
-            elif direction == 'inverse':
-                raise ValueError('Coming soon...')
-
     AQ0 = AQ1 = AQ2 = AQ3 = None
+
     # Inverse Abel transform for quadrant 1 (all include Q1)
-    AQ1 = selected_transform(Q1)
+    AQ1 = abel_transform[method](Q1, direction=direction, **transform_options)
```
DanHickstein: we could probably add another line:
```python
def selected_transform(Z):
    return abel_transform[method](Z, direction=direction, **transform_options)
```
and then simplify the following lines.

DanHickstein:
I had a look at the changes to code and it looks great to me! I made a few suggestions, but I think that it's nearly ready to be merged. 

Are there lingering issues? Just the basex examples not working?

Perhaps open a new Issue for these and hopefully we can recruit @DhrubajyotiDas or @ldes89150 to fix 'em.

DhrubajyotiDas:
I can take a crack at the basex examples once the outstanding PRs are merged.

DanHickstein:
Thanks @DhrubajyotiDas! 

I think that this can be merged soon. Give @stggh to respond to the code suggestions and then I think that we can merge this.

stggh:
@DanHickstein, good advice, and discussion. I will make the changes. Please allow me ~2 hours. Canberra has been suffering a heat wave for the final week of summer (seasons change on the 1st of the month in Oz), and my sleep has been poor.

@DhrubajyotiDas I will make that docstring change. tks.

stggh:
2 hours were not needed. I think the PR is ok, provided it passes the remote unit tests. My local tests pass. 

I really like the neat `transform_options=dict()` discussed in #109

DanHickstein:
Awesome! thanks for doing this!

So, we can stick with the `dict` approach for now? Still a tiny bit cumbersome for a user, but a whole lot more clear from the source-code viewpoint.

DanHickstein:
@stggh, this looks good to me. 

I think that you now have write access to the repo, so do you want to press the big green button?

stggh:
Yay, I've finally seen the "big green button", tks!

DanHickstein:
It's even bigger and greener than you imagine :)

Issue #109, opened 2016-02-25
abel.transform() transform arguments as dict or **kwargs


DanHickstein:
Currently, the `abel.transform()` function is defined like this:
```python
def transform(
    IM, direction='inverse', method='three_point', center='none',
        verbose=True, symmetry_axis=None,
        use_quadrants=(True, True, True, True),
        transform_options={}, center_options={}):
```

which means that it accepts `dict` arguments for the keyword arguments that will be passed to the transform functions and to the centering functions. Like this for example:

```python
AQ1 = abel_transform[method](Q1, direction=direction, **transform_options)
```
This leads to a somewhat clunky syntax if the user wants to specify arguments to these sub-functions:

```python
abel.transform(data, method='basex', 
             transform_options={ 'basis_dir' : '/Users/mydir', 'dr ': 0.1 } 
             center_options={ 'center' : (250, 264) } )
```

In https://github.com/PyAbel/PyAbel/pull/108#issuecomment-188706374, @stggh suggests that we switch to `**kwargs` instead. I think that this would look like:
```python
def transform(
    IM, direction='inverse', method='three_point', center='none',
        verbose=True, symmetry_axis=None,
        use_quadrants=(True, True, True, True),
        **kwargs):
```
where the transform functions are called with: 
```python
AQ1 = abel_transform[method](Q1, direction=direction, **kwargs)
```
The problem is that the `kwargs` for the centering function and the transform functions are lumped together into a big mass of `kwargs` and when we pass `**kwargs` to a transform function, we are also passing additional argument. Thus, each transform function would need to be able to accept any number of keyword arguments. So, we would need to define 

```python
def basex_transform(data, nbf='auto', basis_dir='./', dr=1.0, verbose=True,
                    direction='inverse', **kwargs):
```
where I have added **kwargs, which serves no purpose except to collect excessive keyword arguments that get passed.

The **kwargs method has the following advantages:

1) Allows users to specify the arguments in a simple `<parameter> = <value>` syntax. 

But, it has the following disadvantages compared to the `dict` method:

1) If a user supplies a keyword that is not relevant to a transform method, they should get an error. This occurs with the `dict` method, but does not occur with the **kwargs method. 

2) `**kwargs` method requires that we include `**kwargs` in all functions called by the `abel.transform()` method. This is not really a problem, but annoying and could mask errors. 

3) The `dict` method makes it explicit which arguments are being passed to which function, reducing confusion and bugs.

[...]

PR #114, opened 2016-02-28
center.py - return odd image width

stggh:
As per #109: 
>The maintain-size keyword would still make some sense, even if odd images are returned. The function could still approximately maintain the size and just return the closest odd sized image. I suggest that we add an odd_size=True keyword to set_center.

`abel.tools.center.py`:
* added parameter `odd_size=True` to `center_image()` to return an odd-width image if even.
  NB  simply trims the right-most column, before the center is determined. @PyAbel/developers this may not be the best option (but, I think it is ...).
* integrated `find_image_center_by_slice()`, so that it is called by the common centering function,`find_center(IM, method='slice')`, and, now only returns the center coordinate tuple.
* updated unit test `tests/test_tools_center.py`, finding center by `slice` and `com`, and verifying odd-width image  (tests could be expanded).
* updated examples, where the old `find_image_center_by_slice()` was used.

Added bonus:  `example_GUI.py` also updated, but still incomplete (menus not fully operational, but the button interface works). The matplotlib graphics window with button interface is not ideal.
Now provides sample images, and for the transformed image to be re-used as input. Still work in progress (and may not be the end look that we seek ...)
gui

### DIFF
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -98,7 +98,7 @@ def hansenlaw_transform(IM, dr=1, direction="inverse"):
 
              Image centre `o' should be within a pixel
              (i.e. an odd number of columns)
-             [Use abel.tools.center.find_image_center_by_slice () to transform] 
+             [Use abel.tools.center.center_image(IM, method='com', odd_size=True)] 
 
     dr : float
         Sampling size (=1 for pixel images), used for Jacobian scaling
--- a/abel/tests/test_tools_center.py
+++ b/abel/tests/test_tools_center.py
@@ -20,33 +20,21 @@ def test_center_image():
     # artificially displace center
     IMx = shift(IM, (-1,2))
 
-    # find vertical center
-    # radial range limits comparison to smaller radial range
-    IMy, offset = abel.tools.center.center_image(IMx, center="slice", 
-                               odd_size=True, radial_range=(1,120), axis=1)
+    # find_center using 'slice' method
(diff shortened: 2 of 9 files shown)


stggh:
Follow up comment re GUI. I now think that it would be better to have individual windows/frames for each plot graphic. That way more image detail may be visible in each window (expandable) e.g. a (red)cross-hair could identify the image center etc.

DanHickstein:
@stggh, this looks good to me! Thanks for fixing this! 

The GUI is looking pretty cool as well! 

I'm going to merge this.

DanHickstein:
@stggh, this looks good to me! Thanks for fixing this! 

The GUI is looking pretty cool as well! 

I'm going to merge this.

DanHickstein:
@stggh, one thing: we shouldn't make the unit tests executable. Executable files are not run by `nosetests`, so they just get skipped.

stggh:
@DanHickstein I will be "off-line" until about mid-night (your time), preparing ISMS abstracts. It is fortunate that Australia lives in the future.

Re unit tests. I am pretty sure that I followed that edict for all files in `abel/tests`. The `examples` are another matter.

DanHickstein:
> I will be "off-line" until about mid-night (your time)

Sounds good. I'll wait until you have a change to look at https://github.com/PyAbel/PyAbel/pull/115 before merging it.

Yeah, I don't think that the shebang (`#!/usr/bin/env`) screws anything up (though I'd rather not include it since it's not platform independent), it's just the file permissions need to be set to non-executable.

PR #116, opened 2016-03-01
Docstring improvements to hansenlaw and transform.py

DanHickstein:
I’m slowly working my way towards getting the documentation to look nice. Here are some improvements to the Hansenlaw and transform.py docstrings.

Plenty of work left to go!

### DIFF
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -134,7 +134,7 @@ def basex_core_transform(rawdata, M_vert, M_horz, Mc_vert,
     This is the internal function
     that does the actual BASEX transform. It requires 
     that the matrices of basis set coefficients be passed. 
-
+    
 
     Parameters
     ----------
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -33,72 +33,46 @@ from math import exp, log, pow, pi
 
 
 def hansenlaw_transform(IM, dr=1, direction="inverse"):
-    """
-    Forward/Inverse Abel transformation using the algorithm of:
-    Hansen and Law J. Opt. Soc. Am. A 2, 510-520 (1985).::
+    r"""
+    Forward/Inverse Abel transformation using the algorithm of
+    `Hansen and Law J. Opt. Soc. Am. A 2, 510-520 (1985) 
+    <http://dx.doi.org/10.1364/JOSAA.2.000510>`_ equation 2a: 
+    
+    
+    .. math:: f(r) = \frac{1}{\pi} \int_{r}^{\inf} \frac{g^\prime(R)}{\sqrt{R^2-r^2}} dR,
+    
+    where 
 
+    :math:`f(r)` is the reconstructed image (source) function,
+    :math:`g'(R)` is the derivative of the projection (measured) function
 
-                   ∞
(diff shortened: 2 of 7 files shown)


stggh:
@DanHickstein  excellent work! 

Minor notes:
1. missing '-' sign in the equation above. `\inf` should be LaTeX \infty.
2. (perhaps only in `hansenlaw.py`) Image center marked by 'o'. In the end, I think this should be something like: 'o' mid-pixel of the image center-line. It is only the `axis=1` size that needs to be odd. 
(Yes, I should alter this in the hansenlaw.py code block)
3. `abel.transform' has much better looking `direction` variables than the individual transforms ;-)
4. I find the `abel package' doc to be difficult to follow, because it is ordered alphabetically, rather than placing the key `<method>_transform()` functions at the top. Is it possible to reorder? (Perhaps, not important).
5. Some of the examples are not producing images, `hansenlaw Xenon`, basex xenon`, and `basex gaussian` may have a issue with the scalefactor @DhrubajyotiDas to examine. Are these linked to the current examples, as they all "should" work?

Anyway, this all looks fantastic. You have done a great job. I reckon merge, and then we may find any of the trivial remaining issues.

DanHickstein:
Hi @stggh, thanks for the comments! 

1. In Eq. 2a of Hansen and Law, the integral runs to inf, not -inf: 
image

2. Yes, we should be consistent in general to explain that each of the individual transform methods requires simply a right half on the image, not a right quadrant.

3. I'm not sure what you mean about the `direction` variables. Are you referring to the extra backticks around them? This makes them show up as a fixed width font when the docs are built. I think that it looks really cool, but maybe it makes the source code look stupid.
http://pyabel-dh-fork.readthedocs.org/en/latest/abel.html#module-abel.transform

With the backticks: 
image

Without: 
image

Scipy uses the fixed-width font approach for the various options for string keyword aguments (I think this must be where I got the idea). For example: http://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html#scipy.interpolate.griddata

They get rid of the single-quotes though, so maybe I should do this as well.

4. The methods are organized alphabetically by default, and supposedly you can change this by including 
```
autodoc_member_order = 'bysource'
```
in the `conf.py` file to organize by the order the functions appear in the source code. ([SO post](http://stackoverflow.com/questions/4692865/autodoc-params)) I think that organizing by the source-code order would be best, since we generally pick a halfway-logical ordering. 

However, this didn't seem to work when I built the docs locally. So, I'll open an issue and see if we can resolve this something in the future. 

5. The examples should all work. Some of them are broken and it would be good to get those fixed. @rth's unit tests for the examples should keep us honest. Though we may need to do away with some of the BASEX examples with the actual datasets, since they take a really long time to run.

DanHickstein:
I'm going to merge this so that I can move on with some more docstring edits.

DhrubajyotiDas:
>5. Some of the examples are not producing images, hansenlaw Xenon, basex xenon, and basex gaussian may have a issue with the scalefactor

My first guess is that the scalefactor is not the issue; however I will look at reproducing the example code.

Issue #119, opened 2016-03-01
abel.tools.analytical.sample_image() zooms with changing n


DanHickstein:
@stggh, I was playing with the `example_all_dribinski.py` and I realized that when you change `n`, the functions zoom in and out: 

Here is `n=301` and `n=1001`:

![image](https://cloud.githubusercontent.com/assets/1107796/13433504/62b6621a-df8f-11e5-947b-fcb94ca5e30b.png)

I thought that we fixed this behavior with the previous `sample_image_dribinski` function, but I think that it's back now that we have changed to the `sample_image(name='dribinski')` function. 

The `Ominus` sample image has different behavior: as `n` is changed, the width of the peaks changes (well, the width remains the same in pixels, but as a fraction of the whole image the width changes.)

I think that I have fixed this by changing
```python
    x = np.linspace(-n2, n2, n)
```
to
```python
     x = np.linspace(-n2, n2, n)

    if name=='dribinski': 
        x = x * 180/n2
    elif name=='Ominus':
        x = x * 501/n2 
```

Also, I made `rfact` default to 1. 

I'll send this as a PR and you can see if you like it.

DanHickstein:
I think that this is now fixed with #120, so I'll close this.

Issue #117, opened 2016-03-01
Should we recast all input arrays to float64?


DanHickstein:
I was having some weird issues this morning with the `example_hansenlaw_Xe.py`. The output were negative and I could not figure out what was going on for a long time. Then I realized that the data was being read from the image as `unit16`. During the transform, it was apparently overflowing to negative values. This seems like it might happen frequently with users loading images.

We could probably make some argument about trying to preserve the bit depth of the users image, but I think that the best thing for us to do is to recast all input images to float64 and move on with out lives. If the users want to save their data, they will need to recast to whatever bit depth they wish to save the images. 

For now, I'm going to add another kwarg to the `abel.transform` function: `recast_as_float64=True`. That seems like the safest thing to do for now. 

Any other ideas? @PyAbel/developers

stggh:
@DanHickstein good catch. The Abel transform for Xe had me puzzled. 

Note (for the record) that the inverse Abel transform may, and should, produce negative values, particularly, near the image center (line). Such values are just an over correction, but essential to obtain full the correct inversion profile. The old `basex.exe` forced positive image values, incorrectly truncating the image inversion.

DanHickstein:
Good point @stggh. In this case, almost all of the values were negative, so I knew something must be wrong :)

DanHickstein:
So, I think that the current solution of casting to `float64` in `abel.transform()` should take care of most users. Probably this issue will pop up again when someone tried to run one of the individual transforms methods with other data. One option would be for the individual methods to print a warning of the datatype is not `float`. I'll close this for now.

PR #120, opened 2016-03-01
recast images to float64 and keep sample_image constant zoom

DanHickstein:
This addresses issues https://github.com/PyAbel/PyAbel/issues/117 and https://github.com/PyAbel/PyAbel/issues/119. 

It adds an options (which defaults to True) in `abel.transform` to recast the input image to `float64` to avoid problems with overflowing numbers in common image datatypes like `unit16`. 

Also, this changes the behavior of the sample_image function to keep the sample images the same size and shape as `n`, the number of points, is varied. I think that this is the desired behavior, because it's nice to compare how the transforms work as the sampling is changed, and this is only really possible if the image is remaining the same and only the sampling interval is changing.

The reason that I was working on this is that I would like to change all of the examples to use smaller images so that they can run quickly and be included in the documentation. `n=1001` can take a few minutes for basex to build the basis set.  `n=301` runs quickly enough, and we should probably pick this as the standard for our examples.

@stggh or @DhrubajyotiDas, If you have time, can you make more of the examples run efficiently and then add them to the documentation by including a new `doc/example_<method>.rst` file and then appending the example name to the `doc/examples.rst` file?

### DIFF
--- a/examples/example_hansenlaw_Xe.py
+++ b/examples/example_hansenlaw_Xe.py
@@ -41,17 +41,17 @@ output_plot  = name + '_comparison_HansenLaw.pdf'
 # Step 1: Load an image file as a numpy array
 print('Loading ' + filename)
 #im = np.loadtxt(filename)
-im = plt.imread(filename)
+im = plt.imread(filename) 
 (rows,cols) = np.shape(im)
 print ('image size {:d}x{:d}'.format(rows,cols))
 
-#im = abel.tools.center.center_image (im, (340,245))
 
 # Step 2: perform the Hansen & Law transform!
 print('Performing Hansen and Law inverse Abel transform:')
 
 recon = abel.transform(im, method="hansenlaw", direction="inverse", 
-                       symmetry_axis=None, verbose=True)['transform']
+                       symmetry_axis=None, verbose=True, center=(240,340))['transform']
+                       
 r, speeds = abel.tools.vmi.angular_integration(recon)
 
 
--- a/abel/tools/analytical.py
+++ b/abel/tools/analytical.py
@@ -8,7 +8,7 @@ import scipy.constants as const
 # impementations.
 
 
-def sample_image(n=361, name="dribinski", sigma=2, temperature=200):
+def sample_image(n=361, name="dribinski", sigma=3, temperature=200):
     """
     Sample images, made up of Gaussian functions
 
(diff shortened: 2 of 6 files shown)


stggh:
@DanHickstein, I agree that the sample images should maintain their sizes. Thanks, for finding this bug. However, `hansenlaw` performs much better with a decent sample size. `n=301` puts in it in a bad light, possibly due to a poor gradient estimate. Typical experimental camera images are <math>2<sup>m</sup></math>, such as `n=1024, 2048'. In my own experiment I may use `n=8196'.  So, IMO, the 'read world' example should be `n=1001`, which is a compromise for both `basex` and `hansenlaw` algorithms.

DanHickstein:
@stggh, Yes, you are completely right that full benchmarking should be done under realistic conditions. I see the examples more as demonstrations of the PyAbel syntax more than a formal comparisons of the methods. Even with a size of `301` the obvious conclusion from this example is that to a good first approximation, the methods perform nearly identically:

image

If the examples take too long to run, then the images can't be included in the online documentation, which would be unfortunate. On the other hand, I see your point that some of the examples could be fairly effective benchmarks of the transform techniques.

At some point, we should do some formal benchmarks and draw some real conclusions about the strengths and weaknesses of the various transform methods. 

Perhaps one solution at this point is to make the examples use a small size (n=301) but to include a comment, something to the effect of:
``` python
n=301 
# note that a small number of point is used here to achieve an quick run time for this example
# n should be increased to 1001 or higher to allow for an effective comparison of transform methods
```

It's also an option that we just omit some of the examples from the documentation, which is what we are currently doing. But I think that most users will only find the examples if we can get them onto the readthedocs, so there is a strong incentive to get them up there.

stggh:
Good argument (as usual). I totally agree.

DhrubajyotiDas:
I think this is a great compromise as well.

examples/example_hansenlaw_Xe.py
```diff
@@ -41,17 +41,17 @@
 # Step 1: Load an image file as a numpy array
 print('Loading ' + filename)
 #im = np.loadtxt(filename)
-im = plt.imread(filename)
+im = plt.imread(filename)
```
DhrubajyotiDas: Extra space after line - this will cause PEP8 to complain...

PR #127, opened 2016-03-02
Fix for broken three-point example

DhrubajyotiDas:
@DanHickstein If you could verify that the three-point example compiles properly with these changes, that'd be great.

### DIFF
--- a/examples/example_three_point.py
+++ b/examples/example_three_point.py
@@ -3,6 +3,7 @@ import matplotlib.pyplot as plt
 import abel 
 
 n = 101
+center = n//2
 r_max = 4
 r = np.linspace(-1*r_max, r_max, n)
 dr = r[1]-r[0]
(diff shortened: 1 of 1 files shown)


DanHickstein:
Yeah, it looks pretty sweet! 

You can check for yourself: go to the docs directory and use 
```bash
make clean
make html
```
or, try the new livehtml
```
make livehtml
```

DhrubajyotiDas:
I'm not seeing the three_point example in there. Do I need to add something to an index somewhere?

DanHickstein:
Oh, the three-point example was living in the transform_methods/three_point.rst. So, it appears as part of the description of the three-point method. It should also be included with the other examples. To do that, you need to add a line to `examples.rst` and then add an `example_three_point.rst`, following the convention of the other examples.

DhrubajyotiDas:
Ah, I see. It looks really neat! I'm not familiar with the `liveupdate` thing - if we change the source code, will this update in real time? (If so, that's pretty freaking cool!)

For some reason, the example plot crops out the legend of the second subplot
image

DanHickstein:
Hmmm, it does crop out the legend. Probably because it's outside of the region of the axes.

And yes, the `make livehtml` uses sphinx-autobuild to detect when you save changes to any of the files and to automatically re-build the documentation. It typically works really well, but occasionally it will have a glitch and you will need to `make clean` and `make livehtml` again. I included this info here: https://github.com/PyAbel/PyAbel/blob/master/CONTRIBUTING.rst#documentation

PR #134, opened 2016-03-04
passing center_options in transform

ldes89150:
fix a small bug


DanHickstein:
This makes sense. Thanks for catching this!

PR #137, opened 2016-03-05
doc update for hansenlaw, vmi

stggh:
Updated docs for `doc/transform_methods/hansenlaw.rst` and `abel.tools.vmi.py`. 

NB I renamed `calculate_angular_distributions()` to `radial_integration()` to better align with the partner method `angular_integration()`.  Modified examples to suit.

Hmm, some of the listed changes were just a 80 column reformat, sorry.

@DanHickstein `livehtml` is neat, and certainly speeds up the editing process.

Edit: NB changed name of parameter `method` to `center` in `abel.tools.center.find_center(image, center="<method>")` to match `center_image(image, center="<method>")`

### DIFF
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -39,7 +39,7 @@ def hansenlaw_transform(IM, dr=1, direction="inverse"):
     <http://dx.doi.org/10.1364/JOSAA.2.000510>`_ equation 2a: 
     
     
-    .. math:: f(r) = \frac{1}{\pi} \int_{r}^{\inf} \frac{g^\prime(R)}{\sqrt{R^2-r^2}} dR,
+    .. math:: f(r) = -\frac{1}{\pi} \int_{r}^{\inf} \frac{g^\prime(R)}{\sqrt{R^2-r^2}} dR,
     
     where 
 
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -39,7 +39,7 @@ def hansenlaw_transform(IM, dr=1, direction="inverse"):
     <http://dx.doi.org/10.1364/JOSAA.2.000510>`_ equation 2a: 
     
     
-    .. math:: f(r) = -\frac{1}{\pi} \int_{r}^{\inf} \frac{g^\prime(R)}{\sqrt{R^2-r^2}} dR,
+    .. math:: f(r) = -\frac{1}{\pi} \int_{r}^{\infty} \frac{g^\prime(R)}{\sqrt{R^2-r^2}} dR,
     
     where 
 
--- a/abel/tools/symmetry.py
+++ b/abel/tools/symmetry.py
@@ -6,10 +6,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import numpy as np
-from scipy.ndimage import map_coordinates
-from scipy.ndimage.interpolation import shift
-from scipy.optimize import curve_fit, minimize
-
 
 def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
                         use_quadrants=(True, True, True, True)):
@@ -101,6 +97,7 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
 
     n, m = IM.shape
 
+    # odd size increased by 1
     n_c = n // 2 + n % 2
     m_c = m // 2 + m % 2
 
@@ -139,7 +136,7 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
     return Q0, Q1, Q2, Q3
 
 
-def put_image_quadrants(Q, odd_size=True, symmetry_axis=None):
+def put_image_quadrants(Q, original_image_shape, symmetry_axis=None):
     """
     Reassemble image from 4 quadrants Q = (Q0, Q1, Q2, Q3)
     The reverse process to get_image_quadrants(reorient=True)
@@ -162,8 +159,8 @@ def put_image_quadrants(Q, odd_size=True, symmetry_axis=None):
             | Q2  *  | *   Q3 | 
             +--------+--------+                 
 
-    odd_size: boolean
-       Whether final image is odd or even pixel size
+    original_image_shape: tuple
+       (rows, cols)
        odd size will trim 1 row from Q1, Q0, and 1 column from Q1, Q2
 
     symmetry_axis : int or tuple
@@ -202,15 +199,18 @@ def put_image_quadrants(Q, odd_size=True, symmetry_axis=None):
         Q2 = Q1
         Q3 = Q0
 
-    if not odd_size:
-        Top = np.concatenate((np.fliplr(Q0), Q1), axis=1)
-        Bottom = np.flipud(np.concatenate((np.fliplr(Q2), Q3), axis=1))
-    else:
-        # odd size image remove extra row/column added in get_image_quadrant()
-        Top = np.concatenate(
-                    (np.fliplr(Q[1])[:-1, :-1], Q[0][:-1, :]), axis=1)
-        Bottom = np.flipud(
-                    np.concatenate((np.fliplr(Q[2])[:, :-1], Q[3]), axis=1))
+    if original_image_shape[0] % 2 == 1:
+        # odd-rows => remove duplicate bottom row of Q1, Q0
+        Q0 = Q0[:-1, :]
+        Q1 = Q1[:-1, :]
+
+    if original_image_shape[1] % 2 == 1:
+        # odd-columns => remove duplicate first column of Q1, Q2
+        Q1 = Q1[:, 1:]
+        Q2 = Q2[:, 1:]
+
+    Top = np.concatenate((np.fliplr(Q1), Q0), axis=1)
+    Bottom = np.flipud(np.concatenate((np.fliplr(Q2), Q3), axis=1))
 
     IM = np.concatenate((Top, Bottom), axis=0)
 
--- a/abel/tools/symmetry.py
+++ b/abel/tools/symmetry.py
@@ -199,13 +199,13 @@ def put_image_quadrants(Q, original_image_shape, symmetry_axis=None):
         Q2 = Q1
         Q3 = Q0
 
-    if original_image_shape[0] % 2 == 1:
-        # odd-rows => remove duplicate bottom row of Q1, Q0
+    if original_image_shape[0] % 2:
+        # odd-rows => remove duplicate bottom row from Q1, Q0
         Q0 = Q0[:-1, :]
         Q1 = Q1[:-1, :]
 
-    if original_image_shape[1] % 2 == 1:
-        # odd-columns => remove duplicate first column of Q1, Q2
+    if original_image_shape[1] % 2:
+        # odd-columns => remove duplicate first column from Q1, Q2
         Q1 = Q1[:, 1:]
         Q2 = Q2[:, 1:]
 
--- a/abel/tools/symmetry.py
+++ b/abel/tools/symmetry.py
@@ -126,12 +126,15 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
                   (use_quadrants[0] + use_quadrants[1])
         Q2 = Q3 = (Q2*use_quadrants[2]+Q3*use_quadrants[3])/\
                   (use_quadrants[2] + use_quadrants[3])
+        # quadrants now all valid
+        use_quadrants = (True, True, True, True)
 
     if 1 in symmetry_axis:  # horizontal axis image symmetry
         Q1 = Q2 = (Q1*use_quadrants[1]+Q2*use_quadrants[2])/\
                   (use_quadrants[1] + use_quadrants[2])
         Q0 = Q3 = (Q0*use_quadrants[0]+Q3*use_quadrants[3])/\
                   (use_quadrants[0] + use_quadrants[3])
+
 
     return Q0, Q1, Q2, Q3
 
@@ -180,7 +183,7 @@ def put_image_quadrants(Q, odd_size=True, symmetry_axis=None):
              None             0              1           (0,1)
 
             Q1 | Q0        Q1 | Q1        Q1 | Q0       Q1 | Q1
-            ----o----  or  ----o----  or  ----o----  or ----o----
+           ----o----  or  ----o----  or  ----o----  or ----o----
             Q2 | Q3        Q2 | Q2        Q1 | Q0       Q1 | Q1
 
     """
--- a/abel/tools/symmetry.py
+++ b/abel/tools/symmetry.py
@@ -41,7 +41,7 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
     Notes
     -----
      
-    The symmetry_axis keyword adds quadrants like this: ::
+    The symmetry_axis keyword averages quadrants like this: ::
 
          +--------+--------+
          | Q1   * | *   Q0 |
@@ -54,9 +54,9 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
          +--------+--------+                 
 
         symmetry_axis = None - individual quadrants
-        symmetry_axis = 0 (vertical) - sum Q0+Q1, and Q2+Q3
-        symmetry_axis = 1 (horizontal) - sum Q1+Q2, and Q0+Q3
-        symmetry_axis = (0, 1) (both) - combine and sum all 4 quadrants
+        symmetry_axis = 0 (vertical) - average Q0+Q1, and Q2+Q3
+        symmetry_axis = 1 (horizontal) - average Q1+Q2, and Q0+Q3
+        symmetry_axis = (0, 1) (both) - combine and average all 4 quadrants
     
 
     The end results look like this: ::
@@ -95,6 +95,27 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
         # if the user supplies an int, make it into a 1-element list:
         symmetry_axis = [symmetry_axis]
 
+    if ((symmetry_axis == [None] and (use_quadrants[0]==False
+                                   or use_quadrants[1]==False 
+                                   or use_quadrants[2]==False
+                                   or use_quadrants[3]==False)) or 
+        # at least one empty
+        (symmetry_axis == [0] and use_quadrants[0]==False and 
+                                  use_quadrants[1]==False) or # top empty
+        (symmetry_axis == [0] and use_quadrants[2]==False and 
+                                  use_quadrants[3]==False) or # bot empty
+        (symmetry_axis == [1] and use_quadrants[1]==False and
+                                  use_quadrants[2]==False) or # left empty
+        (symmetry_axis == [1] and use_quadrants[0]==False and 
+                                  use_quadrants[3]==False)    # right empty
+                                                                or
+        (symmetry_axis == (0, 1) and not np.all(use_quadrants))
+        ): 
+        raise ValueError('At least one quadrant would be empty.'
+                         ' Please check symmetry_axis and use_quadrant'
+                         ' values to ensure that all quadrants will have a'
+                         ' defined value.')
+
     n, m = IM.shape
 
     # odd size increased by 1
@@ -118,13 +139,17 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
             'In order to add quadrants (i.e., to apply horizontal or \
             vertical symmetry), you must reorient the image.')
 
+    if symmetry_axis==(0, 1):
+        Q = (Q0 + Q1 + Q2 + Q3)/np.sum(use_quadrants)
+        return Q, Q, Q, Q
+
     if 0 in symmetry_axis:   #  vertical axis image symmetry
-        Q0 = Q1 = Q0 + Q1
-        Q2 = Q3 = Q2 + Q3
+        Q0 = Q1 = (Q0 + Q1)/(use_quadrants[0] + use_quadrants[1])
+        Q2 = Q3 = (Q2 + Q3)/(use_quadrants[2] + use_quadrants[3])
 
     if 1 in symmetry_axis:   # horizontal axis image symmetry
-        Q1 = Q2 = Q1 + Q2
-        Q0 = Q3 = Q0 + Q3
+        Q1 = Q2 = (Q1 + Q2)/(use_quadrants[1] + use_quadrants[2])
+        Q0 = Q3 = (Q0 + Q3)/(use_quadrants[0] + use_quadrants[3])
 
     return Q0, Q1, Q2, Q3
 
--- a/abel/tools/symmetry.py
+++ b/abel/tools/symmetry.py
@@ -109,7 +109,7 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
         (symmetry_axis == [1] and use_quadrants[0]==False and 
                                   use_quadrants[3]==False)    # right empty
                                                                 or
-        (symmetry_axis == (0, 1) and not np.all(use_quadrants))
+        (symmetry_axis == (0, 1) and not np.any(use_quadrants))
         ): 
         raise ValueError('At least one quadrant would be empty.'
                          ' Please check symmetry_axis and use_quadrant'
--- a/abel/tools/symmetry.py
+++ b/abel/tools/symmetry.py
@@ -161,7 +161,12 @@ def put_image_quadrants(Q, original_image_shape, symmetry_axis=None):
 
     original_image_shape: tuple
        (rows, cols)
-       odd size will trim 1 row from Q1, Q0, and 1 column from Q1, Q2
+
+       reverses the padding added by `get_image_quadrants()` for odd-axis sizes
+
+       odd row trims 1 row from Q1, Q0
+
+       odd column trims 1 column from Q1, Q2
 
     symmetry_axis : int or tuple
        impose image symmetry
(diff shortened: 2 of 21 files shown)


stggh:
**Issue**:  `abel.tools.symmetry.put_image_quadrants()` does not know about the original image size.

 `get_image_quadrants()` correctly handles any shaped image, adding one row or column if the respective axis size is odd. The reassembly, `put_image_quadrants()` depends on the odd-evenness of **both** axes. The one passed parameter, `odd_size=True`  assumes, incorrectly, a square original image.

Ideas:
1. tuple  `odd_axis=None, 0, 1, or (0,1)`
2. We encapsulate `get/put_image_quadrants()` into a class, keeping the knowledge of the original axis sizes

I prefer 2.

stggh:
Point 3. has been implemented. 
```python
[Steve:.+Abel/abel/tests]$ python test_tools_symmetry.py 
test image
[[ 1.  1. -1.  0.  0.]
 [ 1.  1. -1.  0.  0.]
 [-1. -1. -1. -1. -1.]
 [ 2.  2. -1.  3.  3.]
 [ 2.  2. -1.  3.  3.]]

reoriented quadrant Q0
[[-1.  0.  0.]
 [-1.  0.  0.]
 [-1. -1. -1.]]

reoriented quadrant Q1
[[-1.  1.  1.]
 [-1.  1.  1.]
 [-1. -1. -1.]]

reoriented quadrant Q2
[[-1.  2.  2.]
 [-1.  2.  2.]
 [-1. -1. -1.]]

reoriented quadrant Q3
[[-1.  3.  3.]
 [-1.  3.  3.]
 [-1. -1. -1.]]

reassembled image
[[ 1.  1. -1.  0.  0.]
 [ 1.  1. -1.  0.  0.]
 [-1. -1. -1. -1. -1.]
 [ 2.  2. -1.  3.  3.]
 [ 2.  2. -1.  3.  3.]]
```

stggh:
@DhrubajyotiDas and @DanHickstein 
Given the timing issues with the basis set methods (`basex` and `three_point`) it may be a good idea to standard all the basis set examples to one size image. The (random) examples that I have run produced: 
```
basex_basis_151_301_151_151.npy
basex_basis_1_501_1_251.npy
basex_basis_251_501_251_251.npy
basex_basis_256_511_256_256.npy
basex_basis_260_693_260_347.npy
basex_basis_512_1023_512_512.npy
```
eg. `example_hansenlaw_basex.py` @DhrubajyotiDas zoom factor may be `zoom(im, 0.4892578125)` to produce a 501x image size, which then matches the sample image sizes. I have made this change to this PR.

DhrubajyotiDas:
@stggh that's a great point. I'll look into standardizing the basis set examples.

stggh:
This PR is ready to be merged. The main changes:
- Updated the text in `doc/transform_methods/hansenlaw.rst`
- `abel.tools.center.center_image()` variable name change of `method=` to `center=`, to match the other methods.
- `abel.tools.symmetry.put_image_quadrants()` now reassembles non-square images, `original_image_shape` variable replaced the single `odd_size` that applied to square images
- **NB** `abel.tools.symmetry.get_image_quadrants()` I removed averaging. The averaging denominator complicates the use of `use_quadrants=(boolean tuple)`, to such an extent that it is virtually useless.  e.g. it is impossible to map a single quadrant, Qi, to the whole image. I have reached the conclusion that it is better not to average. Removed averaging/average from docs. 
- updated the unit test `test_tools_symmetry` and examples to reflect the above changes

abel/hansenlaw.py
```diff
@@ -39,7 +39,7 @@ def hansenlaw_transform(IM, dr=1, direction="inverse"):
     <http://dx.doi.org/10.1364/JOSAA.2.000510>`_ equation 2a: 
     
     
-    .. math:: f(r) = \frac{1}{\pi} \int_{r}^{\inf} \frac{g^\prime(R)}{\sqrt{R^2-r^2}} dR,
+    .. math:: f(r) = -\frac{1}{\pi} \int_{r}^{\infty} \frac{g^\prime(R)}{\sqrt{R^2-r^2}} dR,
```
DanHickstein: Ah, yes, I see that you were saying about `\infty` before - sorry for the confusion.

abel/tools/symmetry.py
```diff
@@ -122,21 +119,17 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
...
-                  (use_quadrants[0] + use_quadrants[3])
+    if 1 in symmetry_axis:   # horizontal axis image symmetry
+        Q1 = Q2 = Q1 + Q2
+        Q0 = Q3 = Q0 + Q3
 
     return Q0, Q1, Q2, Q3
 
 
-def put_image_quadrants(Q, odd_size=True, symmetry_axis=None):
+def put_image_quadrants(Q, original_image_shape, symmetry_axis=None):
```
DanHickstein: It makes sense that `put_image_quadrants` cannot, from the quadrants alone, figure out the size of the image. If it gets (3,3)-sized quadrants, it doesn't know if they came from a (5,5), (5,6), (6,5) or (6,6) image. I think that users will find it annoying that they need to specify the original image size, but there isn't an easy way around this. `abel.transform()` should wrap up the details nicely anyway, so perhaps it's not a big deal.

abel/tools/symmetry.py
```diff
@@ -54,13 +50,13 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
...
          |   *    |    *   |
-         | Q2  *  | *   Q3 |          cQi == averaged combined quadrants
+         | Q2  *  | *   Q3 |          cQi == combined quadrants
          +--------+--------+                 
 
         symmetry_axis = None - individual quadrants
-        symmetry_axis = 0 (vertical) - average Q0+Q1, and Q2+Q3
-        symmetry_axis = 1 (horizontal) - average Q1+Q2, and Q0+Q3
-        symmetry_axis = (0, 1) (both) - combine and average all 4 quadrants
+        symmetry_axis = 0 (vertical) - sum Q0+Q1, and Q2+Q3
```
DanHickstein: The adding seems like it might result in some weird situations. For example, if somebody uses `symmetry_axis=(0)`, and then uses `quadrants=(True, True, False, True)`. In this situation, the top quadrants would be twice as intense as the bottom.

doc/transform_methods/hansenlaw.rst
```diff
@@ -1,3 +1,6 @@
+.. |nbsp| unicode:: 0xA0
```
DanHickstein: What is this?

doc/transform_methods/hansenlaw.rst
```diff
@@ -9,52 +12,126 @@ The Hansen and Law transform is the work of E. W. Hansen and P.-L. Law [1].
...
+
+.. math:: g(R) = 2 \int_R^\infty \frac{f(r) r}{\sqrt{r^2 - R^2}} dr 
+
+inverse Abel transform 
+
+.. math:: f(r) = -\frac{1}{\pi}  \int_r^\infty \frac{g^\prime(R)}{\sqrt{R^2 - r^2}} dR
+
+
+
+The Hansen and Law method makes use of a coordinate transformation, which is
```
DanHickstein: Awesome! Thanks for including this extra information!

doc/transform_methods/hansenlaw.rst
```diff
@@ -9,52 +12,126 @@ The Hansen and Law transform is the work of E. W. Hansen and P.-L. Law [1].
...
+|
+|
+
+:math:`\Phi_n` and :math:`\Gamma_n` are functions with predetermined 
+parameter constants, all listed in [1].
+
+
+The algorithm iterates along each individual row of the image, starting at 
+the out edge, ending at the center-line. This provides for efficient code 
+parallelization.
```
DanHickstein: Not sure what is meant by parallelization here.

DanHickstein:
Thanks for making these changes @stggh!

That is an interesting issue with `put_image_quadrants()` and I see no easy solution. Passing the original image size as you have done, works, and is robust, but I think it might be annoying to users. Would it be a good idea for `put_image_quadrants()` to have a default of `original_image_shape='odd'`, whereby both the rows and the columns would be assumed to have odd values? (Not that the image is square, just that both rows and column are odd.) Or, your current implementation is fine. 

I think that the averaging the quadrants is the way to go, as it avoids weird situations where the adding quadrants would lead to situations where half of the image is twice as intense as the other half. The existing code seems like it would throws a divide-by-zero errors if the user attempted to map a single quadrant to the whole image (i.e., `symmetry_axis=(0,1)`). We could likely fix this by first including:
```python
is symmetry_axis==(0, 1):
	Q = (Q0*use_quadrants[0]+
         Q2*use_quadrants[2]+
		 Q2*use_quadrants[2]+
		 Q3*use_quadrants[3])/(np.sum(use_quadrants))
	
	return Q, Q, Q, Q 
```

Whatever solution we adopt, we should probably also check for situations where the user selects options for `symmetry_axis` and `use_quadrants` that don't make sense. For example, anything that does not produce a full image should be caught an an error raised:
```
symmetry_axis == None, any quadrant == False    -> at least one quadrant is empty
symmetry_axis == 0, top quadrants both False    -> top of image is empty
symmetry_axis == 0, bottom quadrants both False -> bottom of image is empty
symmetry_axis == 1, left quadrants both False   -> left of image is empty
symmetry_axis == 1, right quadrants both False  -> right of image is empty
```

With these changes, I don't think that averaging should pose a problem, and would produce something closer to what the user expects.

And I agree about the basis sets. Paring down to just one (or two) basis sets should make the examples a lot faster to run.

doc/transform_methods/hansenlaw.rst
stggh: blank space, some of the `:math:` is a bit close to adjacent text

doc/transform_methods/hansenlaw.rst
stggh: Probably, incorrect terminology on my part. I mean the use of other cpus(?)

stggh:
@DanHickstein, sorry for the delay I was trapped in a gridlocked road, due to a car accident further along.

I gave up with averaging because it causes issues about which quadrants should be averaged first. eg. with the old pre-PR code try selecting only Q0, ..., Q3 and see the issue. No doubt @DanHickstein you will come up with a clever solution. 

I will revisit. Just heading off to the river for a few hours to escape from the 33C autumn heat.

doc/transform_methods/hansenlaw.rst
DanHickstein: In this case we didn't do anything directly to make it parallel, what we did was vectorize. We are still iterating over the rows in a serial fashion). But, since we can do operations a full column at a time (working with a vector), then we can take advantage of the fact that numpy will pass the vector to some c-library that will perform the operation really fast. I don't know if the actual process in numpy is parallel or serial, but it's really fast. 

Anyway, perhaps rephrase to:
```The algorithm iterates along each individual row of the image, starting at 
 the out edge, ending at the center-line. Since all rows in a column can be processed 
  simultaneously, the operation can be easily vectorized and is therefor numerically 
  efficient.
```

The idea that all rows in a column should be independent should actually be true of all abel transform methods. Still not really sure why Hansen-Law is so fast. I guess it's the fact that it magically manages to avoid looping over each column to calculate the value for every column. I'm still not entirely sure how it does that. It probably makes an approximation in a similar way to the three-point transform. But whatever approximation it makes seems to be a very well chosen one, since it agrees perfectly in a wide variety of situations.

DanHickstein:
> it causes issues about which quadrants should be averaged first

Yes, you are correct that this is a glaring flaw with the current code.

But, the order of averaging should only be an issue with `symmetry_axis=(0,1)`, in the other cases there is only one (or zero) averaging steps for each pair of quadrants. 

If we simply check for `symmetry_axis=(0,1)` first, then we should avoid this problem.

DanHickstein:
Okay, I think that the following code (included before the `if 0 in symmetry_axis` line in `get_image_quadrants`) should allow for the averaging to work: 

```python
if ((symmetry_axis == [None] and (use_quadrants[0]==False
                                   or use_quadrants[1]==False 
                                   or use_quadrants[2]==False
                                   or use_quadrants[3]==False)) or        # at least one empty
        (symmetry_axis == [0] and use_quadrants[0]==False and use_quadrants[1]==False) or # top empty
        (symmetry_axis == [0] and use_quadrants[2]==False and use_quadrants[3]==False) or # bot empty
        (symmetry_axis == [1] and use_quadrants[1]==False and use_quadrants[2]==False) or # left empty
        (symmetry_axis == [1] and use_quadrants[0]==False and use_quadrants[3]==False)): # right empty
        raise ValueError('At least one quadrant would be empty. Please check symmetry_axis '
                          'and use_quadrant values to ensure that all quadrants will have a '
                          'defined value.')
    
    if symmetry_axis==(0, 1):
    	Q = (Q0*use_quadrants[0]+
             Q1*use_quadrants[2]+
    		 Q2*use_quadrants[2]+
    		 Q3*use_quadrants[3])/(np.sum(use_quadrants))
	
    	return Q, Q, Q, Q
```

And here is code to test all situations, it should probably be incorporated into a unit test:
```python
import abel
import numpy as np

z = np.zeros((5, 5))

for symmetry_axis in (None, 0,1,(0,1)):
    for Q0 in (True, False):
        for Q1 in (True, False):
            for Q2 in (True, False):
                for Q3 in (True, False):
            
                    print symmetry_axis, Q0, Q1, Q2, Q3
                    
                    try:
                        abel.tools.symmetry.get_image_quadrants(z,  
                                    symmetry_axis=symmetry_axis, 
                                    use_quadrants=(Q0,Q1,Q2,Q3))
                    except ValueError as e:
                        print e
```

stggh:
@DanHickstein you have a very logical brain. The top section of your code is quite a mouthful. I have implemented it + one addtional test case `symmetry_axis=(0,1) and not np.all((use_quadrants))`.

I have returned (what I could find) the doc references to average.

Your test code, below.  Not yet implemented as a unit test.
```
[Steve:~]$ python ck.py 
None True True True True
None True True True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None True True False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None True True False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None True False True True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None True False True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None True False False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None True False False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None False True True True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None False True True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None False True False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None False True False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None False False True True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None False False True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None False False False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
None False False False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
0 True True True True
0 True True True False
0 True True False True
0 True True False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
0 True False True True
0 True False True False
0 True False False True
0 True False False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
0 False True True True
0 False True True False
0 False True False True
0 False True False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
0 False False True True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
0 False False True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
0 False False False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
0 False False False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
1 True True True True
1 True True True False
1 True True False True
1 True True False False
1 True False True True
1 True False True False
1 True False False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
1 True False False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
1 False True True True
1 False True True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
1 False True False True
1 False True False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
1 False False True True
1 False False True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
1 False False False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
1 False False False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) True True True True
(0, 1) True True True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) True True False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) True True False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) True False True True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) True False True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) True False False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) True False False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) False True True True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) False True True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) False True False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) False True False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) False False True True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) False False True False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) False False False True
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
(0, 1) False False False False
At least one quadrant would be empty. Please check symmetry_axis and use_quadrant values to ensure that all quadrants will have a defined value.
```

doc/transform_methods/hansenlaw.rst
stggh: Good text, tks. 

 Philosophically, I find this aspect of Abel transforms amazing:
>all rows in a column should be independent should actually be true of all abel transform methods.

Each row corresponds to a definite theta, that the reconstruction does not care about the other rows in an image was hard to get my head around.

DanHickstein:
> Each row corresponds to a definite theta, that the reconstruction does not care about the other rows in an image was hard to get my head around.

Yeah, with BASEX three-point, this important consequence of assuming only cylindrical symmetry (independent rows) gets hidden in the matrix multiplications, while the other algorithms (HL and Direct) make it clear that each row is independent.

I think that the reason that this seemed unintuitive to you (and me) is that photolelectron/photoion images have a higher degree of symmetry than just cylindrical (not quite sure what word to use to describe it. Basically they have intensity that is "slowly varying" as a function of theta). To get good transforms with low signal-to-noise photoelectron data, it's probably a good idea to use a method that makes more assumptions than just cylindrical symmetry, like pBASEX or POP (https://github.com/PyAbel/PyAbel/issues/30).
_____

 As far as the crazy `if` statement in the above. It could be split into several if statements each with their own `ValueError`. This would take up more lines of code, but would likely be more informative for the user.

DanHickstein:
And good point about `symmetry_axis=(0,1) and not np.any((use_quadrants))`. 

Actually, shouldn't we just generally raise a ValueError if ` not np.any((use_quadrants))`?

DhrubajyotiDas:
@DanHickstein Just to add a clarification to the independent-row discussion, with three-point, the independence is pretty explicit, as seen in [this line](https://github.com/PyAbel/PyAbel/blob/master/abel/three_point.py#L58). We are, quite literally, cycling through every row in `IM` one-at-a-time and doing stuff to it, and rebuilding the transformed `inv_IM` row-by-row.

stggh:
@DhrubajyotiDas neat, and nice colours.

`get/put_image_quadrants()` is almost doing the correct thing. If I select one quadrant, there is a blank cross in the image center:
1111none



```python
# -*- coding: utf-8 -*-
import numpy as np

import abel

def gauss(r, r0, sigma):
    return np.exp(-(r-r0)**2/sigma**2)

def test_quad(image_shape=(201, 201)):
    rows, cols = image_shape
    r2 = rows//2
    c2 = cols//2
    x = np.linspace(-c2, c2, cols)*201/c2
    y = np.linspace(-r2, r2, rows)*201/c2

    X, Y = np.meshgrid(x, y)
    R, THETA = abel.tools.polar.cart2polar(X, Y)

    IM = np.zeros(image_shape)

    R0, dummy = abel.tools.polar.cart2polar(X[:r2, -c2:], Y[:r2, -c2:])
    IM[:r2, -c2:] = gauss(R0, 50, 4)
    R1, dummy = abel.tools.polar.cart2polar(X[:r2, :c2], Y[:r2, :c2])
    IM[:r2, :c2] =  gauss(R1, 100, 4)
    R2, dummy = abel.tools.polar.cart2polar(X[-r2:, :c2], Y[-r2:, :c2])
    IM[-r2:, :c2] = gauss(R2, 150, 4)
    R3, dummy = abel.tools.polar.cart2polar(X[-r2:, -c2:], Y[-r2:, -c2:])
    IM[-r2:, -c2:] = gauss(R3, 200, 4)


    print("Enter use_quadrants, symmetry_axis eg (1,0,0,0),(0,1)")
    plt.ion()
    while True:
        uq, sa = input("use_quadrants, symmetry_axis? ")

        try:
            Q = abel.tools.symmetry.get_image_quadrants(IM, symmetry_axis=sa,
                         use_quadrants=uq)
        except ValueError, e:
            print(str(e))
            continue

        rIM = abel.tools.symmetry.put_image_quadrants(Q,
                         original_image_shape=IM.shape, symmetry_axis=sa)

        lbl = "{:s} {:s}".format(str(uq), str(sa))
        flbl = lbl.translate(None, "(),' '")
        plt.subplot(211)
        plt.title("input: "+lbl)
        plt.imshow(IM)
        plt.subplot(212)
        plt.title("reassembled")
        plt.imshow(rIM)
        plt.savefig(flbl+".png",dpi=100)
        plt.draw()

if __name__ == "__main__":
    test_quad()
```

DanHickstein:
@DhrubajyotiDas, that is beautiful! Is that real data?

@stggh, that looks good with the put/get methods. Should this be merged now?

stggh:
Not quite ready, sorry for all the re-edits.  There remains a vertical center blank line, which is no doubt a bug in my test code, above, but it will need to be found.

DhrubajyotiDas:
@DanHickstein Yes, pretty much straight out of the instrument without any processing. 

I was considering putting up the image and its transform as a flame example on this repo, but they are fairly bulky (> 5 MB). I know in the past we've taken steps to keep the repo down to a reasonable size, which is one of the reasons why I haven't done so yet.

@stggh `get/put_image_quadrants()` looks pretty cool. I will definitely find a use for it in some of my other projects.

stggh:
@DhrubajyotiDas  `get/put_quadrants()` is my nemesis. It opens up a can of worms concerning how an image may be divided, further compounded by the odd-width requirements of the Abel transform methods. 

Note, the more clever Python code is due to @rth and @DanHickstein.  I just implemented the concept of quadrant combining (from my own VMI analysis), and it has evolved in PyAbel.

stggh:
@DanHickstein this PR merge has done something weird to the doc links, e.g. README.rst

>  :doc:`Hansen–Law <transform_methods/hansenlaw>`

DanHickstein:
I think that this was my previous PR. I made the links look good when the documentation is built (see http://pyabel.readthedocs.org/en/latest/readme_link.html) but they don't work on GitHub. 

I don't know of an easy fix to make them work in both situations... Maybe I will just shorten the readme and move some of the information on the transform methods to a different doc?

DanHickstein:
A solution to this problem with the links is discussed here: http://comments.gmane.org/gmane.comp.python.sphinx.devel/7539 but it looks a little complicated to implement.

Issue #136, opened 2016-03-04
Version 0.7 last call (Benchmarking fun)


DanHickstein:
I think it's about time that we released version 0.7! I should have time to do this on Sunday. 

Anyone have more changes that they would like to make before then?

@PyAbel/developers

stggh:
1. There are a few holes in the docs for `hansenlaw`, `tools.vmi`, that I will try to plug. doc making works on my (home) Macbook. `livehtml` is neat! Python virtualenv is also a good idea, which I will try.

2. I am not 100% happy with the quadrant combining logic in `tools.symmetry.put_image_quadrants()`
   ```python
    if 0 in symmetry_axis:
        Q0 = Q1
        Q3 = Q2

    if 1 in symmetry_axis:
        Q2 = Q1
        Q3 = Q0
   ```
   e.g. for `symmetry_axis=0` In principle `Q1` might be empty, which would then stomp `Q0` etc

3. Extreme benchmark   
(note my VMI images may be of size 8192x8192 pixels)
 ```
 $lscpu
Architecture:          x86_64
     :                            :
Model name:            Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz

32GB RAM
SSD HD

$python benchmark.py
PyAbel benchmark run on x86_64

 =========     Direct Abel implementations ==========
time to solution [milli-sec (=sec/1000)] -> transform (basis sets generation)

 Implementation |    n = 251          |    n = 501          |    n = 1001         |    n = 2001            |    n = 4001         
 -----------------------------------------------------------------------------------------------------------------------------------
  direct_Python |     68.0            |    516.0            |   5335.8            |  40789.8               | 1137469.9            
      HansenLaw |     17.5            |     42.8            |    131.5            |    455.1               |   1197.6            
       direct_C |      9.8            |     72.4            |    573.8            |   4944.3               |  36771.5            
                                                                                                        
 =========    Inverse Abel implementations ==========                                                   
time to solution [milli-sec (=sec/1000)] -> transform (basis sets generation)                           
                                                                                                        
 Implementation |    n = 251          |    n = 501          |    n = 1001         |    n = 2001            |    n = 4001         
 ---------------------------------------------------------------------------------------------------------------------------
          BASEX |     20.3 (  4451.9) |    256.0 ( 22894.8) |   1967.5 (167616.5) |  13784.0 (1535687.8)   |  78991.8 (13295525.7)    
[...]

PR #139, opened 2016-03-07
switched linalg from numpy to scipy in basex and three point

DanHickstein:
This is following up on the benchmarking discussion in https://github.com/PyAbel/PyAbel/issues/136. Apparently, the functions (like `dot` and `inv`) in scipy.linalg use a fortran version of LAPACK, which is compiled from fortran. Numpy uses some implementation of BLAS/LAPACK, and the implementation depends on the OS and the python distribution. In most cases (Apple’s Accelerate framework on OS-X, Intel’s MKL libraries in Anaconda, and openBLAS), numpy is just as fast as scipy. However, there are cases where numpy uses versions of BLAS and LAPACK that are not optimized and provide much slower behavior. Since scipy is a requirement anyway, then it makes sense that we just use scipy.

More info:
http://www.scipy.org/scipylib/faq.html#why-both-numpy-linalg-and-scipy-linalg-what-s-the-difference
http://docs.scipy.org/doc/scipy/reference/tutorial/linalg.html

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-include README.md
+include README.rst
 recursive-include abel *.py
 recursive-include abel *.gz
 recursive-include examples *.py
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -13,8 +13,9 @@ import sys
 
 import numpy as np
 from scipy.special import gammaln
-from numpy.linalg import inv
+from scipy.linalg import inv
 from scipy.ndimage import median_filter, gaussian_filter, center_of_mass
+import scipy
 
 from ._version import __version__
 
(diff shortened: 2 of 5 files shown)


DanHickstein:
As @stggh mentioned in https://github.com/PyAbel/PyAbel/issues/136#issuecomment-193480492, lots or warnings pop up during `python setup.py install`. These are related to the direct C-implementation. They are not major issues, mainly just unused functions from the libraries that are imported. I suppressed these warnings by adding the following flags to the c-compiler in setup.py:
```python
compile_args = {dict( extra_compile_args=['-Wunused-function', 
                                          '-Wunneeded-internal-declaration',
                                          '-Wno-#warnings'])}
```

The install now proceeds without warnings.

DanHickstein:
@DhrubajyotiDas should look over this before we merge since this PR represents a relatively untrained hand tinkering with the inner workings of BASEX and Three-Point. :)

DhrubajyotiDas:
@DanHickstein Thanks a lot for making these changes. Everything seems to be in order, as long as 
```python
scipy.dot(A,B) == A.dot(B)
```

A minor point - stylistically, I've always seen `scipy` imported as `sc`
```python
import scipy as sc
sc.dot(A,B)
```
but I leave this up to you.

stggh:
The output from `example_all_O2.py` looks good, so I would say, no damage done.

DanHickstein:
Sounds good. I will go ahead and merge.

I am undecided about the extra arguments that I included for the C-compiler in the setup.py file. When I tried to install on our linux cluster, I think that a different C-compiler was used and it did not recognize the flags. I think that this may have prevented the direct C-implementation from installing. Or, it may have been my lack of linux skills :). @stggh, can you check the install on linux?

stggh:
Build is good on ubuntu 15.10

DanHickstein:
Awesome!

Issue #140, opened 2016-03-08
Slight Module Structure Improvement


himbeles:
Currently it is not possible to include the "abel" folder as a standalone module into.
This is because of the single line "import abel" in the files transform.py and "benchmark.py".
Replacing this by 
"from .. import abel" 
cures the problem.

DanHickstein:
Hi @himbeles, thanks for suggesting this and welcome to the PyAbel GitHub! 

We have typically been installing PyAbel using
`python setup.py develop` 
to allow the package to be installed, but remain in a user's directory where the source code can be directly modified. (Alternatively, you can use `python setup.py install`).

Changing to `from .. import abel` might allow for the behavior that you are looking for, but it breaks the package for the installed version. Here is the results after making the change:

```python
In [1]: import abel
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-0953192da019> in <module>()
----> 1 import abel

/Users/danhickstein/Documents/PyAbel-dh/abel/__init__.py in <module>()
      7
      8 from . import basex
----> 9 from . import benchmark
     10 from . import direct
     11 from . import hansenlaw

/Users/danhickstein/Documents/PyAbel-dh/abel/benchmark.py in <module>()
      9
     10
---> 11 from .. import abel
     12
     13

ValueError: Attempted relative import beyond toplevel package
```

I think that we would need to change the imports in `transform.py` and `benchmark.py` to be something along the lines of 
``` python
from . import basex
from . import hansenlaw
from . import direct
from . import three_point
etc, etc
```

These changes could be made, but we would want to be careful to make sure nothing gets messed up with the regular installation.

Does the `python setup.py develop` not work for you?

DanHickstein:
Okay, I think that this works as long as we change
```python
import abel
```
to
```python
from . import basex
from . import hansenlaw
from . import direct
from . import three_point
from . import tools
```

I don't really know enough about Python to know which is the preferred style. The former is fewer lines of code, but the latter seems to ensure that we are using packages from the same PyAbel directory. Also, it lets PyAbel run without being installed, which I suppose is a good thing...

himbeles:
Thanks for the quick reply!
I think, this change is really useful for many applications, where you build an integrated solution and want to include a fixed version of pyabel for compatibility reasons (and you dont want to resort to python virtual environments).

stggh:
@DanHickstein 

[...]

PR #142, opened 2016-03-08
Removed "import abel"

DanHickstein:
As suggested in https://github.com/PyAbel/PyAbel/issues/140, I removed the `import abel` from `benchmark.py` and `transform.py` and replaced with
```
from . import basex
from . import hansenlaw
from . import direct
from . import three_point
from . import tools
```

This allows the PyAbel package to be run from its directory without installing the package.

I’m not completely sure if we want to do this, but I’ve made the changes in order to spark the discussion. 

@PyAbel/developers, @himbeles


ldes89150:
I think it's a right decision.

stggh:
Does not appear to work (not installed):
```python
[etalon:.+n/VMI/PyAbel-DH]$ git clone https://github.com/DanHickstein/PyAbel.git -b imports
Cloning into 'PyAbel'...
remote: Counting objects: 3234, done.
remote: Compressing objects: 100% (174/174), done.
remote: Total 3234 (delta 103), reused 0 (delta 0), pack-reused 3059
Receiving objects: 100% (3234/3234), 19.88 MiB | 3.62 MiB/s, done.
Resolving deltas: 100% (2247/2247), done.
Checking connectivity... done.
[etalon:.+n/VMI/PyAbel-DH]$ cd PyAbel
[etalon:.+yAbel-DH/PyAbel]$ cd abel
[etalon:.+-DH/PyAbel/abel]$ python benchmark.py 
  File "benchmark.py", line 9, in <module>
    from . import basex
ValueError: Attempted relative import in non-package

``

**Comment by DanHickstein** [comment]
I think that it works installed (it works on Travis and Appveyor). And try `cd ..` (Back up to /PyAbel-DH) and `import abel`.

**Comment by DanHickstein** [comment]
Yeah, this does take away our ability to `python bechmark.py` and include stuff in `main()`. But this is probably sloppy anyways.

**Comment by stggh** [comment]
Ok, appears to work.   

Perhaps, add running instructions at the head of  `benchmark.py` (I think that  this file should live in the `tools` directory in any case (#113)), or, probably not a good option, add another file in the `abel` directory, `run_benchmark.py` with 
```python
import abel
abel.benchmark.AbelTiming()
```

DanHickstein:
Are you a fan of the tools directory? To me, I kinda prefer a flat organization so that I don't need to type `.tools.` when I'm calling functions...

stggh:
Only in the sense that it keeps the core transform programs neatly separate from the auxiliary code.

DanHickstein:
Yeah, I like the way that the folders separate the source code, but I don't like typing more lines. 

Anyway, I think that we can merge this.

PR #145, opened 2016-03-10
center_image via abel.transform()

stggh:
1. `abel.transform.py`: Calling `center_image()` via `abel.transform(..., center="method"...)` requires path of the function - added.
2. `abel.tools.symmetry` as per #137 only need to check `not np.any(use_quadrants)`.
3. Trival doc changes


abel/tools/symmetry.py
```diff
@@ -109,7 +109,7 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
         (symmetry_axis == [1] and use_quadrants[0]==False and 
                                   use_quadrants[3]==False)    # right empty
                                                                 or
-        (symmetry_axis == (0, 1) and not np.any(use_quadrants))
+        not np.any(use_quadrants)
```
DanHickstein: Yes, good point!! :)

abel/transform.py
```diff
@@ -253,7 +253,7 @@ def transform(
             raise ValueError('Image must have an odd number of columns. \
                               Use a centering method.')
     else:
-        IM = center_image(IM, center, **center_options)
+        IM = tools.center.center_image(IM, center, **center_options)
```
DanHickstein: Oh, good catch on this one too!

DanHickstein:
Thanks @stggh! Two huge bugs squashed.

PR #146, opened 2016-03-10
Implement angular_integration  into abel.transform()

stggh:
Implemented angular integration (speed distribution) into `abel.transform()`, and updated docs.  
`abe.transform()` now also returns `['angular_integration']` data.

Q: @DanHickstein should I update the `__main__` example at end of the `transform.py` file to use the this call format? 

additional parameters:  `angular_integration=True/False`
                                     `angular_integration_options=dict()`

e.g.
```python
import numpy as np
import abel
IM = np.loadtxt("examples/data/O2-ANU1024.txt.bz2")
res = abel.transform(IM, center="slice", method="hansenlaw", direction="inverse", angular_integration=True)
plt.plot(*res['angular_integration'])
plt.show()
```

### DIFF
--- a/abel/tests/test_direct.py
+++ b/abel/tests/test_direct.py
@@ -6,7 +6,6 @@ from unittest.case import SkipTest
 from numpy.testing import assert_allclose
 import scipy.ndimage as nd
 import abel
-from abel.tools.polar import CythonExtensionsNotBuilt
 
 def test_direct_shape():
     if not abel.direct.cython_ext:
--- a/abel/tests/test_transform.py
+++ b/abel/tests/test_transform.py
@@ -30,11 +30,11 @@ def test_transform_shape():
         pass
 
     iZ = abel.transform(Zodd) 
-    # check odd width
+    # check odd width returned with default transform
     assert iZ['transform'].shape[1] % 2 == 1
 
     iZ = abel.transform(Zeven, center="com")
-    # check odd width
+    # check odd width returned for even image, centered using 'com'
     assert iZ['transform'].shape[1] % 2 == 1
 
 
--- a/abel/tests/test_transform.py
+++ b/abel/tests/test_transform.py
@@ -65,7 +65,7 @@ def test_transform_angular_integration(n=101):
     assert 'angular_integration' in results.keys()
 
     radial = results['angular_integration']
-#    assert np.shape(radial) == (2, 72)   # not sure where 72 comes from
+    assert np.shape(radial) == (2, int(np.sqrt(c2**2+r2**2))) 
 
     max_position = radial[1].argmax()
     assert np.allclose(max_position, c2, rtol=0, atol=2) 
--- a/abel/tests/test_transform.py
+++ b/abel/tests/test_transform.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+from numpy.testing import assert_allclose
+import abel
+
+# The role of abel.transform() is to:
+# 0. (optional) center image and make odd-width
+# 1. split an image into quadrants,
+# 2. transform each quadrant
+# 3. reassemble image
+# 4. (optional) evaluate speed profile
+# The specific functions will have been tested separately
+# here we check that the functions combine correctly
+
+def test_transform_shape():
+    Zeven = np.ones((4,4))
+    Zodd = np.ones((4,5))
+
+    try:
+        # even width image should be rejected
+        iZ = abel.transform(Zeven) 
+        # fail if get here
+        assert iZ['transform'].shape[1] % 2 == 1
+    except:
+        pass
+
+    iZ = abel.transform(Zodd) 
+    # check odd width
+    assert iZ['transform'].shape[1] % 2 == 1
+
+    iZ = abel.transform(Zeven, center="com")
+    # check odd width
+    assert iZ['transform'].shape[1] % 2 == 1
+
+
+def test_transform_angular_integration(n=101):
+    gauss = lambda r, r0, sigma: np.exp(-(r-r0)**2/sigma**2)
+
+    image_shape=(n, n)
+    rows, cols = image_shape
+    r2 = rows//2 + rows % 2
+    c2 = cols//2 + cols % 2
+    x = np.linspace(-c2, c2, cols)
+    y = np.linspace(-r2, r2, rows)
+
+    X, Y = np.meshgrid(x, y)
+    R, THETA = abel.tools.polar.cart2polar(X, Y)
+
+    IM = gauss(R, c2, 2) # Gaussian donut located at R=c2
+
+    # forward Abel transform
+    fIM = abel.transform(IM, method="hansenlaw",
+		    direction="forward")['transform']
+
+    # inverse Abel transform, and radial intensity distribution
+    results = abel.transform(fIM, method="hansenlaw", direction="inverse",
+                             angular_integration=True)
+
+    assert 'transform' in results.keys()
+    assert 'angular_integration' in results.keys()
+
+    radial = results['angular_integration']
+#    assert np.shape(radial) == (2, 72)   # not sure where 72 comes from
+
+    max_position = radial[1].argmax()
+    assert np.allclose(max_position, c2, rtol=0, atol=2) 
+        
+
+if __name__ == "__main__":
+    test_transform_shape()
+    test_transform_angular_integration()
--- a/abel/tools/polar.py
+++ b/abel/tools/polar.py
@@ -157,17 +157,3 @@ def polar2cart(r, theta):
     y = r * np.sin(theta)   # θ referenced to vertical
     x = r * np.cos(theta)
     return x, y
-
-
-class CythonExtensionsNotBuilt(Exception):
-    pass
-
-# why is this here??
-CythonExtensionsNotBuilt_msg = CythonExtensionsNotBuilt(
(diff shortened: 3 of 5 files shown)


DanHickstein:
Thanks for doing this @Stggh!

I think that you can just delete the `__man__()` function in `transform.py`. The file cannot be run anymore now that we switched to the `from . import basex` nomenclature (I think that this was the only disadvantage of that change). 

I just wrote the `main()` function to quickly test `abel.transform` while I was coding it up, so it's not really needed anymore. We can just use the other example scripts to test the behavior of `abel.transform`.

stggh:
call example passing transform parameter `dr=0.5` (which also gets sent to `angular_integration`):

```python 
 np.shape(res['angular_integration'])
# (2, 724)

resdr = abel.transform(IM, center="slice", method="hansenlaw", direction="inverse", angular_integration=True, transform_options=dict(dr=0.5))
np.shape(resdr['angular_integration'])
#(2, 1448)
```

Note, for small image sizes `dr=0.5` may improved the look of the speed distribution pixel granulation. However, tests have show there is nothing to be gained by using an even smaller grid. Use a larger pixel image size instead!

stggh:
1. I think VMI the `dr=1` vs `dr=0.5` issue was due to execution speed.
2. doc good idea.
3. `dr`: it is good to experience another users perspective. `dr` does map the pixel grid to the radial coordinates. I see now that for different functions, the values may differ. However, if the user forgets to speciy `dr` in both `transform_options` and `angular_integration_options` then the result may be wrong (associated radial grid). How about, I copy the `dr` from `transform_options` only if there is no `dr` in `angular_integration_options`?
4. I note that there are no unit tests for transform.py. I will look at writing something, if I find some free time today.

DanHickstein:
1. Yeah, so maybe leave at `1` to keep things running efficiently?

2. awesome

3. I don't think that there should be any copying of `dr` values since they are completely unrelated. Most experimentalists will not care about the `dr` being passed to the transform function, since it simply represents an arbitrary multiplication of their data. The most important number for them is the `dr` that is passed to the `angular_integration` function. We could just as easily have called one `scaling_factor` and the other `radial_grid_spacing`. Perhaps if you added something to the `abel.transform` docstring suggesting that users looking for higher resolution should set `dr=0.5` that would take care of things. 

4. Unit tests for `transform.py` would be most welcome!

stggh:
I think that you are correct that a new variable may be in order. From my perspective, the speed (or angular_integrations) for `dr=0.5, 1` should overlap in the speed plots. Unless `dr` is passed to both `_options` dicts, they will not. I will think about this a little more.

DanHickstein:
The question of whether or not the speeds calculated from `dr=1.0

stggh:
I am nearly convinced, however, we are both imagining different axes.  I expect the speed distributions to overlap in the radial sense. I am not concerned (too much, but should be) about the intensity. For `dr=0.5` there are twice as many data points, but the detector radius is still the same, so, `angular_integration()` needs the correct value for `dr`.

stggh:
I think we may be on the same page. The only change required is in `abel.tools.vmi.angular_integration()` and include `dr` in the sum, to correctly scale the speed intensity:
```python
    speeds = np.sum(polarIM, axis=1)*dr
```
```
  :
In [4]: IM = np.loadtxt("examples/data/O2-ANU1024.txt.bz2")
In [5]: res = abel.transform(IM, center="slice", method="hansenlaw", direction="inverse", angular_integration=True, transform_options=dict(dr=0.5), angular_integration_options=dict(dr=0.5))
In [6]: res1 = abel.transform(IM, center="slice", method="hansenlaw", direction="inverse", angular_integration=True)
  :
```

speed-ck

As an after thought, perhaps, the angular sum should be an integration, which would then correctly treat the intensity variation due to `dr` and `dt` ?

DanHickstein:
Yes, I think that you are correct about making it an integration as far as `dt` is concerned. As far as `dr` is concerned, I think that the most intuitive behavior would be for the intensity to change with the bin size in order to keep a constant total number if the user were to integrate all of the r-bins. However, I think that what you are trying to achieve is also valid. You are keeping it in terms of "counts per pixel in the original image". Just make it clear in the angular_integration docstring that this is what the user will be getting and if they want to keep their units straight, then they had better take into account deviations of dr from 1.0. I suppose the fact that they actually have the array of r-values makes everything much less confusing anyway...

DanHickstein:
The results with `dr=1.0` and `dr=0.5` look really similar to me, even with your super-sharp peaks. I would suggest that we stick with a default of `dr=1.0` for performance reasons (of course, add something in the docstring of abel.transform suggesting that users can boost the performance to 0.5 if they need it).

stggh:
1. What we are both talking about is the Jacobian, for non-unity `dr` (and in the end `dt`). Provided, Jacobian scaling is taken care of, which it now is (for `dr`), the "intensity to change with the bin size in order to keep a constant total number" is achieved. 

2. "look really similar to me,"
v23
Agreed, but note the maximum intensities for v'=3 and v'=2, swap ;-)

DanHickstein:
Yeah, I suppose if the user has access to `r` as well as the intensity, then they can take care of the Jacobian themselves. Can you just add a note in the docstring to the effect that the intensity is a "per pixel" value, and while the intensity will remain (nearly) the same as `dr` is changed, the sum of the intensity array will change unless the spacing of the `r` array is taken into account.

stggh:
@DanHickstein we appear to be actioning at the same time (but 18 hours out of phase)

1.  >they can take care of the Jacobian themselves

  I think that we are taking care of the Jacobian (with the `*dr` factor), and more fully/explicitly with `Jacobian=True`.  There are no surprises AFAICT. I may be missing some apsect(?).

2. The unit test `test_transform_angular_integration()` brought back some past hassles. You may be able to refresh my memory. In `test_transform.py` for an image size `(101, 101)` `['angular_integration']` returns size `(2, 72)`. The `72` is the diagonal length of a `51*51` square. For the radial intensity it should be no bigger than `51`. 

  If I perform another test, everything is fine. 
  ```
  In [1]: import numpy as np
  In [4]: import abel
  In [5]: IM = abel.tools.analytical.sample_image()
  In [6]: speed = abel.tools.vmi.angular_integration(IM)
  In [8]: speed[1].shape
  Out[8]: (256,)
  # alternative method
  In [9]: res = abel.transform(IM, method='hansenlaw', angular_integration=True)
  In [10]: res['angular_integration'][1].shape
  Out[10]: (256,)
  ```
  Something about the unit test triggers the diagonal size.

stggh:
Also, `class CythonExtensionsNotBuilt`
> why is this here??

Was left over after from the original `tools.py` file, after the functions were distributed into 'themed' files.  I am just not sure where best to put it.

DanHickstein:
@stggh, Good discussion! I think that I was home sick on the day that we learned about the Jacobian in high-school calculus class. :)

1. I wanted users to be able to sum their abel-transformed image and get a number, say 1,234 counts. Then, when they sum their "speeds" array, I wanted them to get the same number of counts. This would occur if we took into account the r^2 sin(theta) Jacobian during the angular integration and then returned an array of I(r) values that had a bin size of 1 pixel. If the bin size were changed to 0.5 pixels, then they would have twice as many bins and the counts in each bin should change accordingly.

I think that you are envisioning a slightly different scenario, where the user has both the `r` and `I(r)` arrays and can *integrate* `I(r)` instead of summing (i.e., they take into account the spacing of the `r` array). I think that both views are equally valid, and your approach has the added benefit that the `I(r)` curves have the same intensity for different `dr` settings. So, go with that. But do please add a note in the angular_integration docstring describing the behavior. 

2. At one point we decided that there should be a keyword argument to `reproject_image_into_polar` that would set the region of the image returned (basically, we can envision only looking at the "valid circle" or the circle encompassing the entire image). I think that it proved difficult to implement this in the `reproject...` function, but perhaps we should implement a similar keyword in the `angular_integration` function? Something like `crop='inner'/'outer'`? It would then either leave the array uncropped (outer) or crop it to `np.min(width/2, height/2)`.

3. I don't think that `class CythonExtensionsNotBuilt` is used anymore. If you want to want to check if the Direct_C is installed, you can use `import direct` and check if `direct.cython_ext` is `True` or `False`. 

So, you can remove the following lines from `polar.py`: 
```python
class CythonExtensionsNotBuilt(Exception):
    pass

# why is this here??
CythonExtensionsNotBuilt_msg = CythonExtensionsNotBuilt(
    "Cython extensions were not propery built.\n"
    "Either the complilation failed at the setup phase"
    " (no complier, compiller not found etc),\n"
    "or you are using Windows 64bit with Anaconda that has a known issue"
    " with Cython\n"
    "https://groups.google.com/a/continuum.io/forum/#!topic/anaconda/\
    3ES7VyW4t3I\n")
```

Also, please remove the only other mention of `CythonExtensionsNotBuilt`, which occurs on line 9 of `test_direct.py`:
``` python
from abel.tools.polar import CythonExtensionsNotBuilt
```

stggh:
@DanHickstein, I will be away from Canberra for the next 3 days. 

If if is ok with you I would like to (or you may) merge this PR. It offers the ability to calculate the speed distribution using `abel.transform()`, and fixes intensity scaling due to `dr`.

I have removed the legacy references to `CythonExtensionsNotBuilt`.

The integration, and crop features can be implemented after I return.

DanHickstein:
Sounds good! 

Just so that we remember, I think that the only things left to do are 
1) add a note to the docstring
2) think about adding `crop` kwarg to the angular_integration function.

stggh:
Ok, I misread the integration discussion. Tks.

PR #150, opened 2016-03-13
add the option to use fourier transform to symmetrize the data.

ldes89150:


DanHickstein:
Looks really useful! Thank you for adding this.

I'll have a more careful look at the code later on, probably tomorrow, and then we can merge.

abel/tools/symmetry.py
```diff
@@ -31,6 +35,16 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
        Include quadrant (Q0, Q1, Q2, Q3) in the symmetry combination(s)
        and final image
 
+    symmetrize: str
+       Method used for symmetrizing the image.      
+ 
+       average: Simply average the quadrants.
+       fourier: Axial symmetry implies that the Fourier components of the 2-D projection 
+                should be real and even. Removing the imaginary components in reciprocal
```
DanHickstein: Should this be reworded? I think that axial symmetry implies that we have an "even" function, which implies that the Fourier components are real, but makes no even/odd statement about the Fourier components. Probably you can just remove the word "even" here. Maybe I misunderstand...

abel/tools/symmetry.py
```diff
@@ -31,6 +35,16 @@ def get_image_quadrants(IM, reorient=True, symmetry_axis=None,
...
 
+    symmetrize: str
+       Method used for symmetrizing the image.      
+ 
+       average: Simply average the quadrants.
+       fourier: Axial symmetry implies that the Fourier components of the 2-D projection 
+                should be real and even. Removing the imaginary components in reciprocal 
+                space leaves a symmetric projection.
+                ref: Overstreet, K., et al. "Multiple scattering and the density 
+                     distribution of a Cs MOT." Optics express 13.24 (2005): 9672-9682.
```
DanHickstein: can you add the DOI link: http://dx.doi.org/10.1364/OPEX.13.009672

abel/transform.py
```diff
@@ -19,7 +19,7 @@
 def transform(IM,
               direction='inverse', method='three_point', center='none',
               symmetry_axis=None, use_quadrants=(True, True, True, True),
-              angular_integration=False,
+              symmetrize='average', angular_integration=False,
```
DanHickstein: Do we want to call this `symmetrize_method`? Calling it just `symmetrize` seems to imply that it's a boolean.

DanHickstein:
I looked carefully through the code, and this seems great to me! (See minor line-notes.)

`get_image_quadrants()` is @stggh's specialty, so I will give him a day to return to the office before merging. 

Thanks again @ldes89150!

ldes89150:
@DanHickstein Thanks for the suggestions:D

stggh:
This is a neat concept. 

It will be interesting to see the effect of such enforced symmetry on an off-center image.

A few minor points:
 1. `abel.transform()` currently has no way to transfer `symmetrize_method="fourier"` to `get_image_quadrants()`. Add this as a parameter(?). 
 
  @PyAbel/developers we should look at whether the symmetry parameters `symmetry_axis`, `symmetrize_method`, `use_quadrants()` should be packed together and passed to `transform()` using `symmetrize_options=dict()`. In fact each dict could be inspected for keys, eliminating the need for `center="method"`, `angular_integration=True`, and to the extreme, moving `method="<method>"`, `direction` into `transform_options=dict()`?

 2. `fft2, ifft2` are imported, but not used. 

     BTW I like @DhrubajyotiDas comment #139 about `import scipy as sc`.

 3. minor typo in the warning message `4 qudrants`.

 4. pep8 will identify a few long lines (some legacy).

Good work. Thanks, for making this technique available.

DanHickstein:
Thanks for taking a look at this @stggh!

I tried a few sample images, and the results of `average` and `fourier` appear to be identical. Apparently taking the real part of the FFT is almost exactly the same as averaging both sides. Fourier transforms continue to amaze me. 

image

1. @ldes89150 did add `symmetrize_method="average"` as a kwarg to abel.transform, which I think is the right decision. In general, I don't see a problem with a mixed approach to the kwargs, that is, having the most important kwargs be direct keyword arguments to abel.transform and having some (less frequently used) arguments passed in the dictionaries. In general, I think that we should try to avoid having the majority of users need to use the dictionary approach since it's a bit more cumbersome than a direct argument to the function. 

    For now, I would suggest that we don't need a `symmetry_options` dict. If the symmetrize function gets a lot more options, then maybe we should do this. 

2. Yes, `fft2` should be removed. 

   Regarding how to import `scipy`: looking at the [scipy docs](http://docs.scipy.org/doc/scipy/reference/api.html#guidelines-for-importing-functions-from-scipy), it seems like we should be using the format:
```python
   from scipy import fftpack
```
    and then calling the functions with:
```python
   fftpack.fft()
```

stggh:
Opps! Sorry, @ldes89150, I totally missed the `transform.py` changes that you made. I looked at your code, after returning from a road trip, and in the chaos must have inspected one of my old files.

ldes89150:
Thanks for catching my typo.

DanHickstein:
This looks good to me. I think that we can merge!

PR #155, opened 2016-03-20
Incorporate onion peeling method into PyAbel distribution

stggh:
As per #56 the `onion_peeling` method is just about ("see Issues") ready for prime time.

@DanHickstein coded the algorithm in python, mirroring the matlab code of @ChrisRallis, however, there were differences between reconstructed and original images. We found these are reduced by including a `1/r` factor Jacobian.

This PR fully incorporates the `onion_peeling` method into the PyAbel stable. 
It includes:
- `onion_peeling.py` method file, including the Jacobian correction, removed conditions requiring the image and reconstruction to be positive (see #56), added citations
- updated `transform.py` to include `onion_peeling`
- unit test `test_onion_peeling.py` (currently includes a scaling factor)
- updated docs, `onion_peeling.rst` and any legacy descriptions 
- `example_hansenlaw_onion.py`, updated `example_all_O2.py`
- `abel.benchmark` now with `onion_peeling`:
   ```
In [1]: import abel

In [2]: abel.benchmark.AbelTiming()
Out[2]: 
PyAbel benchmark run on i386


=========     Direct Abel implementations ==========
time to solution (millisec)

Implementation     n = 301              n = 501          
----------------------------------------------------------
  direct_python     142.4                940.4            
      hansenlaw      20.8                 50.1            


=========    Inverse Abel implementations ==========
time to solution (millisec)

Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex       7.7                 56.1            
       basex_bs    6495.9              20926.6            
  direct_python     149.0                901.9            
      hansenlaw      26.7                 63.3            
  onion_peeling     642.0               2114.3            
    three_point      17.9                 46.2            
 three_point_bs    1425.8               4215.1 
```

**Issues**:
1. The 1-d Gaussian transform does not agree as well as the 2-d:
onion_1d_2d
     The code for this test is here [check_onion.py](https://gist.github.com/stggh/a8345737676a24e0a091)

2. NB the reconstructed Gaussian's are scaled, I am not sure what the common normalization factor should be ... 
 For the test case `n=100`
    ```
    $ python check_onion.py 
    1d ratio = 20.7249
    2d ratio = 21.4343 
    ```

3. The transform for O2<sup>-</sup> PES has an extended tail, characteristic of insufficient background subtraction.  According to the published papers (see #56) this method *should* do better.
example_all_o2

4. The onion peeling algorithm is formulated on the image grid, not pixel grid like the other methods. I added code to `onion_peeling.py` to shift the image `1/2` pixel. to achieve this. However, it makes little difference, and does not improve the agreement. In general, `1/2`-pixel shift does not change much for large `n>100` size images.

5. The Jacobian devised by @DanHickstein, has values starting at 2 `np.linspace(2,w,w)`, which is not dividing the intensity by clean `0, 1, 2, ...w-1` values, but something close.  I have a feeling that a minor change to the original code may achieve an equivalent result. I spent a great deal of time investigating this but did not get very far.

6. Docs for this method could be made more consistent, which I will do, but feel free to contribute.

This turned out to be a bigger task than I expected.
Enjoy!

Edit: apologies for the bad item numbering, poor markdown skills on my part (and it is very late)

### CODE REVIEW

File: abel/onion_peeling.py [hunk 1]
```diff
@@ -0,0 +1,151 @@
...
+        the inverse Abel transformed half-image 
+
+    """
+
+
+    # onion-peeling uses grid rather than pixel values, 
+    # odd shaped whole images require shift image (-1/2, -1/2)
+    if shift_grid:
+        IM = shift(IM, -1/2)
+
```
DanHickstein: Throw an error if direction='forward'

abel/onion_peeling.py
```diff
@@ -0,0 +1,151 @@
...
+    IM = np.atleast_2d(IM)
+
+    # The original pythod code operated on the left-half image
+    # Other methods use a right-half oriented image, flip for common use
+    IM = IM[:, ::-1]
```
DanHickstein: You could also use "IM = np.fliplr(IM)". Same behavior, perhaps more legible code.

abel/onion_peeling.py
```diff
@@ -0,0 +1,151 @@
...
+#
+#  http://scitation.aip.org/content/aip/journal/rsi/67/6/10.1063/1.1147044
+#
+# 2016-03-19: Jacobian intensity correction, allow negative image values
+#             Steve Gibson and Dan Hickstein
+# 2015-12-16: Python version of the Matlab code, by Dan Hickstein see issue #56
+#
+################################################################################
+
+def init_abel(xc, yc):
```
DanHickstein: should this be a private member to keep it out of the docs? Just put a underscore before the name.

DanHickstein:
@stggh, Great work! So many things are involved in getting one of these methods working. 

Fundamentally, it would be really nice if one of us (or, @ChrisRallis) could go through `onion.py` and add some description of what the algorithm is actually doing, and hopefully change the variable names to more descriptive things. 

I am basically just guessing my way through the code. But, one thing that I did figure out, is that we can eliminate the need for the `1/r` correction by removing the source. It's in the init_abel() function, line 42. We just need to change this:
```python
val2[idist, jdist] = np.sqrt((idist+1)**2+(jdist+1)**2)/(idist+1)
```
to this:
```python
val2[idist, jdist] = 1.0/(idist+1)
```

Then, we can get rid of all of the code doing the `1/r` correction. 

In your test code, you were using a spherical gaussian function, but I think that it's easier to compare 1D and 2D with a cylindrical gaussian function. So, change
```python
IM = gauss(R, 0, sigma) # Gaussian max located at pixel=0
```
to
```python
IM = gauss(X, 0, sigma)
```

After this, I get the following ratios: 
```
1d ratio = 21.1569
2d ratio = 21.1572
```
So, they are nearly the same. Really, what we want to be doing is to compare to to the analytical result for the abel transform of a gaussian (these are in abel.tools.analytical). The analytical result for a gaussian is sigma*sqrt(pi)*gaussian. sigma*sqrt(pi) is 35.45 in this case. So, it's still off by a little bit, but close. We should probably track this down, or see if we can add a scaling factor.

stggh:
@DanHickstein excellent suggestions, all implemented. I will also update the unit test (soon).

DanHickstein:
The onion peeling transform still does not seem to be in great agreement with the gaussian. Maybe there is still some issue with the center being mid-pixel or not...

stggh:
I did not have much free time, yesterday, to look at the `onion_peeling()` offsets. I did look at the `_init_abel()` vectorization. However, the code, below, is not working correctly. Just thought I would post here in case the vector expert @DanHickstein is already examining the same issue.
```python
def _init_abel(xc, yc):

    i = np.arange(xc+1, dtype=int)
    j = np.arange(yc+1, dtype=int)
    ones = np.ones((xc+1, yc+1))

    J1, I = np.meshgrid(j+1, i)
    J1, I1 = np.meshgrid(j+1, i+1)

    val0 = np.triu(I/J1)
    val1 = np.triu(I1/J1)
    val1 = np.arcsin(val1) - np.arcsin(val0)

    val2 = ones/I1

    return val1, val2
```

DanHickstein:
Is the _init_abel taking more time than the main onion_peeling function? It might not matter if we vectorize or not...

stggh:
Good point. I was responding to the challenge :-)
```python
def _init_abel(xc, yc):
    # this seems like it could be vectorized pretty easily
```

DanHickstein:
Ha! Yes, I remember skipping that challenge myself as I hastily translated the code from matlab to Python. :)

I think that you are close with the vectorization. You want to be careful about which way your arrays are set up. If you are in the numpy (row, column) indexing mindset, then [meshgrid](http://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.meshgrid.html) has weird default behavior. It's intended to be used like:

```python
x, y = np.meshgrid(x,y)
```

So, I think that you can either switch your 
```python
J, I = np.meshgrid(j, i)
```

to:

```python
I, J = np.meshgrd(i, j)
```
or you can do:
```python
J, I = np.meshgrid(j, i, indexing='ij')
```

The `'xy'` notation makes perfect sense to me in 2D, but as soon as I deal with 3D matrices, I quickly realize that it's completely unintuitive and retreat the the sensibility of `'ij'` indexing.

stggh:
Back on the main `onion_peeling` code:

After playing with radom indices I find that the agreement is improved if `1` is subtracted from `col_index-1` for the initial value of `rest_col`. Note, also, `col_index in range (1, w):` (which was change in the first PR commit).  I don't understand the reason for this, but this is the only offset that produced an improvement.

```python
# or "distance in i direction", I guess
        rest_col = rest_arr[:, col_index-1]
```

onion_1d_2d


The following O<sup>-</sup><sub>2</sub> example uses an even image (not strictly correct size for HL), using the `Q0` and using respective `abel.<method>.<method>_transform()`:

example_hansenlaw_onion_o2


The `onion_peeling` is still not perfect, but it is now, IMO,  in the acceptable range.  

I think that we are now in the regime of tracking down `1/2`-pixel shifts for **all** transform methods.
example_all_o2

stggh:
Vectorized `_init_abel()`  ~30% improvement.

Here is another comparison:
example_hansenlaw_onion

The relative timings:
```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex      14.1                152.3            
       basex_bs    6363.1              19310.7            
  direct_python      69.3                322.2            
      hansenlaw      13.5                 29.9            
  onion_peeling     147.6                493.1       vector _init_abel()     
    three_point       7.9                 78.0            
 three_point_bs     453.3               1481.7 
```
```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex      16.1                120.2            
       basex_bs    6338.4              20298.4            
  direct_python      76.1                334.1            
      hansenlaw      13.4                 29.0            
  onion_peeling     202.8                654.3         original _init_abel()   
    three_point       8.2                 66.6            
 three_point_bs     437.0               1212.6      
```

stggh:
Still to do, work out the correct normalizing factor, now `19.1` (from `21`)

stggh:
Ok, I have double checked:
1. The vector timing  is about 25-30% faster (Mac Book):
 ```
---------------------------------------------------------
          basex       8.9                 51.2            
       basex_bs    7304.7              21100.8            
  direct_python     162.5                883.6            
      hansenlaw      28.1                 64.1            
  onion_peeling     295.7                996.0        _init_abel_vec()          
    three_point      16.2                 46.6            
 three_point_bs    1543.6               4259.0    
 ```
 ```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex       9.5                 66.8            
       basex_bs    6344.1              20176.8            
  direct_python     210.1                845.0            
      hansenlaw      26.3                 67.9            
  onion_peeling     388.7               1327.4            _init_abel()
    three_point      15.4                 46.9            
 three_point_bs    1585.9               4201.1  
 ```
The only remaining vector issue the 1d-case

DanHickstein:
@stggh, we need to check if the arrays produced by the `init_abel` and `init_abel_vec` functions are identical. Currently, it appears that even the array shape is different:

```
In [10]: x, y = abel.onion_peeling._init_abel_vec(5,1); print np.shape(x), np.shape(y)
(6, 2) (6, 2)

In [11]: x, y = abel.onion_peeling._init_abel(5,1); print np.shape(x), np.shape(y)
(5, 5) (5, 1)
```

We may want to carefully read the [Bordas paper](http://scitation.aip.org/content/aip/journal/rsi/67/6/10.1063/1.1147044) and get a better understanding of how the algorithm is supposed to work. For instance, one thing that I notice is that the algorithm is built to incorporate a variable, rho, that can correct the projection for deviations for cases where the ratio of the VMI extraction field is not huge compared to the particle kinetic energy. This is a very unique capability.

If we are just looking to implement the basic onion peeling algorithm, then an efficient algorithm is actually given on P1148 of the [Dasch paper](https://www.osapublishing.org/ao/abstract.cfm?uri=ao-31-8-1146) that @DhrubajyotiDas used to implement the three_point transform. The evaluation of the onion peeling transform proceeded along very similar lines: a basis set is calculated and then the actual transform is just a dot product. (Note that the Dasch paper is just dealing with one-dimensional data, so we would need to loop over the rows, as we have done for the three-point transform. So, the `ij` indices in that paper don't mean quite what you expect them to mean.)

One detail: I thought that we were thinking to just refer to this method as "onion" in the file and function names.

stggh:
1. `onion_peeling` vs `onion` (easy question first). 

 I felt that `onion` was a little too terse, under the same logic `three_point` would be `three`. The user has the option to import a short name. However, it's not a big deal, I am happy to change the name if you think the short form is better.

2. Ok, good point about the array shapes. I will investigate. All my tests were for square and 1d arrays.

3. I was busy, yesterday, on other activities, and have a Workshop today. My descriptions of issues were not the best. 

  IMO, once I sort the vector issue, `onion_peeling` is ready for prime time. It is no worse than the other inverse Abel methods. They **all** do not align peaks for detailed spectral images. 

 I agree that the code should be examined in detail, but this may follow at a later stage.

 You are right about my interpretation of `i, j`. I will revisit the Dasch paper.

DanHickstein:
1) You have convinced me. `onion_peeling` it is!

3) Yes, I agree. If we can get the onion_peeling working on some level, it would be good to put it out there. I would say that we should just add the relevant references to the doc page in case people want to improve it in the future. 

4) <-- Pro tip: I have discovered that if I use parentheses rather than periods for the list numbering, then GitHub won't try to automatically number the list. :)

stggh:
The vector issue is resolved. The problem was in sharing meshgrids, which gave incorrect matrix sizes. Specifically,  `val1.shape = (xc, xc), val2.shape=(xc, yc)`, requires 3 meshgrids.
```
Image shape (h, w) ? 5,1
(1, 1)
[[ 1.57079633]]
------
(1, 5)
[[ 1.  1.  1.  1.  1.]]
====== vec ===
(1, 1)
[[ 1.57079633]]
------
(1, 5)
[[ 1.  1.  1.  1.  1.]]



Image shape (h, w) ? 1,5
(5, 5)
[[ 1.57079633  0.52359878  0.33983691  0.25268026  0.20135792]
 [ 0.          1.04719755  0.38989075  0.27091852  0.21015893]
 [ 0.          0.          0.84106867  0.3244633   0.23198426]
 [ 0.          0.          0.          0.72273425  0.28379411]
 [ 0.          0.          0.          0.          0.64350111]]
------
(5, 1)
[[ 1.        ]
 [ 0.5       ]
 [ 0.33333333]
 [ 0.25      ]
 [ 0.2       ]]
====== vec ===
(5, 5)
[[ 1.57079633  0.52359878  0.33983691  0.25268026  0.20135792]
 [ 0.          1.04719755  0.38989075  0.27091852  0.21015893]
 [ 0.          0.          0.84106867  0.3244633   0.23198426]
 [ 0.          0.          0.          0.72273425  0.28379411]
 [ 0.          0.          0.          0.          0.64350111]]
------
(5, 1)
[[ 1.        ]
 [ 0.5       ]
 [ 0.33333333]
 [ 0.25      ]
 [ 0.2       ]]

```

Here are the relative benchmarks (22% improvement):
**vector**
```
In [2]: abel.benchmark.AbelTiming()
Out[2]: 


=========     Direct Abel implementations ==========
time to solution (millisec)

Implementation     n = 301              n = 501          
----------------------------------------------------------
  direct_python      68.7                285.1            
      hansenlaw      14.0                 28.7            


=========    Inverse Abel implementations ==========
time to solution (millisec)

Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex      16.3                118.8            
       basex_bs    6190.3              19308.3            
  direct_python      71.1                314.1            
      hansenlaw      13.9                 29.0            
  onion_peeling     151.8                504.5            
    three_point       8.1                 64.5            
 three_point_bs     415.9               1163.5   
```
**element**
```
In [2]: abel.benchmark.AbelTiming()
Out[2]: 


=========     Direct Abel implementations ==========
time to solution (millisec)

Implementation     n = 301              n = 501          
----------------------------------------------------------
  direct_python      67.1                287.6            
      hansenlaw      13.5                 30.9            


=========    Inverse Abel implementations ==========
time to solution (millisec)

Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex      14.8                125.8            
       basex_bs    6390.8              20608.4            
  direct_python      71.0                319.1            
      hansenlaw      14.2                 34.1            
  onion_peeling     207.6                650.1            
    three_point       8.6                 64.5            
 three_point_bs     449.3               1433.4   
```

The code could do with better variable names. Once we are happy with it the element version of `_init_abel()` may be removed (to reduce confusion?).
```python
def _init_abel_vec(xc, yc):
    # vectorized 
    i = np.arange(xc, dtype=int)
    j = np.arange(yc, dtype=int)

    Ip1, I = np.meshgrid(i+1, i)
    Ijp1, Iip1 = np.meshgrid(i+1, i+1)

    val1 = np.arcsin(np.triu(Iip1/Ijp1)) - np.arcsin(np.triu(I/Ip1))

    Jp1, I1 = np.meshgrid(j+1, i+1)
    val2 = 1.0/I1[:]

    return val1, val2

def _init_abel(xc, yc):
    # this seems like it could be vectorized pretty easily
    val1 = np.zeros((xc, xc))
    val2 = np.zeros((xc, yc))

    for ii in range(xc):
        for jj in range(ii, xc):
            val1[ii, jj] = np.arcsin((ii+1)/(jj+1)) - \
                           np.arcsin((ii)/(jj+1))

        val2[ii, :] = 1.0/(ii+1)

    return val1, val2
```

DhrubajyotiDas:
@stggh Looking at the Gaussian profiles in [your comment above](https://github.com/PyAbel/PyAbel/pull/155#issuecomment-199641883), the onion_peeling solution seems to be systematically off everywhere from the analytical Gaussian solution. Has that issue been resolved with your latest vector-issue-solution? Apologies if I'm a bit slow to the whole discussion.

@DanHickstein The Dasch paper does have a number of different algorithms including the onion, which are similar in implementation to the three_point. It would be nice if PyAbel could embrace them all :)

stggh:
@DhrubajyotiDas yes, but the difference is sub pixel (I believe). This issue is still present, and is not related to the vectorization. More likely some index offset, which might only be resolved with some detailed investigation into the algorithm code, as @DanHickstein alluded to. To be fair, all the methods should be compared, under the same test, before we may state that this is a bad or acceptable result.

ChrisRallis:
Regarding performance, the matlab implementation was written to process the image column by column, since matlab's internal storage is column major.  This allowed matlab to go off into its native functions for more work.  When I implemented the C version, I processed the rows individually, due to way better CPU cache hits and parallel threading.  That would process an image in a few ms, if I remember right.  I have only briefly used numpy, so I'm not sure what programming patterns are better performing than others.  If we were looking at a C\C++ implementation, I would be of a lot more help.

I can certainly add comments to the code when it's stabilized.

stggh:
@ChrisRallis thanks for your helpful comments. If you have C-code available to examine, that would be great, to track down any row/column issues, or index off-by one problems. Numpy and C have a common array storage structure. In general, C-programs will execute faster than Python, until various Python optimizations are made. The big time saving in using Python is in coding the algorithm (which tends to be fun), as you may have found with your Matlab code.

stggh:
**Solved**  (sort of)

Here is a comparison of the speed distributions for the various methods (- direct)

cf_2d

The excellent agreement is due to (apart from the small scales), the same kludge that I use for `hansenlaw`:
```python
    # for some reason shift by 1 pixel aligns better? - FIX ME!
    # Just like hansenlaw
    abel_arr = np.c_[abel_arr[:, 1:],abel_arr[:, -1]]
```

stggh:
Just to qualify my statement above. The `forward` and `inverse` `hansenlaw` algorithms use identical (array index) code (apart from changing `f` to `g'` and a slight modification to the gamma function). 

The `forward` transform is identical to the input Gaussian (apart from scaling). The `inverse` algorithm requires the reconstructed `Q0` image to be shifted 1-column to the left.

stggh:
With the correct normalization <math>&sigma; &radic; &pi;</math>, the remaining ratio is given in the legend. Note, `hansenlaw` improves with the image size, here `n=201`. The two onion boxes correspond to `n=200 -ev` and `n=201 -od`.

cf_2d

`three_point` looks magnificient

Code [check_allv2.py](https://gist.github.com/stggh/39e26eba764c3f0e3b88)

ChrisRallis:
I can't seem to find the working C source.  It's on some cloud storage or HDD somewhere, but I don't have the time to go looking for it now.  In any case, it was directly based off the matlab code and produced the exact same result.  I wouldn't be surprised if there was an off-by-one error, but it was never enough to substantially affect the range of KER we were looking at.  It could definitely cause issues with the center-line and really low r.  A careful comparison of the Bordas equations and the code should reveal any mistakes.

The centering\shifting that is done at the beginning was just to compensate for the detector not being centered in the camera.

DanHickstein:
@ChrisRallis, While I understand the general idea of the Onion-peeling algorithm, I find the equations in Bordas to be really confusing, and I have struggled for a long time to connect what is happening in your matlab program to what is happening in the Bordas paper and what makes intuitive sense. 

So, perhaps you could add some comments to the core sections of your matlab code to let us know what is going on, and perhaps what equations in Bordas they correspond to. Or, maybe you can provide a good explanation of how we should think about numerically implementing the onion peeling transform. With some of the other transform methods, we processed column-by-column, and this worked pretty well in Python (the general rule in python, like matlab, is to avoid too many nested loops)

`init_abel.m`:
```matlab
function [val1,val2]=init_abel(xcg,ycg)
xc = double(xcg);
yc = double(ycg);
val1=zeros(xc+1,xc+1);
val2=zeros(xc+1,yc+1);

for ii = 1: xc+1
    for  jj = ii: xc+1
        val1(ii,jj)=asin(ii/jj)-asin((ii-1)/jj);
    end
end
for idist = 1:xc+1
    for  jdist = 1:yc+1
        val2(idist,jdist)=sqrt(idist^2+(jdist-1)^2)/idist;
    end
end
end
```

From `aggie_image_processing.m`
```matlab
....
% Initialize abel routine
[val1,val2]=init_abel(xcl,yc);
abel_arr=work_im*0;
rest_arr=work_im;
vect=(1:ymax)*0;

% process right half of the image
for col_index = xmax:-1:xcr
    idist = col_index-xcr;
    rest_row = rest_arr(:,col_index);
    normfac = 1/val1(idist+1,idist+1);
    for i = idist:-1:0
        ic = xcr + i;
        shadow=rest_row*normfac*val1(i+1,idist+1);
        rest_arr(:,ic)=rest_arr(:,ic)-shadow;
        rest_arr(:,ic)=max(rest_arr(:,ic),0);
    end
    for row_index =1:ymax
        jdist=abs(row_index-yc);
        vect(row_index)=val2(idist+1,jdist+1);
    end
    abel_arr(:,col_index)=normfac*rest_row.*vect';
end
```

stggh:
1. @DanHickstein 
 > processed column-by-column

 This is interesting terminology. I have always considered the transform operation to be processing  one image row at a time, and hence, row-by-row. However, you are strictly correct, the Python code loops over the each column (index), of a given row.  ;-)

2. How do you view the [above fits ](https://github.com/PyAbel/PyAbel/pull/155#issuecomment-200280220), sorry about the incorrect plot titles, should be `n=200`, `n=201`.

  I think that this PR is almost ready to merge. Any issues with index offsets (for `onion_peeling` and/or `hansenlaw`) may be dealt with following more in depth code reviews.

stggh:
Here is my version of `dasch_onion_peeling.py`:
```python
# -*- coding: utf-8 -*-

import numpy as np
import abel
from scipy.linalg import inv
from scipy import dot

def dasch_onion_peeling_transform(IM, dr=1, direction="inverse"):
    """ Onion-peeling deconvolution of Dasch et al.
    Parameters
    ----------
    IM : 1D or 2D numpy array
        right-side half-image (or quadrant)

    dr : float
        not used (grid size for other algorithms)

    direction: str
        only the `direction="inverse"` transform is currently implemented
   

    Returns
    -------
    AIM: 1D or 2D numpy array
        the inverse Abel transformed half-image 

    """

    if direction != 'inverse':
        raise ValueError('Forward "dasch_onion_peeling" transform not'
                         ' implemented')

    # make sure that the data is the right shape 
    #(1D must be converted to 2D):
    IM = np.atleast_2d(IM)

    cols, rows = IM.shape
    AIM = np.zeros_like(IM)

    W = np.zeros_like(IM)

    I, J = np.diag_indices_from(IM)
    Iu, Ju = np.triu_indices(W.shape[0], k=1)

    for i in I:
        W[i, i] = np.sqrt((2*i+1)**2 - 4*i**2)

    for i,j in zip(Iu, Ju):
        W[i, j] = np.sqrt((2*j+1)**2 - 4*i**2) -\
                  np.sqrt((2*j-1)**2 - 4*i**2)

    D = inv(W)

    for i, P in enumerate(IM):
        AIM[i] = dot(W, P)

    if AIM.shape[0] == 1:
        # flatten array
        AIM = AIM[0]

    return AIM
```

stggh:
The comparison, bottom left figure:
cf_2d

I am not sure about the normalization, at the moment using `1/cols`, but this apparently requires another `x0.32`

The meshgrid code above could probably be used in `three_point` to eliminate the `for loops` and `if tests`.

stggh:
benchmarks:
```
=========    Inverse Abel implementations ==========
time to solution (millisec)

Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex      14.8                119.7            
       basex_bs    6161.6              19567.9            
dasch_onion_peeling     118.9                398.9            
  direct_python      68.7                356.6            
      hansenlaw      14.5                 30.9            
  onion_peeling     198.0                616.8            
    three_point       7.8                 67.1            
 three_point_bs     445.6               1189.2     
```

DOP (dasch_onion_peeling) may benefit from the "sneaky" basis storage of `three_point` and `basex` ;-)

stggh:
Unfortunately, my `dasch_onion_peeling` joy is short lived. The comparison using the Dribinski sample image `n=501` (to make `hansenlaw` look better ;-) ) reveals an issue with the `dasch_onion_peeling`:

cf_all_dribinski

code for the above figure [here](https://gist.github.com/stggh/ce1ff510c9948a90dccf)

stggh:
Now all good!
I found a typo Eq(1)      `AIM[i] = dot(D, P)   # Eq. (1)`   `W->D`, which probably occurred while I was messing with vectorization.

cf_all_dribinski

stggh:
Now the code template is available for `dasch_onion_peeling.py` it is relatively straightforward to code `two_point.py`. Although, not quite there, diverges near the center.

@DhrubajyotiDas are there typos in Eq. (8) or (9)?

dasheq89


Clearly, `J_ij` is invalid for `j=i`.

cf_all_dribinski

stggh:
Just to summarize the status of this PR:

1. I believe that `onion_peeling` ready for release.  Any offset issues may be examined later.  I made the `example_onion_peeling.py` (and dropped the comparative `example_hansenlaw_onion.py`). If agreeable, `_init_abel()` can be removed, to leave only the vectorized version?

2. Dasch's `dasch_onion_peeling` is also available, including `example_dasch_onion_peeling.py`, and docs. The only issue is the transform normalization factor, which I thought may be `1/n`, but this is not the case. This method probably needs one of @DanHickstein's clever/compact names. I also need to write a unit test.

3. `two_point.py` is close. No docs, examples, or unit test yet.

DanHickstein:
Wow! So much cool stuff! Thanks @stggh!!

There is a typo in Eq. 9 of Dasch. This is described on p170 of [Karl Matrin's thesis]:(https://www.lib.utexas.edu/etd/d/2002/martinkm07836/martinkm07836.pdf)

image
)

image

DanHickstein:
And yes, I think that you do want to store the basis set for the `dasch_onion_peeling` transform in order to speed up repeated transforms. If we do this, then is should be one of the fastest methods - comparable to the other Dasch methods, since they all reduce to just a dot product once the basis sets have been generated.

Also, I think that I have figured out a way to avoid looping over the rows when we are performing the row-wise dot product. In the onion peeling algorithm, I replaced 

```python
inv_IM = np.zeros_like(IM)

for i, P in enumerate(IM):
    inv_IM[i] = scipy.dot(D, P)
return inv_IM
```

with 
```python
return np.tensordot(IM,D,axes=(1,1)) # a one-line Abel transform!
```
It seems to produce identical results with a significant speedup for 500x500 images. More testing may be needed to make sure that this scaling continues for larger images and that the results are identical with the for-loop method. 

Also, I think that this "tensordot" code obscures what is actually happening, and so probably should be accompanied by the original loop code (commented out).

It seems that you have found a faster way to generate the basis sets. Perhaps that should be incorporated into the three_point algorithm as you suggest.

As far as names go, it seems that our implementation of the Dasch OP method is currently faster and cleaner (code-wise) than the "Bordas" OP method. Also, the Dasch paper is 1992 and Bordas is 1996. So, I wonder if we should keep the Dasch method as the `onion_peeling` method and call the other method `onion_bordas` or something like this. Or, we could put them on equal footing and refer to them as `onion_dasch` and `onion_bordas`. 

All of this talk of onions is making me hungry! Last night, in honor of our developments with the onion peeling methods, I cooked a recipe for [chicken and plantains](http://www.food.com/recipe/roast-chicken-with-plantains-and-onions-137093) that used 3 whole onions. :)

>This is interesting terminology. I have always considered the transform operation to be processing one image row at a time, and hence, row-by-row. 

Yes, in the Abel transform, the rows are independent, so (somewhat counterintuitively), this allows us to avoid looping over the rows!

DhrubajyotiDas:
@DanHickstein That recipe looks good, but I think it needs more code review. Could you post some examples? ;)

Good catch on the tensordot! The looping over all rows was rather inelegant for what seemed to be a pretty elementary mathematical operation. 

@stggh Great job with coding up the other Dasch algorithms! The original paper concluded that the `three_point` algorithm provided superior noise-suppression. It would be interesting to see if that still holds with raw data that has such high S/N as ours. If the speed advantage is significant enough, it could be one step closer to the dream of 'real-time' inversions.

DanHickstein:
@DhrubajyotiDas, yes, some optimizations are needed for the recipe. I would recommend doubling the plantains and using less chicken. Parallelizing the onion chopping and the initial roasting step saves time. :)

I don't want to speak too soon, but with the `tensordot` hack, I saw a 1001x1001 three-point transform take only 28 ms of my Macbook, which would allow us to make a claim of "HD video rates".

stggh:
Excellent! What a team.

cf_all_dribinski


1. `two_point` now operational, corrected algorithm according to Matrin's thesis. Thanks guys!
    @DanHickstein `np.tensordot` is a great find. Note, I prefer clean code, and so remove the old code lines, on the basis that it may be explained in the docs. However, it's your call.

2. Renamed the onion methods to `onion_dasch` and `onion_bordas`. Updated docs. I will check through them today (Friday, Easter holiday in Oz) and add some information on `two_point`. I think that I have mangled the docs in places. Also, transform descriptions, could describe all the onion transforms under one, rather than several headings.

3. For `onion_dasch_transform()` note `D = inv(W)` forces a square image (I believe). Hence, this may be one limitation. The square image limitation may also apply for `two_point`, I will check.

4. Normalization, appears to be simply `dr`, I will check.

5. Benchmarks (Macbook pro 2.4GHz i5 - python seems to think is i386?). `two_point` appears to be the slowest, so still room for improvement with that code with `J(i, j)` and storing the basis (I will also look at):
```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex       5.5                 44.2            
       basex_bs    6505.1              20342.3            
  direct_python     133.4                886.5            
      hansenlaw      25.7                 77.7            
   onion_bordas     368.4               1489.6            
    onion_dasch     200.8                729.7            
    three_point      12.2                 32.8            
 three_point_bs    1498.6               4707.1            
      two_point     536.8               1723.2            

```
Updated code than produced the plot [cf_all_dribinski.py](https://gist.github.com/stggh/7ac020205d8f8c6fb4db)

DanHickstein:
@stggh, great work! As usual!

1) Sure, you can go ahead and remove the old code, just add a comment in the code that says what the weird tensor-dot thing is doing: 

```python
# performing a dot product of each row of IM with D:
return np.tensordot(IM,D,axes=(1,1))
```

2) It seems fine to combine `onion_dasch` and `onion_bordas` onto one doc file/page.

3) The Dasch paper is fundamentally one-dimensional, so his method should be applicable to any sized image (we were/are looping over the rows). I think that the requirement that D is square doesn't mean that the image must be square, I think it just means that the transform of a row has the same shape as the row which seems reasonable. 

4) Normalization is just `dr`?! Hurray! It's nice when math works.

5) These should see huge speedups with the basis set generation extracted. Are you benchmarking three-point with the `tensordot` of the old version?

DhrubajyotiDas:
The only normalization I had to do for `three_point` was `dr` as well. I would expect the same for Dasch `onion` and `two_point` as well. 

@DanHickstein, you're correct about the relationship between the square `D` matrix and the size of each row of the image. This was something that I tripped up on as well. For the purposes of the Dasch algorithms, the images are all one-dimensional rows of pixels, and the square `D` matrix accounts for all interactions between the `i`th pixel and all non-`i`th pixel in each row. Hence it's size needs to be `n x n`, where `n` is the number of pixels in each row of the image. The row-independence of the Abel transform is made explicit here.

stggh:
Thanks for the explanation of the D-matrix size.

Restructured `two_point` to removed `if` test, and `for` loops, now the timing that we expect.
```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex       6.3                 40.7            
       basex_bs    7561.8              19287.6            
  direct_python     184.7                725.3            
      hansenlaw      30.8                 62.6            
   onion_bordas     436.8               1192.6            
    onion_dasch     274.4                548.3            
    three_point      16.2                 26.0            
 three_point_bs    1789.6               4009.5            
      two_point       9.4                 22.5        
```

stggh:
Same for `onion_dasch`, removed `for` loops:
```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex       5.5                 41.7            
       basex_bs    6838.4              19606.2            
  direct_python     138.9                753.1            
      hansenlaw      27.0                 68.3            
   onion_bordas     355.1               1257.0            
    onion_dasch       7.8                 28.1            
    three_point      11.8                 50.4            
 three_point_bs    1513.0               4003.2            
      two_point      10.0                 27.5        
```

stggh:
Looks like you are having some fun with the angular integration #156. I will try to look at that later today.

This PR is ready to be merged. 

I have added basis caching to `two_point` and `onion_dasch`. Note the basis storing function is common to most basis methods, so it is placed in `abel/tools/basis.py` as `abel.tools.basis.get_bs_cached()`. I can update `three_point` to use it, but thought that could wait until `three_point` basis calculation is optimized. The docs have been updated, but could probably do with improving.  I left the onion descriptions separate. Unit tests have been added for `onion_bordas`, `onion_dasch`, and `two_point`.

Here is some benchmark timings (Macbook pro 2.4GHz i5):

pyabel-benchmarks

```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex       5.4                 54.5            
       basex_bs   10343.2              24118.9            
  direct_python     205.6                808.0            
      hansenlaw      54.4                 64.4            
   onion_bordas     616.4               1150.3            
    onion_dasch       5.9                 11.9            
 onion_dasch_bs       6.5                 17.5            
    three_point      18.0                 28.5            
 three_point_bs    2131.9               4705.3            
      two_point       5.1                  8.2            
   two_point_bs       6.0                 15.5      
```

DanHickstein:
Strong work @stggh!

The times for the basis set generation for two_point and onion_dasch look much too fast. Can you check that they are actually being regenerated?

stggh:
I added `verbose=True` and received all the basis generation messages. Recall that the basis generation is trivial (and almost hardly worth the use of basis caching):
```
In [2]: abel.benchmark.AbelTiming()
In [2]: abel.benchmark.AbelTiming()
A suitable operator matrix for 'two_point' was not found.
 A new operator matrix will be generated.
A suitable operator matrix for 'onion_dasch' was not found.
 A new operator matrix will be generated.
A suitable operator matrix for 'two_point' was not found.
 A new operator matrix will be generated.
A suitable operator matrix for 'onion_dasch' was not found.
 A new operator matrix will be generated.
 :
```

DanHickstein:
Woah... Those are really fast then!!

stggh:
Actually, you have a point. I am not sure that I read the pre-calculated basis sets in the transform calculation.
Checking ...

stggh:
The basis sets were in fact being recalculated!  Here's the correct timing diagram. Note, `two_point` follows `onion_dasch` as that they both have the same one-line Abel transform.  

pyabel-benchmarks
```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex       9.0                 60.0            
       basex_bs    7799.6              25064.5            
  direct_python     206.3               1500.9            
      hansenlaw      58.8                 91.2            
   onion_bordas     650.6               1610.4            
    onion_dasch       3.0                 11.3            
 onion_dasch_bs       7.7                 30.6            
    three_point      24.3                 60.3            
 three_point_bs    1745.1               5552.8            
      two_point       5.1                 10.9            
   two_point_bs       5.7                 24.8
```
On these timings, it is probably not worth going to the trouble to save the basis for `onion_dasch` and `two_point`.

stggh:
`benchmark.py` was not quite correct for `two_point`, and `onion_dasch`. I added `_core_transform(IM, D)` methods, is is the case for `basex`, `three_point` to allow direct transform when the basis is known.  
This, further reduces the 2pt, dasch execution times. `benchmark.py` is a bit of a mess. I will try to re-write it, to use <method> loops.
pyabel-benchmarks

```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex       9.1                 49.9            
       basex_bs    6482.2              22786.7            
  direct_python     153.4                878.7            
      hansenlaw      27.0                 67.1            
   onion_bordas     352.7               1259.0            
    onion_dasch       2.5                  6.8            
 onion_dasch_bs       6.8                 21.5            
    three_point      15.9                 47.0            
 three_point_bs    1498.4               5362.5            
      two_point       1.4                  7.4            
   two_point_bs       5.5                 16.6      
```

DanHickstein:
@stggh, great work on this! Those times are impressive. I assume that you have not made the same change to `three_point`? You can probably go ahead and do that. And yes, I find benchmark.py a bit confusing at this point. Perhaps this should be saved for a separate PR? I think that maybe it should take a keyword argument of the methods that should be run. I often find myself wanting to use benchmark.py to make a comparison between two or three methods, and get annoyed that Basex and direct_python are in there slowing things down :)

README.rst
```diff
@@ -34,11 +34,15 @@ The outcome of the numerical Abel Transform depends on the exact method used. So
...
     3. ``direct`` - numerical integration of the analytical Abel transform equations.
 
-    4. ``threepoint`` - the "three point" method of Dasch and co-workers.
+    4. ``two_point`` - the "two point" method of Dasch and co-workers.
 
-    5. ``fh`` - Fourier–Hankel method (not yet implemented).
+    5. ``three_point`` - the "three point" method of Dasch and co-workers.
 
-    6. ``onion`` - onion peeling method (not yet implemented).
+    6. ``onion_dasch`` - the "onion-peeling" deconvolution method of Dasch and co-workers.
```
DanHickstein: "onion-peeling" here but "onion peeling" (no hyphen) in next sentence.

abel/onion_bordas.py
```diff
@@ -0,0 +1,158 @@
...
+
+################################################################################
+#
+# Onion peeling algorithm, also known as "back projection"
+#
+# This algorithm was adapted by Dan Hickstein from the original Matlab 
+# implementation, created by Chris Rallis and Eric Wells of 
+# Augustana University, and described in this paper:
+#
+#  http://scitation.aip.org/content/aip/journal/rsi/85/11/10.1063/1.4899267
```
DanHickstein: can you also put these references into the main onion_peeling_transform docstring?

abel/onion_bordas.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # onion-peeling uses grid rather than pixel values, 
+    # odd shaped whole images require shift image (-1/2, -1/2)
+    if shift_grid:
+        IM = shift(IM, -1/2)
+
+    # make sure that the data is the right shape (1D must be converted to 2D):
+    IM = np.atleast_2d(IM)
+
+    # The original pythod code operated on the left-half image
```
DanHickstein: Maybe this should be restated to: we would like to work from the outside to the inside of the image, so flip the image to put the "outside" at low index values.

abel/onion_dasch.py
```diff
@@ -0,0 +1,102 @@
...
+#################################################################################
+#
+#  Dasch onion-peeling deconvolution
+#    as described in Applied Optics 31, 1146 (1992), page 1148 sect. D.
+#    see PR #155
+#
+# 2016-03-25 Dan Hickstein - one line Abel transform
+# 2016-03-24 Steve Gibson - Python code framework
+# 2015-12-29 Dhrubajyoti Das - three_point code and highlighting the Dasch paper
+#                              see issue #61
```
DanHickstein: Maybe include the url for the issue. This code might work it's way into something completely different in the future, and it would be good to just have the link.

abel/onion_dasch.py
```diff
@@ -0,0 +1,102 @@
...
+        raise ValueError('Forward "onion_dasch" transform not implemented')
+
+    # make sure that the data has 2D shape
+    IM = np.atleast_2d(IM)
+
+    rows, cols = IM.shape
+
+    D = abel.tools.basis.get_bs_cached("onion_dasch", cols, basis_dir=basis_dir)
+
+    # one-line Abel transform - dot product of each row of IM with D
```
DanHickstein: move this comment to the core_transform function above the tensor-dot

abel/onion_dasch.py
```diff
@@ -0,0 +1,102 @@
...
+    inv_IM = _onion_dasch_core_transform(IM, D)
+
+    if inv_IM.shape[0] == 1:
+        inv_IM = inv_IM[0]  # flatten array
+
+    return inv_IM/dr
+
+def _onion_dasch_core_transform(IM, D):
+    """Convenience function to allow direct evaluation of transform if
```
DanHickstein: I don't know if "convenience function" is accurate - this is where the actual transform is done!

abel/onion_dasch.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+
+    # basis weight matrix 
+    W = np.zeros((cols, cols))
+
+    # diagonal elements i = j, Eq. (11)
+    I, J = np.diag_indices(cols) 
+    W[I, J] = np.sqrt((2*J+1)**2 - 4*I**2)
+
+    # upper triangle j > i
```
DanHickstein: Is this also Eq 11?

abel/tests/test_onion_bordas.py
```diff
@@ -0,0 +1,136 @@
+from __future__ import absolute_import
```
DanHickstein: Are all of the tests of the transform methods really similar? Perhaps we just need one test file that loops over all of the methods?

abel/tools/basis.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
```
DanHickstein: This seems like a good idea to use a single function to deal with the loading and saving of basis sets. Is it possible for all of the basis-set-requiring transform functions (basex, three-point) to use this? You can perhaps open a new issue that we need to do this.

abel/transform.py
```diff
@@ -84,8 +87,16 @@ def __init__(self, IM,
             ``direct``
                         a naive implementation of the analytical
                         formula by Roman Yurchuk.
+            ``two_point``
+                        the two-point transform of Dasch and co-workers
```
DanHickstein: Actually, I now realize that the Dash paper is a single-author paper!

doc/Makefile
```diff
@@ -52,7 +52,7 @@ clean:
 	rm -rf $(BUILDDIR)/*
 
 livehtml:
-	sphinx-autobuild -b html -z ../ --open-browser $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	sphinx-autobuild -p 9001 -b html -z ../ --open-browser $(ALLSPHINXOPTS) $(BUILDDIR)/html
```
DanHickstein: What does this do?

doc/abel.rst
```diff
@@ -42,6 +66,11 @@ abel.direct module
     :undoc-members:
     :show-inheritance:
 
+
+Pre- and post-transform image tools
```
DanHickstein: Maybe "Image processing tools"

doc/transform_methods.rst
```diff
@@ -5,19 +5,23 @@ Transform Methods
...
+    4. ``*`` The :doc:`three_point <transform_methods/three_point>` method of Dasch and co-workers, which provides a fast and robust transform by exploiting the observation that underlying radial distribution is primarily determined from changes in the line-of-sight projection data in the neighborhood of each radial data point. This technique works very well in cases where the real difference between adjacent projections is much greater than the noise in the projections (i.e. where the raw data is not oversampled). 
 
-    5. (Planned implementation) The :doc:`Fourier–Hankel <transform_methods/fh>` method, which is computationally efficient, but contains significant centerline noise and is known to introduce artifacts. 
+    5. The :doc:`two_point <transform_methods/two_point>` method is also well described by Dasch. It is a simpler approximation to the `three point` transform. Computationally, very efficient in Python.
```
DanHickstein: the two-point and onion-dasch need asterisks to indicate that they need basis sets.

doc/transform_methods/hansenlaw.rst
```diff
@@ -135,7 +135,7 @@ rediscovered by Jason Gascooke (Flinders University, South Australia) for use in
 his velocity-map image analysis, and written up in his PhD thesis: 
 
 J. R. Gascooke, PhD Thesis: *"Energy Transfer in Polyatomic-Rare Gas Collisions and Van Der Waals Molecule Dissociation"*, Flinders University (2000).
-Unfortunately, not available in electronic format.
+Available in `PDF format <https://github.com/PyAbel/abel_info/blob/master/Gascooke_Thesis.pdf>`
```
DanHickstein: Huzzah!

doc/transform_methods/onion_dasch.rst
```diff
@@ -0,0 +1,67 @@
...
+
+
+How to use it
+-------------
+
+To complete the inverse transform of a full image with the ``onion_dasch`` method, simply use the :class:`abel.Transform` class: ::
+
+    abel.Transform(myImage, method='onion_dasch', direction='inverse').transform
+
+Note that the forward Three point transform is not yet implemented in PyAbel.
```
DanHickstein: Does Dasch provide an equation for the forward transform?

doc/transform_methods/two_point.rst
```diff
@@ -0,0 +1,52 @@
...
+
+
+How to use it
+-------------
+
+To complete the inverse transform of a full image with the ``two_point method``, simply use the :class:`abel.Transform` class: ::
+
+    abel.Transform(myImage, method='two_point', direction='inverse').transform
+
+Note that the forward Three point transform is not yet implemented in PyAbel.
```
DanHickstein: Three -> two

DanHickstein: Three -> two

doc/transform_methods/onion_dasch.rst
DhrubajyotiDas: @DanHickstein Not in the 1992 paper (or anywhere else to the best of my knowledge).

stggh:
@DanHickstein thanks, for all the comments. I will address the changes. 
1. The issue with `benchmark.py` is the the method interfaces are not similar enough that a common call statement may be made. `get_bs_basex_cached()` has `n_nvert`, `n_horiz`, while `three_point`, `two_point`, and `onion_dasch` only generate a square using a single `n` parameter. `direct` has a `backend=C` parameter. Finally, the basis limit for `basex` and `three_point`, and the `n_max_slow` for `direct_python` inhibits simple a simple loop call for each method. 
2. ;-) The "convenience function" was because it is unnecessary, purely for `benchmark` 
3. `livehtml

stggh:
As requested Dasch `three_point` evaluated using np.indices is available. The benchmark is similar to `two_point` and `onion_dasch`.
pyabel-benchmarks
```
Implementation     n = 301              n = 501          
----------------------------------------------------------
          basex       5.4                 44.7            
       basex_bs    6775.0              20729.2            
  direct_python     138.9                892.1            
      hansenlaw      34.8                 64.1            
   onion_bordas     478.8               1147.6            
    onion_dasch       1.6                  5.2            
 onion_dasch_bs      12.7                 17.1            
    three_point       5.6                  5.3            
 three_point_bs      33.9                 45.2            
      two_point       2.5                  5.5            
   two_point_bs       7.8                 15.1      
```
I imagine that similar gains could be made in the `basex` basis calculation, although it is not obvious because of the code complexity.

Q: Where should the old 3pt code be moved (apologies @DhrubajyotiDas )?
In this fork the original `three_point` is currently called `original_three_point()`

I temporarily removed 2 unit tests `test_three_point` and `test_transforms` - until I have time to fix, a bit later.

DanHickstein:
@stggh, thanks for doing this! This tiny PR has evolved into Godzilla. But the improvements have been incredible, as that latest benchmark clearly shows. 

We should probably just check that the new three-point works and then move the legacy code out of the repo. We could place it in the ["PyAbelLegacy"](https://github.com/PyAbel/PyAbelLegacy) repo and then include a hyperlink in three_point.py. In general, all legacy codes that we may want to refer back to can be placed in the PyAbelLegacy. I think that all members of PyAbel Developers should have have write access, but I'll check that this is the case...

DhrubajyotiDas:
@stggh @DanHickstein Moving the legacy code out of the code base is perfectly fine with me!

stggh:
Here is a rough comparison program [compare_3pt.py](https://gist.github.com/stggh/c265232275eb3f2a73ab)
that `assert_allclose`'s (a) the difference in speed profiles for the dribinski sample image, and (b) the D-operator for a `3x3` matrix:
```
python compare_3pt.py 
D = 
 [[ 0.52454873 -0.29289619 -0.10069527]
 [ 0.10364217  0.20655313 -0.16471349]
 [ 0.          0.07412417  0.15436472]]
legacy D = 
 [[ 0.52454873 -0.29289619 -0.10069527]
 [ 0.10364217  0.20655313 -0.16471349]
 [ 0.          0.07412417  0.15436472]]
```
3pt_diff

DanHickstein:
Looks like they are very close! Can you plot the relative difference between the two methods? If they are really the same down to the precision of float64, then I think that they should be the same to within ~10^-15: https://en.wikipedia.org/wiki/Double-precision_floating-point_format

stggh:
A code structure Q:

The only difference between the codes in `two_point`, `three_point`, and `onion_dash` is in the calculation of the basis operator `D` matrix, called `_bs_<method>()` in each file.  To reduce code repetition, I moved each of the duplicated `<method>_core_transform()`'s,  that do the one-line Abel transform `np.tensordot(IM, D, axes=(1, 1))` to `abel.tools.basis.abel_transform()`. This also reduces doc duplication. I am not 100% sure this was a good idea, since the final Abel transform is no longer visible in each <method> file.  However, it seems to me that all the Dasch basis codes (`_bs_two_point()`, `_bs_three_point()`, `_bs_onion_dasch()`) could live in one file `abel/dasch.py`. The only issue is would be the breaking of our standard call method, in this case to `abel.dasch.two_point_transform()` etc, or `abel.dasch.dasch_transform(method="two_point", ...)`. Thoughts?

I realize this PR is causing a bottle-neck for subsequent PRs. It may be merged, subject to our decision on the item above. If time I would like to update `benchmark.py` (to loop over methods, and select methods for timing as @DanHickstein  suggested), however, that could wait for another PR.

Rel diff (matplotlib has a funny y-scale. The ratio is 1.00000...):
3pt_diff

DanHickstein:
@stggh, don't worry too much about bottlenecking other PRs - this one is a big one, and we are covering some important issues.

I am in favor of avoiding code duplication by moving all of the dash methods to a common `dasch.py`. I don't think that it's a big deal to deviate slightly from our previous nomenclature, so long as the call to each method is well described in the docs. Hopefully most users will be calling `abel.transform` anyways.

DanHickstein:
Oh, and differences between new and old three-point on the order of 1e-13 are fine. Larger than I would expect just from the precision of float64, but still very small.

stggh:
I have streamlined `abel/benchmark.py` to loop over the Abel transform methods (and basis set calculations). There is an additional parameter `select` to select transforms to be bench marked:
```
In [1]: import abel

In [2]: abel.benchmark.AbelTiming(n=[101, 501], select=["hansenlaw", "two_point"])
Out[2]: 
PyAbel benchmark run on i386

times in milliseconds


Implementation     n = 101              n = 501          
----------------------------------------------------------
   two_point_bs       2.2                 17.3            


Implementation     n = 101              n = 501          
----------------------------------------------------------
      hansenlaw      11.5                115.6            


Implementation     n = 101              n = 501          
----------------------------------------------------------
      hansenlaw      13.7                 66.5            
      two_point       0.4                  6.1  

          
In [3]: abel.benchmark.AbelTiming(n=[101, 501], select=['all'])
Out[3]: 
PyAbel benchmark run on i386

times in milliseconds


Implementation     n = 101              n = 501          
----------------------------------------------------------
       basex_bs     290.6              20636.2            
onion_peeling_bs       1.4                 19.6            
 three_point_bs       2.4                 50.9            
   two_point_bs       0.6                 16.0            


Implementation     n = 101              n = 501          
----------------------------------------------------------
  direct_Python       8.9               1108.8            
      hansenlaw       5.3                 54.8            


Implementation     n = 101              n = 501          
----------------------------------------------------------
          basex       1.0                 40.1            
  direct_Python      11.8                811.3            
      hansenlaw       6.9                 63.8            
   onion_bordas      38.9               1295.9            
  onion_peeling       0.1                  5.5            
    three_point       0.2                  6.0            
      two_point       0.1                  5.6            

```

abel/dasch.py
```diff
@@ -0,0 +1,315 @@
...
+
+    inv_IM = abel_transform(IM, D)
+
+    if rows == 1:
+        inv_IM = inv_IM[0]  # flatten array
+
+    return inv_IM/dr
+
+
+def abel_transform(IM, D):
```
DanHickstein: I'm hesitant to call a function "abel_transform". Maybe "dasch_transform"?

DanHickstein:
@stggh, I think that this looks good! 

@PyAbel/developers, any more suggestions/code-review before we merge this?

DanHickstein:
I tried abel.benchmark.AbelTiming() and got the error:
```TypeError: basex_core_transform() got an unexpected keyword argument 'direction'```

Also, DirectC seems to always run. I think that the following code-block needs to be above the section that deletes the unwanted methods from the list:
```python
        if direct.cython_ext:
            res['forward']['direct_C'] = []
            res['inverse']['direct_C'] = []
```

stggh:
Good find. My test system did not have cython. 

I have not been able to reproduce the error:
```
In [1]: import abel

In [2]: abel.benchmark.AbelTiming()
Out[2]: 

times in milliseconds


Implementation     n = 301              n = 501          
----------------------------------------------------------
        basex_bs    6496.4              20435.7            
onion_peeling_bs       7.5                 28.7            
  three_point_bs      11.1                 29.8            
    two_point_bs       4.1                 12.5            


Implementation     n = 301              n = 501          
----------------------------------------------------------
   direct_Python      70.9                360.9            
       hansenlaw      15.3                 27.8            


Implementation     n = 301              n = 501          
----------------------------------------------------------
           basex      15.7                134.4            
   direct_Python      78.0                372.1            
       hansenlaw      14.4                 30.3            
    onion_bordas     212.4                649.7            
   onion_peeling       4.3                 19.4            
     three_point       4.8                 19.3            
       two_point       4.5                 21.1            

In [3]: abel.benchmark.AbelTiming(select=["basex"])
Out[3]: 

times in milliseconds


Implementation     n = 301              n = 501          
----------------------------------------------------------
        basex_bs    6243.6              19723.2            


Implementation     n = 301              n = 501          
----------------------------------------------------------
           basex      13.8                119.7            
```

stggh:
I re-cloned, and now have direct_C, but still no issue:
```
In [1]: import abel

In [2]: abel.benchmark.AbelTiming()
Out[2]: 

times in milliseconds


Implementation     n = 301              n = 501          
----------------------------------------------------------
        basex_bs    6184.1              19311.2            
onion_peeling_bs       6.5                 28.8            
  three_point_bs       9.8                 27.6            
    two_point_bs       4.2                 11.5            


Implementation     n = 301              n = 501          
----------------------------------------------------------
        direct_C      16.2                 71.1            
   direct_Python      64.2                285.0            
       hansenlaw      11.9                 26.7            


Implementation     n = 301              n = 501          
----------------------------------------------------------
           basex      13.9                119.0            
        direct_C      22.4                 95.3            
   direct_Python      71.5                302.5            
       hansenlaw      13.8                 31.3            
    onion_bordas     209.7                642.7            
   onion_peeling       4.1                 18.7            
     three_point       4.0                 18.9            
       two_point       4.1                 18.8            

In [3]: abel.benchmark.AbelTiming(select=["basex", "hansenlaw"])
Out[3]: 

times in milliseconds


Implementation     n = 301              n = 501          
----------------------------------------------------------
        basex_bs    6835.4              22183.9            


Implementation     n = 301              n = 501          
----------------------------------------------------------
       hansenlaw      15.8                 27.1            


Implementation     n = 301              n = 501          
----------------------------------------------------------
           basex      13.9                123.0            
       hansenlaw      15.3                 29.0     
```

stggh:
There does appear to be an issue that is size dependent. `n=601` works, `n = 701`  fails. Investigating ...

stggh:
@DanHickstein, it took me a while to catch on ... 
The `n > n_max_bs` where no basex calcuation is made (and no basis calculated) was not handled correctly.  Added code to correct. 
```
In [3]: abel.benchmark.AbelTiming(n=[701,])
Out[3]: 

time in milliseconds

Implementation      n = 701       
        basex_bs       nan         
onion_peeling_bs      66.3         
  three_point_bs      67.1         
    two_point_bs      23.7         


Implementation      n = 701       
        direct_C     187.5         
   direct_Python       nan         
       hansenlaw      43.5         


Implementation      n = 701       
           basex       nan         
        direct_C     245.4         
   direct_Python       nan         
       hansenlaw      49.6         
    onion_bordas    1461.4         
   onion_peeling      48.1         
     three_point      48.0         
       two_point      48.0         

```

DanHickstein:
Thanks @stggh! 

I think that you can probably merge now if you like.

stggh:
Closing summary (of PR 155)
------------

- This PR started out as a bug fix for @DanHickstein's `abel/onion_peeling.py` algorithm #56. Trial and error code hacks eventually gave transforms close to the other methods, although there is a general lack of satisfaction due there being no particular justification for the changes.

    `onion_bordas` (renamed) is the **only method** currently available in PyAbel that "officially" transforms **odd-width size images** (not confirmed).

- Vectorization of the onion-peeling basis code was inspired by @DanHickstein code comment:

   > def _init_abel(xc, yc):
            # this seems like it could be vectorized pretty easily

    This led to some fun with meshgrid, and finally a vector solution that gave ~20% execution speed improvement.

- @DanHickstein  and @DhrubajyotiDas, also provided another challenge:
 
  >If we are just looking to implement the basic onion peeling algorithm, then an efficient algorithm is actually given on P1148 of the Dasch paper that @DhrubajyotiDas used to implement the three_point transform  [link](https://github.com/PyAbel/PyAbel/pull/155#issuecomment-199850696)

    >Dasch paper does have a number of different algorithms including the onion, which are similar in implementation to the three_point. It would be nice if PyAbel could embrace them all :) [link](https://github.com/PyAbel/PyAbel/pull/155#issuecomment-200095291)

    The knowledge from the onion peeling vectorization, and the use of numpy array indice functions provide for a simple implementation of Dasch "two-point" algorithm. 

 @DanHickstein  [revealed](https://github.com/PyAbel/PyAbel/pull/155#issuecomment-200897056)  the "magical" `np.tensordot(IM, D, axes=(1, 1))` to yield a one-line Abel transform, for a given basis matrix operator `D`.

    `two_point.py` provided a template for implementing "onion-peeling", and eventually "three-point". `three-point.py` debugging was made much simpler by the availability of the @DhrubajyotiDas's version, which now re-located in [PyAbel-legacy](https://github.com/PyAbel/PyAbelLegacy).

    The three new Abel transform methods are located in the file `abel/dasch.py`, with the separate  Bordas onion peeling code renamed to `onion_bordas.py`.

- The new algorithms added more code to the bulky `abel/benchmark.py`. This has been streamlined (including the bug above!), to hopefully make it easier to add yet more algorithms to be benchmarked.

Conclusion
--------------
This has been a wonderful team effort, resulting in 3-new (Dasch) Abel transform methods that are code compact and very fast. `three_point.py` may prove to be the algorithm of choice, until the next development ...

(final) benchmark figure (note transform colors not the same as in the figures above):
pyabel-benchmarks

PR #156, opened 2016-03-25
add averaged radial intensity

ldes89150:


DanHickstein:
This looks good to me! @stggh can merge this if it's not going to cause issues with his other PR. Otherwise we can wait and merge that one first.

stggh:
Thanks @ldes89150, this PR actually provides a good discussion point:

1. In principle the angular_integration should be done properly, as an integration, using a quadrature function.

2. Within the current implementation, is there any need to stick with `sum`? It could be replaced with `average` with the saving on the extra parameter and docs.

DanHickstein:
@stggh, interesting discussion!

1) Unless there are features that oscillate rapidly as a function of angle, a simple sum should provide the same results as a more sophisticated numerical integral. But, I suppose there there is really no reason not to replace:
```python
speeds = np.sum(polarIM, axis=1)*dr
```
with something like
```python
speeds = np.cumtrapz(polarIM, axis=1, dx=dr)
```
(Note that [`np.romb`](http://docs.scipy.org/doc/scipy-0.14.0/reference/integrate.html) and [`np.simps`](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.integrate.simps.html#scipy.integrate.simps) require that the array have a certain size, so they would complicate things.

2) This question of a sum versus an average has me struggling with the Jacobians all over again. To make sure that we are doing things correctly, I think that we should perform a test where we make a sample image with a certain total integrated intensity. Then, we should demonstrate that we can do the angular integration with different `dr` and `dt`, integrate the result, and get the same number. At the end of the day, that is what we need to achieve. Information about the total number of counts should not be lost.

Currently, it seems that if `dt` is changed, then the value of the sum would change. So, it seems that multiplying by some factor of `dt` is needed. I think that the average would almost work, but it would likely be off by some factor of pi.

My intuition is that both the average and the integral should exist. There will be people – like photoion spectroscopists – who want values that can be interpreted as the total number of ions collected. On the other hand, there will be people processing images who want to compare the radial integration with simple slices through the image. So, I think that there are cases for both the integral and the average.

ldes89150:
@DanHickstein 
I'm not sure why the summation should times `dr`.

My naive idea is:

integration
```
np.sum(polarIM, axis=1)*(dt/degree)
```

averaged radial intensity
```
np.average(polarIM, axis=1)
```

DanHickstein:
The summation was originally not multiplied by `dr`. @stggh included this in order to keep the values in the output array the same for all values of `dr`.

I'm pretty lost in the math at the moment. I think that we just need to try a practical test and make sure that the integration maintains the right value.

ldes89150:
@DanHickstein 
I believe `dr` should be included in `index_coords` function. the `x` and `y` should be scaled accroding to `dr`.

DanHickstein:
> I believe dr should be included in index_coords function. the x and y should be scaled accroding to dr.

Yeah, I was coming to the same conclusion...

ldes89150:
@DanHickstein 
The unit of `dr` seems to be `pixel of the image`

DanHickstein:
Yes, that's true. That's a physical unit for most CCD cameras. In any case, it's important to keep track of how the bins of the integrated image relate to the original bins in order to keep track of the absolute intensity.

ldes89150:
@DanHickstein 

I tried to plot something like this:
```
r, fr = abel.tools.vmi.angular_integration(ab,dr = 1, average=False)
plt.plot(r, fr,)
```

The intensity scales with `dr`. However, I believe `dr` should only affect the sampling rate instead of the intensity.

DanHickstein:
The intensity changes with dr? Hmmmm.... I thought that was what @stggh was trying to avoid.

See our discussion here: https://github.com/PyAbel/PyAbel/pull/146

ldes89150:
@DanHickstein Can you test it on your computer? I saw this phenomenon in mine.

DanHickstein:
Hmmm, I don't think that the intensity changes with dr. 

Here is my test: 

```python
import abel

IM = abel.tools.analytical.sample_image()

r1, speed1 = abel.tools.vmi.angular_integration(IM, dr=1)
r2, speed2 = abel.tools.vmi.angular_integration(IM, dr=2)

plt.plot(r1, speed1, label='dr=1')
plt.plot(r2, speed2, label='dr=2')

plt.legend()
plt.show()
```
output:
figure_1

ldes89150:
@DanHickstein  uh... It works differently on my computer.

Mine version is `commit b176f87af12847ae140c4c23c100ac2e086c9590`

DanHickstein:
My mistake. My fork was out-of-date. Using the newest version it works as you say:

figure_1

I don't necessarily think that this is crazy behavior, as the bin size is increased, the number of counts in the bin should change. @stggh would disagree...

abel/tools/vmi.py
```diff
@@ -62,7 +64,11 @@ def angular_integration(IM, origin=None, Jacobian=False, dr=1, dt=None):
     if Jacobian:  # x r sinθ
         polarIM = polarIM * R * np.abs(np.sin(T))
 
-    speeds = np.sum(polarIM, axis=1)*dr
+    if average:
```
ldes89150: How about this??

```python
speeds = np.average(polarIM, axis=1)*dr
if not average:
   speeds *= 2*np.pi
```

ldes89150:
I also found a bug at `reproject_image_into_polar` while processing a rectangular image.

DanHickstein:
Is that actually a bug, or does it just expect the center to be in (x,y) format rather than (row,column) format?

ldes89150:
@DanHickstein If I don't set the origin, the result will be incorrect.

DanHickstein:
Really

ldes89150:
the azimuthal integration
```
integration = \int_0^2pi f(r, theta) d theta
= sum_i^n f(r, 2pi/n*i) 2pi/n
= sum_i^n f(r, 2pi/n*i)/n * 2pi
= average * 2pi
```

ldes89150:
@DanHickstein 
```python
import numpy as np
Nx = 1024
Ny = 768
x = (np.arange(1,Nx+1)-Nx/2)
y = (np.arange(1,Ny+1)-Ny/2)
xx, yy = np.meshgrid(x,y)
data = np.array(np.sqrt(xx**2+yy**2)<100, dtype = np.float)
plt.imshow(data)
plt.show()
import abel
o,r,t = abel.tools.polar.reproject_image_into_polar(data)
plt.pcolormesh(t,r ,o)
plt.show()
```

DanHickstein:
Yes, I can reproduce the weird behavior. 

I think that the problem may lie in the polar2cart function (or, perhaps, cart2polar) in polar.py. polar2cart advertises that it generates x and y referenced from theta in the vertical direction, but it provides x = r cos(theta), which seems to assume that theta=0 is in the x-direction.

ldes89150:
change
```
theta = np.arctan2(x, y)  # θ referenced to vertical
```

to 

```
theta = np.arctan2(y, x)  # θ referenced to vertical
```

?

ldes89150:
I think this will fix all the issue

ldes89150:
```python
import numpy as np
Nx = 1024
Ny = 768
x = (np.arange(1,Nx+1)-Nx/2)
y = (np.arange(1,Ny+1)-Ny/2)
xx, yy = np.meshgrid(x,y)
data = np.array((np.sqrt(xx**2+yy**2)<100)*(yy>0), dtype = np.float)
plt.imshow(data)
plt.show()
import abel
o,r,t = abel.tools.polar.reproject_image_into_polar(data)
plt.pcolormesh(t,r ,o)
plt.show()
```

DanHickstein:
Flipping sin and cos in polar2cart makes sense to me. Why did you add a negative in front of the x-term?

ldes89150:
@DanHickstein vertical is the `theta=0` and counter clockwise is positive for `theta` . In this convenvtion, `x=-r sin( theta)`

You can draw it on paper :D

stggh:
Good discussion, although somewhat frustrating that this code must be revisited:

1. `dr`is the sampling/grid size in pixels. It should not change the integrated intensity:
 ```python
IM = np.loadtxt("O2-ANU1024.txt.bz2")

# case (a)  dr = 1
AIM = abel.Transform(IM, method="hansenlaw").transform

speed = abel.tools.vmi.angular_integration(AIM)

# case (b)  dr = 0.5
AIM2 = abel.Transform(IM, method="hansenlaw", transform_options=dict(dr=0.5)).transform

speed2 = abel.tools.vmi.angular_integration(AIM2, dr=0.5)

plt.plot(*speed, label='$dr=1$')
plt.plot(*speed2, label='$dr=0.5$')
  :
 ```
 dr

2. Concerning `reproject_image_into_polar()`. My images are fully symmetric and so this issue was not apparent to me. It is good that you have checked this.
  a.  The original code operated on image files with (0,0) at the bottom-left, hence our numpy arrays are `flipud` at the start. 
  b. At the risk of being directed to a drawing, I believe the update code is not quite there, the angle should be positive?
orig
polar

DanHickstein:
@stggh,

Thanks for doing this analysis! 

The `dr` that is passed to `angular_integration` is, as you say, just an arbitrary grid size that should not affect the intensity *of images processed by `angular_integration`*. This `dr` has nothing to do with the `dr` that is passed to the transform function, which would only be used if the units of the original image are known. Basically, I think that the only application of this `dr` is to achieve agreement with analytical results. Experimentally, I think that it's fine to leave the units in pixels (`dr=1`). 

In general, the `dr` values will be different, and perhaps we should give them more descriptive names.

The `dr` that is passed to angular integration determines how the angular integration is completed. I think that as this `dr` is changed the results of the `angular_integration` function should not change.

Regarding the positive versus negative angle convention - I'm not sure if we had really decided anything formally about where the angles were referenced and which direction they rotated. I think that @ldes89150 was using the convention that zero-angle was in the vertical direction and that it rotated counter-clockwise. So, I think that using this convention (and using the convention that `origin='upper'`) the results are correct. We can discuss if this is really the best convention. 

In general, it seems that we are moving away from the "scientist" convention of having the lower-left corner be 0,0 and towards the computer-scientists definition of having the upper-left corner be (0,0). In addition, we seem to be converging on `center=(row, column)` rather than `center=(x, y)` format. We should probably state this preference explicitly in the docs (perhaps even the readme) and make an effort to standardize PyAbel to use this convention. I think that this section of code that we are dealing with right now is actually the main offender in terms of using (x, y) centers.

ldes89150:
Currently, I followed the convention written in the comment in line 62
```python
r, theta = cart2polar(x, y)  # convert (x,y) -> (r,θ), note θ=0 is vertical
```

I think maybe we should remove the parameter `origin` in `angular_integration` since one can always center their images with the tools provided in `center.py`. (Then we don't need to worry about the convention of x&y here (at least for users) :D)

Besides, do you agree the following change?
```python
speeds = np.average(polarIM, axis=1)
if not average:
   speeds *= 2*np.pi
```

stggh:
1. Coordinates - you are right. I had pictured in my head that theta varies between 0 and &pi;, with positive angles, clockwise from the vertical zero.  I wrote the code comment. It does seem that the convention is anticlockwise.

2. The `origin` parameter is **essential** for `angular_integration`. For image quadrants, `origin=(0,0)`. I agree that we should stick with numpy axes.

3. `2*np.pi` is this from the &phi; integration?  Why is it only applied `if not average`?
    In the end I believe that we should apply the correct integration: 
        <math> &int; r dr  &int; d&phi; &int; I(r, &theta;) sin&theta; d&theta; </math>
   (guessing from memory, it may be an r<sup>2</sup>) , which gives your 2&pi; factor, the correct
   weighting and Jacobian factors.

4. `dr`: From my perspective, provides for sub-pixel sampling when an image is small. e.g., a case where there is improvement for `dr < 1`:
dr

    My `dr` does not alter the transition peak pixel positions.
I am not sure how `dr` parameter is used in `basex`. As some form of scaling parameter to convert pixels to mm (inches ;-))?  It is clear that we may have different meanings, and this should be resolved.

5. Any plan for an PyAbel code Easter egg?  ;-)

DanHickstein:
@stggh and @ldes89150, thanks so much for this discussion. I know it feels a little bit like we are going in circles with this, but it's tough math and we are making good progress!

1) **Coordinate system:** I think that we should standardize to the "mathematical" system of spherical coordinates as shown on the [wikipedia page](https://en.wikipedia.org/wiki/Spherical_coordinate_system):

image

However, we are letting theta run from 0->2pi an phi go from 0->pi. Alternatively, there is a "physics" definition of spherical coordinates on wikipedia that matches what we are doing, except that theta and phi are swapped. So, we could use that system as long as we are okay switching the code from "theta" to "phi". 

This means that the coordinates are referenced from the x-axis (horizontal). I like this for two reasons: first, it's a standard coordinate system and second, it places the zero-angle in the horizontal. I think that most objects that people will be transforming will be things like rods, cylinders, or other objects that have a peak in the density along the vertical axis (the axis of symmetry of the Abel transform). Aesthetically, it doesn't look as nice to generate a polar image that breaks the object into two parts (positive and negative small angles). I think that it would look nicer if the object were at pi/2 and 3pi/2, and would consequently not be split. 

2) I agree that `origin` should stay, as it makes the `angular_integration` function much more useful. 

3) The 2*pi factor comes from d theta which is 2*pi/n. The full Jacobian for integration in spherical coordinates is `r^2 * sin(theta) * dr * dtheta * dphi`. One of the r-factors corresponds to the increasing volume element size with radius, and occurs in the `reproject_image_into_polar` function. The other `r * sin(theta)` factor comes from the fact that we need to integrate over phi (indeed, the Jacobian for integration in two-dimentional polar coodinates is just `r`). 

So, we have already taken into account `r^2 sin(theta)`, but we still need to take into account `dr`, `dtheta`, and `dphi`. I think that we just say that `dphi` is `pi`. `dtheta` is `2*pi/n` as @ldes89150 suggests. Since we are not actually integrating over the r-coordinate, then I think that `dr` should not be used. It could be applied if we want to give the user an array that they can simply sum in order to get the total intensity. 

4) Yes, the `dr` for the `angular_integration` means what you think it does. The `dr` for the transform methods is just a scaling factor. 

5) Perhaps PyAbel should detect that it is being run on Easter Sunday, and return only egg-shaped distributions on that day :)

ldes89150:
@DanHickstein My personal preference is to use the "physics" definition of spherical coordinates. How do you think?

stggh:
1)  I agree that &theta; and &phi; labels should be swapped (physics definition).  This is all that need be done. The reference angle &theta; (physics label) must be with respect to the vertical symmetry axis, &theta; = 0 in the direction of the z-axis. 

I may have misinterpreted the following?

> it places the zero-angle in the horizontal

3)  `angular_integration()` returns the radial array, so `*dr` is required, the grid-sampling factor.

> It could be applied if we want to give the user an array that they can simply sum in order to get the total intensity. 

  With a function called `angular_integration()` the user might expect integration over angle to be involved. However, as @DanHickstein stated, the approximation using `sum or average()xJacobians` may good enough.  

We appear to be in agreement, just a matter of ensuring correct coding of the math.

DanHickstein:
1) Okay, so we want to use the physics definition of the spherical coordinate system. Do I understand correctly that our images are in the xz-plane, and that the axis of symmetry is the z-axis? In this case, it seems that theta is defined are going clockwise. In this case, we would not want to swap to phi, but instead, stick with theta.

3) 
> With a function called angular_integration() the user might expect integration over angle to be involved. 

We are integrating over angle. We are summing and applying the correct dtheta and dphi terms. This is a good approximation of an integral. Are you suggesting that we use some sort of trapezoidal rule? or do I misunderstand your comment. 

We are not integrating over r, so there is a question if we should apply the `dr` to the result or leave that to the user. Both options seem valid to me.

stggh:
Mind blowing.
1) Good point, you are correct.

2) Yes, I was suggesting a trapezodial rule, but I think you are also correct "good approximation of an integral", so no issue. Also, fine with `dr`

DanHickstein:
Okay, to summarize, PyAbel should use the "physics" polar coordinate system described on the [Wikipedia page](https://en.wikipedia.org/wiki/Spherical_coordinate_system):
image
The axis of symmetry is the z-axis, the x-axis is the direction in which the projection takes place (i.e., it points straight into the detector/camera). Our images, therefore, are in the yz-plane. 
This is fits with the drawing of the [Abel transform on Wikipedia](https://en.wikipedia.org/wiki/Abel_transform#Geometrical_interpretation):
image

Also, we need to decide on the range of theta and phi. We should allow theta to cover 2*pi and phi to only cover pi. Do we want to use this convention? 
```
0<theta<2*pi
0<phi<pi
```
Alternatively, we could use
```
-pi<theta<pi
-pi/2<phi<pi/2
```

Regarding @stggh's desire to use a real integral, it should be trivial to convert `np.sum(polarIM, axis=1`) to:
```python
d_theta = 2*np.pi/n
np.trapz(polarIM, axis=1, dx=d_theta)
```
Those two lines could easily be combined, but I think that it's nice to make the d_theta explicit. It has the advantage of providing a slightly more accurate result, and also makes it clear that we are integrating.

ldes89150:
the arctan returns
```
 -pi<theta<pi
```
Therefore, I would suggest using ``` -pi<theta<pi```.

ldes89150:
Besides, I think it's better to use rad instead of degree as the unit of `dt`.

ldes89150:
BTW, should we change data type of the `origin` from `int` to `float`:
```python
origin = (nx//2 + nx % 2, ny//2 + ny % 2) 
```

DanHickstein:
>  I think it's better to use rad instead of degree as the unit of dt or replace dt with tn.

I think that either of those options sound like a good idea. I lean towards specifying the number of theta values (I would call this `nt`). 

> BTW, should we change data type of the origin from int to float

If the code still works fine (or can be made to work) with a float origin, then it seems like a good idea to do this...

ldes89150:
The test failed.... I'm not sure why.

DanHickstein:
You can see the results here: 
https://travis-ci.org/PyAbel/PyAbel/jobs/119117350

You can run the tests on your computer to figure out what's going on:
https://github.com/PyAbel/PyAbel/blob/master/CONTRIBUTING.rst#unit-tests

ldes89150:
@DanHickstein I rollbacked the commit. I think it's better to fix the `origin` issues in another PR.

DanHickstein:
That's fine. But you were probably very close to implementing it successfully. What was failing was a unit test that we wrote to ensure that the size of the arrays generated by the `angular_integration` function were the sizes that we expected. Probably you were off-by-one since you switched to a float. 

I think that we are going to wait a tiny bit more to merge this until @stggh's huge PR is done. So, if you'd like to try to make the float center work, please go ahead. Probably your function is working correctly and you just need to modify the unit test a little bit.

DanHickstein:
I think that this looks fine to me!

@stggh, do you want to hit the green button if you are also satisfied.

stggh:
I ran a couple of tests:

1. radial speed test - looks fine
  speedtest

2. angle polar test = @ldes89150 code above, except `*(yy>0)*(xx>0)` for the image data:
 polar1
 polar2

 There may be some residual <math>&theta; < 0</math> structure?

3. For discussion @DanHickstein, not necessarily for this PR
a. *Radial range of speed distribution*

I often get confused by the fall off in the speed distribution at large radius, since we supply `r = 0, sqrt(w**2+h**2)`  (or `w//2, h//2` for quadrant image sizes). I think the default should be `r_max = min(w, h)`, with perhaps, a bool parameter to provide the extended range?  Here is a dribiniski sample image test case. There are similar results for a Gaussian function.

radialtest

  The above test was run with `Jacobian=True`, to emphasize the effect. The fall-off beyondf `r~180` is due to the extended `r` range.

  b. Scaling may be  is lost for `Jacobian=True`, the figure below has a `112654x ratio` of intensities: 
speedj

DanHickstein:
@stggh, Thanks for taking a careful look at this!

I would suggest that we go ahead and merge @ldes89150's PR and we can address these other issues separately, since I'm not sure if they are directly related to the changes made here.

Regarding Point 2, you are correct that the little slice of intensity on the left (negative angles) doesn't look quite right. Something is shifted by one or two pixels (perhaps your sample image, or the center?)

Regarding 3a, This "fall-off" is just because we are returning the intensities over the whole range of `r` values that we have available, and the outer regions do not have complete data. We thought about including an option to return a different range, but decided that this was fine for the time being. It's fine to re-open that discussion, but we can start a separate issue or PR for that.

Regarding 3b, I'm not sure what you are showing in that plot. Turning the Jacobian on/off should change the scaling, so that looks kinda like I would expect.

Thanks again for the comments!

stggh:
Sorry, for the lack of follow-up on this, I was tied up during the W/E.

I agree, the PR is ready to merge. 
2) The 'bleeding' may be a matplotlib issue.
3a) I will start another discussion.
3b) Yes, the normalized result is what is expected. The issue is the intensity (maximum) differs for `Jacobian=True/False. Also, move to an issue.

DanHickstein:
Great! Thanks for this PR @ldes89150!  -- Sorry to have taken so long, there were lots of other changes going on, and so we moved slowly. I promise that we'll be more efficient in the future! :)

@stggh, thanks for pointing out the issues and the great discussion!

PR #160, opened 2016-04-14
Version bump to 0.7.3 to make new release

DanHickstein:


stggh:
Good idea. 

I am in the process of adding `get/put_image_halves()` (needs another one of your clever function names) to `abel/tools/symmetry.py`, which will simplify image processing if `symmetry_axis=1` symmetry is not invoked. This can be included in the later 0.7.4.

DanHickstein:
That sounds good. But doesn't `_image_quadrants()` work fine for any situation? I worry that adding more functions will make the code more confusing, but you're more closely acquainted with `symmetry.py` than I am.

stggh:
In principle, you may be right. Quadrant processing is sufficient, but it adds unnecessary image processing if the symmetry is not exploited.

In order to split an image into 2, for an odd-size image an extra row/column must be added. The row operation is purely to obtain the quadrants, and not necessary for the Abel transform (unless `axis=1` symmetry is present).

My argument, is that if we were writing PyAbel from scratch, our natural processing block would be the 1/2-image.

My motive is to track down why the full-image transform appears to have less resolution that a single quadrant transform. This may simply be image-center related.

Ok, I agree, unless the 1/2-image stuff proves useful, I will keep it hidden.

DanHickstein:
Oh, the full image transform looks worse than a single quadrant? More centering conundrums I suppose!

stggh:
Only on my nitpicking (spectroscopic) scale. I need to run more tests. This is not a big deal, but it would be nice to see if there is a problem to be resolved.

PR #161, opened 2016-04-14
set up travis CI to automatically deploy releases to PyPi

DanHickstein:
(Hopefully this works…)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,36 +1,36 @@
 language: python
 sudo: false
 matrix:
-    include:
-    - python: "3.5"
-      env: DEPS="numpy scipy nose cython"
-    - python: "3.4"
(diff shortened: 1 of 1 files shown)


stggh:
Any simply summary as to how this clever stuff is done?

DanHickstein:
Apparently it is as simple as just including a few lines of text in the `.travis.yml` file: https://docs.travis-ci.com/user/deployment/pypi

```
deploy:
  provider: pypi
  user: ...
  password: ...
  on:
    tags: true
```

Currently, I have been sending the releases to PyPi from the command line, but this is annoying and I think that only I can do it. Now, anyone with write access to the repo can make a release and, in theory, it should end up on PyPi. Going to test it now....

PR #164, opened 2016-04-15
Fix travis deploy

DanHickstein:
Stackoverflow told me how to fix Travis to automatically release to PyPi.

http://stackoverflow.com/questions/36635650/travisci-deploy-cannot-find-setup-py/36642901#36642901

Basically, I think that I need to remove the line: 
```
cd ~/
```
I'm not sure why this was needed...

Hopefully this time it works!

Also, I fixed the typos in the readme that @rth found.


DanHickstein:
Hmmm, that created another problem. Basically, the test could not be run under Python3 because we were already in the PyAbel directory and, therefore, "import abel.tests" is importing locally instead of the abel package. I tried changing the command from 
```
python  -c "import abel.tests; abel.tests.run_cli(coverage=True)"
```

to 
```
nosetests -vv abel/tests
```

We will see if that works...

DanHickstein:
Okay, that seems to work. I inadvertently omitted the coverage information, so I will include that with
```
nosetests -vv --with-coverage --cover-package=abel abel/tests
```

PR #171, opened 2016-04-19
CHANGELOG -- abel.transform() -> abel.Transform()

furii:


DanHickstein:
Looks good! Thanks again!

(For future reference, you can just add additional commits to the same branch that you made the PR from, and the PR will be updated. No problem to have a second PR, but for related changes, typically we just modify the existing PR with additional commits.)

PR #172, opened 2016-04-19
Change the default of angular_integration() from Jacobian=False to Jacobian=Ture in tools/vmi.py

PhantomYuan:
Because in VMI imaging processing, we always use the correct Jacobian r*sin(theta) term when doing Abel inversion.

Yuan

### DIFF
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -12,13 +12,15 @@ from scipy.ndimage.interpolation import shift
 from scipy.optimize import curve_fit
 
 
-def angular_integration(IM, origin=None, Jacobian=False, dr=1, dt=None,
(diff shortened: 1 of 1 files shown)


DanHickstein:
Looks good to me. (I will let @stggh press the merge button on this one.)

For the record, the complete Jacobian for integration in spherical actually `r^2 * sin(theta)`. The `angular_integration` function itself applies only the `r * sin(theta)` part, while the `reproject_image_into_polar` applies the extra factor of `r`. 

Numerical integration in spherical/polar coordinates gets confusing. I think that the unwritten experimentalist rule is: when in doubt, multiply or divide by `r` to achieve agreement with reality! :)

PR #173, opened 2016-04-22
If `average` is `True`, don't use jacobian.

ldes89150:
fix a small bug.

### DIFF
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -12,8 +12,7 @@ from scipy.ndimage.interpolation import shift
 from scipy.optimize import curve_fit
 
 
-def angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None,
-                        average=False):
+def angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None):
     """ 
     Angular integration of the image.
 
@@ -48,9 +47,6 @@ def angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None,
         is equal to the height of the image (which should typically ensure
         good sampling.)
 
-    average: bool
-        If ``average=True``, return the averaged radial intensity instead.
-
     Returns
     -------
     r : 1D np.array
@@ -61,9 +57,6 @@ def angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None,
 
      """
 
-    if average:
-        Jacobian=False
-
     polarIM, R, T = reproject_image_into_polar(
         IM, origin, Jacobian=Jacobian, dr=dr, dt=dt)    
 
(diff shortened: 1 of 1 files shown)


DanHickstein:
So, it seems that all the `average` keyword  does is to set the `Jacobian=False` and then divide the results by 2*pi. Is it perhaps simpler to get rid of the `average` keyword and just include something in the docstring to the extent of: 

`Note: if an average radial value is desired (rather than the integration), then simply set `Jacobian=False` and divide the resulting array by 2*pi`

ldes89150:
hi @DanHickstein 

If so, I would suggest to add a new function:
```python
def averaged_radial_intensity(IM, origin, dr=1, dt=None):
    R, speeds = angular_integration(IM, origin, Jacobian=False, dr=dr, dt=dt)
    speeds /= 2*np.pi
    return R, speeds
```

DanHickstein:
I think that this solution might be the most clear. What does @stggh think?

stggh:
I am not in favour of adding a new function (with docs) to the API which simply involves dividing the result by <math>2 &pi;</math>.  I prefer @DanHickstein's solution to remove the `average` keyword and add a comment to the docs.

ldes89150:
@stggh Thanks for your comment. However, I think providing such a function can be quite convenient for some people and more friendly to people who just being to use this package. Also, people who do cold atom experiment may use inverse abel transform to recover the local density distribution of the cold atom cloud. In this case, what we need is the radial intensity only.

stggh:
Ok, I am not that fussed by the issue.

DanHickstein:
I agree with @stggh that we should minimize the amount of code/docs that we need to maintain. But, if @ldes89150 thinks that the function will be useful, then lets go ahead and include it. 

But what is the best format for the docstring for this new function? The docstring that @ldes89150 included is fine, but it basically duplicates the docstring for `angular_integration`, meaning that if we make a change to angular integration, then we need to also change this docstring in an identical way. I would think that perhaps it's just better to have a very short docstring that describes the function and links to the `angular_integration` docs.

Also, it might be simplest to pass all `kwargs` to `angular_integration`.
So, the function might look something like this:

```python
def average_radial_intensity(IM, **kwargs):
	"""
	Calculate the average radial intensity of the image, averaged over all angles.
    This differs form :func:`abel.tools.vmi.angular_integration` only in that it returns
    the average intensity, and not the integrated intensity of a 3D image. 
    It is equavalent to calling :func:`abel.tools.vmi.angular_integration` with 
    `Jacobian=True` and then dividing the result by 2*pi.


	Parameters
	----------
	IM : 2D np.array
	 The data image.

    kwargs : 
      additional keyword arguments to be passed to :func:`abel.tools.vmi.angular_integration`

	Returns
	-------
	r : 1D np.array
	  radial coordinates

	intensity : 1D np.array
  	  one-dimentional intensity profile as a function of the radial coordinate.

	"""
	R, intensity = angular_integration(IM, Jacobian=False, **kwargs)
	intensity /= 2*np.pi
	return R, intensity
```

Note: that code was not checked - it may have some bugs, and I'm not quite sure if I included those links properly.

But anyway, the goal of the above code it to make it so that if small changes are made to `angular_integration`, then this function will not need to change.

DanHickstein:
Sorry to be slow about this @ldes89150. If you want, we can just merge what you have now and address any reformatting of the docsstrings later...

ldes89150:
Thanks for the advice, Dan.  I'll modify it as what you suggest. It's a
good idea.

DanHickstein:
Ah, you made the change. I think this is good to merge now, right?

ldes89150:
@DanHickstein I think so.

DanHickstein:
Merged! Thanks!

stggh:
@DanHickstein for some reason this merge has caused major conflicts with my fork when I attempt to rebase.  Your own PR#175 rebased fine.  For the few lines code in the PR#173, I wonder what has caused this problem. I will see what I can edit to achieve the rebase.

DanHickstein:
Oh, this is not good! So many Git troubles...

Have you changed lots of files? One option would be to create a new branch from the master, and re-implement your changes (hopefully just pasting a few .py files).

Issue #176, opened 2016-05-07
Make a function to extract anisotropy versus radial coordinate


DanHickstein:
I got an e-mail from a PyAbel user a few days ago: he was accustomed to using the BASEX software and he wanted to calculate the anisotropy parameter as a function of the radial coordinate. Currently, our `abel.tools.vmi.anisotropy_parameters()` function only calculates the anisotropy parameter for one specific curve. While it is possible to pass several radial ranges to `abel.tools.vmi.radial_integration()`, it's not possible to get the "anisotropy-versus-radius curve" without a little bit of coding. 

So, the question is, should this just be a function in PyAbel? It's only a few lines of code, but it might save interested users some time.

I provided the user following code that achieves this, and it could be wrapped up into a function pretty easily:

```python
# We start with AIM as the transformed image.
# first we set up the ranges for radial integration

r = np.arange(0,300,2) # the step here should be changed to give good signal-to-noise
r_range = zip(r[:-1], r[1:])

# map to intensity vs theta for each radial range
intensities, theta = abel.tools.vmi.radial_integration(AIM, radial_ranges=r_range)


# The following just loops over the intensity values and calculated the anisotropy
# There is probably a way more Pythonic way to do this. Probably with the "map" function...
anisotropies = np.zeros(np.shape(intensities))
errors       = np.zeros(np.shape(intensities))

print("radial-range      anisotropy parameter (beta)")
for i, (rr, intensity) in enumerate( list(zip(r_range, intensities)) ):

    (beta, error_beta), (amplitude, error_amplitude) = abel.tools.vmi.anisotropy_parameter(theta, intensity)
    
    anisotropies[i] = beta
    errors[i] = error_beta
```

stggh:
@DanHickstein this is very good, and your code `r_range = zip(r[:-1], r[1:])` is an education.

[...]

PR #178, opened 2016-07-05
anisotropy versus radial coordinate

stggh:
From issue #176. This is really @DanHickstein's baby, that I somehow hijacked.  It simply allows the determination of the anisotropy parameter for a radial grid, determined from @DanHickstein clever code [example](https://github.com/PyAbel/PyAbel/issues/176#issue-153615099):
```python
r = np.arange(0,300,2)
r_range = list(zip(r[:-1], r[1:]))
```
In this PR I have:
 1. Created new function `abel.tools.vmi.anisotropy()` to provide a single call to obtain the anisotropy parameter. Previously, 2 steps were required, one to evaluate intensity vs angle (for a given radial range) `abel.tools.vmi.radial_integration()`, the second to evaluate the anisotropy parameter from a fit to the intensity variation `abel.tools.vmi.anisotropy_parameter()`. See `examples/example_O2_PES_PAD.py`

 2. Modified `abel.tools.vmi.anisotropy_parameter()` to accept an array of intensity vs thetas
    previously this required a for loop to process the return arrays of `abel.tools.vmi.angular_integration()`. Again, see `examples/example_O2_PES_PAD.py`

 3. (Yet) another example  'example_PAD.py` demonstrating the new function calls
  Individual peaks (red points, with error bars in plot below):
     ```python
     r_range = [(93, 111), (145, 162), (255, 280), (330, 350), (350, 370),
           (370, 390), (390, 410), (410, 430)]

     # anisotropy parameter for each tuple r_range
      Beta, Amp, R = abel.tools.vmi.anisotropy(AIM, r_range)
    ```
    and the whole radial grid (to emulate `basex`/`linbasex`), (green line in plot below):
    ```python
    # OR  anisotropy parameter for ranges (0, 20), (20, 40) ...
    Br, Ar, Rr = abel.tools.vmi.anisotropy(AIM, 20)
    ```

example_pad

I am not 100% happy with the whole radial-grid output (green line in plot), I would prefer it to look like the output from `basex`/`linbasex`.  I guess  I should plot the anisotropy parameter outputs from `basex`/`linbasex` for comparison. 

Open for comment ...

### DIFF
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -248,8 +248,8 @@ def anisotropy(AIM, radial_ranges=None):
     Covenience function providing a single call to 'radial_integration' and
     'anisotropy_parameter'
 
-    Note: in general specifying a radial range tuple that encompasses a spectral
-    feature that has good intensity, will provide a more accurate
+    Note: in general specifying a radial range tuple that encompasses a
+    spectral feature that has good intensity, will provide a more accurate
     anisotropy parameter
 
     Parameters
@@ -279,4 +279,5 @@ def anisotropy(AIM, radial_ranges=None):
     IvsTheta, theta, Rbar = radial_integration(AIM, radial_ranges)
     Beta, Amp = anisotropy_parameter(theta, IvsTheta)
 
-    return Beta, Amp, Rbar
+    # tranpose gives beta vs r, ebeta vs r etc
+    return np.transpose(Beta), np.transpose(Amp), Rbar
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -47,8 +47,8 @@ def angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None):
         is equal to the height of the image (which should typically ensure
         good sampling.)
 
-    Returns
-    -------
+    [eturns
+    a#------
     r : 1D numpy.array
          radial coordinates
 
@@ -213,9 +213,50 @@ def anisotropy_parameter(theta, intensity, theta_ranges=None):
         intensity = intensity[subtheta]
 
     # fit angular intensity distribution
-    popt, pcov = curve_fit(PAD, theta, intensity)
-
-    beta, amplitude = popt
-    error_beta, error_amplitude = np.sqrt(np.diag(pcov))
+    try:
+        popt, pcov = curve_fit(PAD, theta, intensity)
+        beta, amplitude = popt
+        error_beta, error_amplitude = np.sqrt(np.diag(pcov))
+    except:
+        beta, error_beta = nan, nan
+        amplitude, error_amplitude = nan, nan
 
     return (beta, error_beta), (amplitude, error_amplitude)
+
+
+def anisotropy(AIM, radial_step):
+    """
+    Combines `radial_intensity()` and `anisotropy_parameter()` to return the
+    anisotropy parameter for each of the radial ranges.
+
+    Parameters
+    ----------
+    AIM : numpy 2d float array
+        inverse Abel transformed image slice
+    radial_step: int
+        radial integration over 'radial_step' radii
+
+    Returns
+    -------
+    beta : array of tuples
+        (beta0, error_beta_fit0), (beta1, error_beta_fit1), ... 
+        corresponding to the radial ranges
+    amplitude : array of tuples
+        (amp0, error_amp_fit0), (amp1, error_amp_fit1), ... 
+        corresponding to the radial ranges
+
+    """
+    rows, cols = AIM.shape
+    radial_grid = np.arange(0, cols//2)
+    # clever code from @DanHickstein
+    radial_ranges = zip(radial_grid[:-radial_step], radial_grid[radial_step:])
+
+    IvsTheta, theta, Rbar = radial_integration(AIM, radial_ranges)
+    Beta = []
+    Amp = []
+    for i, angle in enumerate(theta):
+        (beta, ebeta), (amp, eamp) = anisotropy_parameter(angle, IvsTheta[i])
+        Beta.append((beta, ebeta))
+        Amp.append((amp, eamp))
+
+    return Beta, Amp
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -173,16 +173,9 @@ def radial_integration(IM, radial_ranges=None):
         Intensity_vs_theta.append(intensity_vs_theta_at_R)
         radial_midpt.append(np.mean(rr))
 
-        # evaluate anisotropy parameter from intensity_vs_theta_at_R fit to
-        #        amp*(1 + beta P_2(cos theta)) 
-        # returns (beta_i, fit_error_i) for each radial range
         beta, amp = anisotropy_parameter(theta, intensity_vs_theta_at_R)
         Beta.append(beta)
         Amp.append(amp)
-     
-    Beta = np.transpose(Beta)
-    Amp = np.transpose(Amp)
-    Intensity_vs_theta = np.array(Intensity_vs_theta)
 
     return Beta, Amp, radial_midpt, Intensity_vs_theta, theta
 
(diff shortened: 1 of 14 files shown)


stggh:
Hmm, I am not sure why Travis-CI for Python3.3 fails, all others (including the AppVeyor build) pass.

abel/tools/vmi.py
```diff
@@ -13,13 +13,13 @@
...
-    """ 
-    Angular integration of the image.
+    """Angular integration of the image.
 
     Returns the one-dimentional intensity profile as a function of the
     radial coordinate.
-    
-    Note: the use of Jacobian=True applies the correct Jacobian for the integration of a 3D object in spherical coordinates.
+
+    Note: the use of Jacobian=True applies the correct Jacobian for the
```
DanHickstein: maybe change "correct Jacobian" to "correct Jacobian (r*sin(theta))"

DanHickstein: Oh, this is mentioned in the Parameters section. Disregard.

DanHickstein:
Thanks for adding this @stggh! I didn't look to carefully at the code yet, but the general idea seems great.  

The disagreement between the red dots and green curve is just a result of different bin boundaries being used, correct? And we would expect them to be on top of each other given the same bins?

What are you unhappy about regarding the output? BASEX.exe provided one anisotropy point per output pixel, so perhaps you can emulate the old (crappy) behavior by just setting your bin step to 1. In this case, the output is going to be really noisy, so when you plot it, you'll likely need to set some intensity threshold and only plot the anisotropy where you have sufficient signal (set all of the below threshold anisotropy points to np.nan).

stggh:
Hmm, a small uniform grid fails dismally, due to the lack of intensity in most regions of the fit:
anisotropy

In reality, the fixed spaced grid determination of the anisotropy parameter, from a least-squares fit to the intensity variation, has little utility, due to the large error bar associated with fits for low intensity spectral radial regions. It lacks the advantages inherent in the `basex`/`linbasex` approaches. However, this PR does provide a nice interface to extract anisotropy parameters from any image:

```python
Beta, Amp, Rmid = abel.tools.vmi.anisotropy(AIM, radial_ranges=[(rmin0, rmax0), (rmin1, rmax1), ...])
```
which I claim should provide the most accurate anisotropy parameter determination.

I will tidy up `example_PAD.py` which might better be adapted from `example_linbasex_hansenlaw.py` as in the figure above.

DanHickstein:
well, the small uniform grid is almost doing the right thing in the regions where there is signal. But linbasex (and the blue dots) really does a nice job.

stggh:
I updated `examples/example_PAD.py`

example_pad

`linbasex` does a pretty good job determining the anisotropy parameter, and it is nice that it is a by-product of the method.

Q: @DanHickstein, @DhrubajyotiDas  is it possible for `basex` to return the anisotropy parameter variation in the same way as `linbasex`?

I will (hope to!) add something to the docs to discuss the anisotropy parameter evaluation.

For this PR, specifically the code in `abel/tools/vmi.py` it may be better to simply have `radial_integration()` return the anisotropy parameter fits. ie. remove the pseudo `anisotropy()` function and add the line `Beta, Amp = anisotropy_parameter(theta, IvsTheta)` to `radial_integration()`.  I will do this later tonight or early tomorrow.

DanHickstein:
This looks really nice @stggh!

Using basex, it should be possible to analytically calculate the anisotropy parameter. However, I'm not sure if the anisotropy parameter would actually be much nicer than what you get from fitting functions to the image as your function can do for any transform method. The reason is that Basex is using a basis set that corresponds to a pixel-by-pixel transform. In contrast, the linbasex method is fitting to functions that are quite smooth in the angular dimension. It's the ideal case for extracting the anisotropy parameters, and I doubt that any method that doesn't fit to polar basis functions like these could really compete for producing nice anisotropy distributions. It's probably worth noting this observation in the PyAbel docs somewhere.

stggh:
The image anisotropy parameter determination is now rationalized to the single function `abel.tools.vmi.radial_integration()`. Updated `examples/example_PAD.py` to match. 

```python
Beta, Amp, Rmid, Ivstheta, theta = abel.tools.vmi.radial_integration(HIM.transform, r_range)
```

Added `doc/anisotropy_parameter.rst`.  Note:

 1.  this is photoelectron specific.
 2.  my description of `linbasex` method is vague. 


For consideration:
    
  There is a namespace collision between `linbasex` use of `Beta` and the anisotropy parameter. Currently, `linbasex` returns `Beta[0]` = speed, and `Beta[1]` = anisotropy parameter. I think, strictly speaking, this data "should" be in the array as `Beta[2]` ?  @gerber-t  sorry for the lack of rigour. I will fix this issue later.

stggh:
This PR extends the evaluation of the anisotropy parameter &beta; for specific radial regions (usually those that cover strong spectral features) to emulate the full radial determination provided by `linbasex` and `basex`. It achieves this by providing for an integer `step` parameter, which to define a radial grid `(0, step), (step, 2*step), ... , (rmax-step*2, rmax-step)`, using the [clever](https://github.com/PyAbel/PyAbel/issues/176#issue-153615099)  python code of @DanHickstein.

The anisotropy parameter for each radial range is determined by first extracting the (integrated) intensity vs angle, and then fitting  `amp*(1 + beta * P2(cos(theta))` to the intensity, with `beta` and
`amp` as parameters, returned together with their parameter fit errors.

As for all the anisotropy parameter determinations, the most accurate values are determined where the image intensity is strong. Consequently, it turns out, that a uniform `step` grid is of little utility for this method. A more accurate determination of the anisotropy parameter is obtained by limiting the radial (integration) range to spectral features of good intensity.

The `linbasex` method fits Legendre polynomials to each radial (Newton sphere) and consequently directly provides the radial distribution of the anisotropy parameter directly.

As illustrated in the figure [above](https://github.com/PyAbel/PyAbel/pull/178#issuecomment-231002656) the two methods provide complementary information.

Main changes: 

 (1) `radial_ranges` now accepts an integer `step` parameter to define a uniform radial grid over which to extract `I  vs  theta` and determine the anisotropy parameter from a least-squares fit.

 (2) `abel.tools.vmi.radial_integration()` has been rationalized to now return the anisotropy parameter (and parameter fit error), instead of only intensity vs angle for each radial range (which then required a separate call to `abel.tools.vmi.anisotropy_parameter()`).

 (3) Updated examples that involve the anisotropy parameter determination: `example_GUI.py, example_linbasex_hansenlaw.py, example_O2_PES_PAD.py, example_simple_GUI.py`, with a new example, `example_PAD.py`, illustrating the two `PyAbel` methods to determine anisotropy parameters (image above).
   Made improvements to `example_GUI.py` to include `linbasex` transform, and to plot the radial variation of the anisotropy parameter, available for this case.

 (4) New documentation `anisotropy.rst`.

 (5) Added unit test for anisotropy parameter, in `abel/testing/test_tools.py`,

stggh:
@PyAbel/developers, I think that this PR is ready for merging. Overall, it is only a small change to PyAbel.

DanHickstein:
Thanks for doing this @stggh! Sorry to be absent from the conversation - I've been quite busy for the past few days putting together a conference talk. I'll try to take a look at this in the next few days. Otherwise we can merge as is.

stggh:
@DanHickstein no worries.  Take your time. My main constraint is that I will be travelling in 3 weeks time, for a period of 3 weeks.

stggh:
@DanHickstein I would like to merge this PR to allow a little time to work more on aspects of the `linebasex` method processing, before I travel.

 I have been playing with the anisotropy determination and it provides more of what a user might expect when `Beta[2]` (or the attribute `.Beta[1]`) is plotted not normalized. e.g.
beta1

Note, this is only a stylistic change (does not affect the anisotropy parameter), but the anisotropy output may be more familiar to uses of other inverse Abel programs - I think?

DanHickstein:
Yup, go ahead and merge! Thanks @stggh!

Issue #183, opened 2017-01-05
Cython extension not building on Windows


DanHickstein:
In the process of getting PyAbel on conda-forge, I noticed that the Cython extensions were failing on AppVeyor. It looks like some of our attempts to supress the warnings from the C-compiler may be causing problems: https://ci.appveyor.com/project/DanHickstein/pyabel/build/1.0.493

```
C:\Miniconda3\envs\abel-env\Scripts\gcc.bat -mdll -O -Wall -IC:\Miniconda3\envs\abel-env\lib\site-packages\numpy\core\include -IC:\Miniconda3\envs\abel-env\include -IC:\Miniconda3\envs\abel-env\include -c abel\lib\direct.c -o build\temp.win32-3.4\Release\abel\lib\direct.o -Wunused-function -Wunneeded-internal-declaration -Wno-#warnings
gcc.exe: error: unrecognized command line option '-Wunneeded-internal-declaration'
**************************************************
WARNING: Cython extensions failed to build (used in abel.direct).
Typical reasons for this problem are:
  - a C compiler is not installed or not found
  - issues using mingw compiler on Windows 64bit (experimental support for now)
This only means that the abel.direct C implementation will not be available.
```

Perhaps we should simple remove these flags and accept that fact that the C-compiler will spam us with lots of warnings during install?

rth:
@DanHickstein Do you think it's related to https://github.com/PyAbel/PyAbel/issues/125 ? if so I can look into it this tomorrow. 

I'm +1 on removing those custom warning related flags in [`setup.py`](https://github.com/PyAbel/PyAbel/blob/master/setup.py#L55) (as they are compiler specific): if we want to remove them from the install stdout, a better way IMO would be fix the corresponding code in Cuython rather than hiding the warning.

DanHickstein:
Yeah, it's likely that we broke things by trying to make the install process less verbose. If you have time to fix the Cython, then this would be great! Otherwise, we should probably remove all of the warning suppressions from the compiler arguments, since we are seeing that they cause problems for different compilers.

DanHickstein:
In the process of submitting PR https://github.com/PyAbel/PyAbel/pull/184, I noticed that PyAbel fails the AppVeyor build, with an error relating to the Cython code: https://ci.appveyor.com/project/DanHickstein/pyabel/build/1.0.495

[...]

Issue #125, opened 2016-03-02
Appveyor 64-bit build not working


DanHickstein:
This seems like something related to the Cython direct Abel implementation. I turned off the 64-bit Appveyor build because I was tired of seeing the red X next to all of the PRs. 

If @rth has time, he should look into this.

rth:
Sorry everybody, just saw this now. 
@DhrubajyotiDas  Does your PR above fixes the problem (by disabling the 64 bit version which is problematic)?

rth:
Should be fixed in PR #188 .

PR #185, opened 2017-01-06
FIX Cython extension not building on Windows

rth:
This PR fixes issue #183 (and possibly #125, if that was still an issue) with the compilation on Windows. All windows builds are tested again, which takes some time with appveyor but is probably safer.

The solution was to remove the `libpython` (mingw compiler) from the list of installed packages so that the already present Visual Studio compiler gets used. 

Also removing all flags to hide warnings (as those might actually be useful),  reverting to the default flags.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,15 +51,10 @@ else:
 
 
 if sys.platform != 'win32':
-    compile_args = dict( extra_compile_args=['-O2', '-march=native',
-                                             '-Wno-unused-function',
-                                             '-Wno-unneeded-internal-declaration',
-                                             '-Wno-#warnings'],
+    compile_args = dict( extra_compile_args=['-O2', '-march=native'],
                              extra_link_args=['-O2', '-march=native'])
 else:
-    compile_args = dict( extra_compile_args=['-Wunused-function',
-                                              '-Wunneeded-internal-declaration',
-                                              '-Wno-#warnings'])
+    compile_args = dict( extra_compile_args=[])
 
 # Optional compilation of Cython modules adapted from
 # https://github.com/bsmurphy/PyKrige which was itself adapted from a StackOverflow post
--- a/setup.py
+++ b/setup.py
@@ -53,8 +53,10 @@ else:
 if sys.platform != 'win32':
     compile_args = dict( extra_compile_args=['-O2', '-march=native'],
                              extra_link_args=['-O2', '-march=native'])
+    libraries = ["m"]
 else:
     compile_args = dict( extra_compile_args=[])
+    libraries = []
 
 # Optional compilation of Cython modules adapted from
 # https://github.com/bsmurphy/PyKrige which was itself adapted from a StackOverflow post
@@ -74,16 +76,23 @@ class TryBuildExt(build_ext):
                   "  - issues using mingw compiler on Windows 64bit (experimental support for now)\n"
                   "This only means that the abel.direct C implementation will not be available.\n")
             print("**************************************************")
-            # continue the install
-            pass
+            if os.environ.get('CI'):
+                # running on Travis CI or Appveyor CI
+                if sys.platform == 'win32' and sys.version_info < (3, 0):
+                    pass # Cython extensions are not built on Appveyor (Win) for PY2.7
+                         # see PR #185
+                else:
+                    raise
+            else:
+                # regular install, Cython extensions won't be compiled
+                pass
         except:
             raise
-
 
 ext_modules=[
     Extension("abel.lib.direct",
              [os.path.join("abel","lib","direct.pyx")],
-             libraries=["m"],
+             libraries=libraries,
              **compile_args),
     ]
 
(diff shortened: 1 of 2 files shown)


DanHickstein:
Thanks for doing this @rth!

While the AppVeyor checks now pass, the Cython extensions are not actually building on AppVeyor.

rth:
@DanHickstein My mistake, thanks for spotting this! I'll fix it, it has something to do with `libm` being in `msvcr` when using MSVC compiler...

rth:
The situation on this PR is the following. Cython extensions build fine on Appveyor for PY3 but not for PY2. The issue is the libm (that is used for `sqrt` and `arcosh` function calls) is that should be included in the runtime enviroment of Microsoft visual studio compiler. This does work as expected with MSVC 2015 (PY3.5) but for some reason non on MSVC 2008. 

I agree with @stggh  on this one, it would be disproportionate to spend more time debugging this, given the little value this actually brings compared to other features in PyAbel (at least for me). 

I added some code to make sure that failed cython exceptions do raise an error in Travis / Appveyor, then explicitly ignored this error for PY2 on Windows. For users, it's still can silently fail and will just print a warning. 

If this is a problem for conda-forge we could reopen this, or just say that PyAbel includes this direct C implementation only on Linux and PY3 Win. Another argument for not getting rid of cython altogether is that once we have this workflow, it's relatively easy to add some existing C implementations or potentially optimize existing algorithms with Cython without additional effort.

setup.py
```diff
@@ -51,15 +51,12 @@
...
-                                             '-Wno-#warnings'],
+    compile_args = dict( extra_compile_args=['-O2', '-march=native'],
                              extra_link_args=['-O2', '-march=native'])
+    libraries = ["m"]
 else:
-    compile_args = dict( extra_compile_args=['-Wunused-function',
-                                              '-Wunneeded-internal-declaration',
-                                              '-Wno-#warnings'])
+    compile_args = dict( extra_compile_args=[])
+    libraries = []
```
DanHickstein: What happens if you provide the "m" library on win32?

DanHickstein:
I agree that we shouldn't spend too much time on this, since it will appeal to only a small subset of users. Still, see my comment above: are we just not giving the win32 install access to the m library?

setup.py
rth: There is no such library with the MSVC compiler (adding it throws an error), the corresponding code is in the `msvcrt` (adding it doesn't help either), the [runtime library](https://msdn.microsoft.com/en-us/library/2kzt1wy3.aspx) which should be linked with the `\MD` flag (on by default). This indeed works with MSVC 2015 (PY3.5) but not MSVC 2008 (PY2.7) for some reason. Not sure I want to learn more details about the MSVC compiler regarding this )

DanHickstein: According to SO, MSVC 2008 doesn't include the ``acosh`` function: http://stackoverflow.com/questions/15539116/atanh-arc-hyperbolic-tangent-function-missing-in-ms-visual-c

Should we just define the function ourselves using `sqrt`?

rth: I had this suspicion (makes sens now, thanks for checking it), and had tried to reimplement the `acosh` function. The problem was that,
  - it worked fine on linux https://travis-ci.org/rth/PyAbel/jobs/189767399
  - but the results were not correct on Windows for some reason https://ci.appveyor.com/project/rth/pyabel/build/1.0.71/job/5hgtn6iic43qukax 

besides I imagine the built-in `acosh` might be better optimized and faster which would be important here.

DanHickstein: Oh, that is weird. It means that ``acosh`` is not equal to ``ln( 1-sqrt(x**2 -1) )`` on MSVC. That's pretty strange. 

I agree, we should just merge this. directC doesn't work with the MSVC2008, which is fine, since hopefully not many people should be in this situation going forward.

rth: It is strange, but well it's not critical and Python 3 is working, so it's just another incentive to migrate to PY3 on Windows )  Feel free to merge this PR then...

DanHickstein:
Cython extensions are now building on MSVC 2015 (Win+Py3 on Appveyor), but now on MSVC 2008 (Win+Py2 on Appveyor).

Partial victory declared!! :)

rth:
> Cython extensions are now building on MSVC 2015 (Win+Py3 on Appveyor), but now on MSVC 2008 (Win+Py2 on Appveyor).
> 
> Partial victory declared!! :)

A strategic retreat would be more accurate :)

stggh:
Great work @rth and @DanHickstein, BUT, what happened between:

> Oh, that is weird. It means that acosh is not equal to ln( 1-sqrt(x**2 -1) ) on MSVC.

and

> Cython extensions are now building on MSVC 2015 (Win+Py3 on Appveyor), but now on MSVC 2008 (Win+Py2 on Appveyor).

Oh, you mean "**not**"

Given that Python3 is developing fast, and Conda provides a simple OS-independent Python3 installation, perhaps Python2 support for `PyAbel` will be next ;-)

DanHickstein:
Yeah, I meant "not". :)

PR #188, opened 2017-01-10
Acosh fix

DanHickstein:
Including definition of acosh to `direct.pyx` to allow Cython extensions to build, even with compilers that don't have `acosh`. As discussed in https://github.com/PyAbel/PyAbel/issues/183


rth:
Thanks for fixing it @DanHickstein !

Issue #186, opened 2017-01-10
image circularization, distortion correction


stggh:
**Python methods to circularize a velocity-mapped image**
This issue is a discussion on how to best implement image circularization (distortion correction) in `PyAbel`.

_Background_ 
For velocity-map imaging experiments the charged particles emanate as Newton spheres that have spherical symmetry, and thus the 2D projection should be perfectly circular.  Experimental velocity-map images may have distortion in the circular charged particle energy structure, due to, for example, stray magnetic fields, or optical distortion of the camera lens that images the particle detector. The effect of distortion is to reduce the radial (or velocity or kinetic energy) resolution, since a particular energy peak will "walk" in radial position, depending on the particular angular position on the detector. Imposing a physical circular distribution of particles, improves the kinetic energy resolution, at the expense of uncertainly in the absolution kinetic-energy position of the transition. A nice description of distortion, and a method of correction is given in the thesis of [Mikhail Ryazanov]( 
http://digitallibrary.usc.edu/cdm/ref/collection/p15799coll3/id/112619) from page 152. Tks, @DanHickstein for the reference source.

_Approach_
The ANU approach to image circularization, is based on a idea of Jason Gascooke (Flinders University, SA Australia) to compare the radial positions of features in angular slice profiles. i.e. follow the radial position of a peak as a function of angle. A linear correction is applied to the radial grid to align the peak at each angle. 

Using the example data `O2-ANU1024.txt.bz2` (before and after circularization) we may see how idea this looks:
![circ](https://cloud.githubusercontent.com/assets/10932229/21787764/1a446288-d71f-11e6-9fa7-89db1cea1b85.png)

The velocity-map image has a slight ellipticity (due to the optical camera lens), so that the vertical radial scale is compressed relative to the horizontal scale. This is seen in the slice profiles (bottom left figure) where the photodetachment peaks do not align for the horizontal an vertical slices. It is also evident in the polar coordinate image as a "wavy" vertical line.  A circularized image straightens the polar image lines, aligning the peaks with angle (right bottom figures).  [_Hmm, this example does not clearly illustrate the situation as I had hoped. Take my word for it!_]

_Python to the rescue_

- array slicing to trivially extract slice profiles (as above)
- `PyAbel`'s `abel.tools.polar.reproject_image_into_polar()` to easily generate the polar coordinate image
- [`peakutils`](http://pythonhosted.org/PeakUtils/) detection of peaks in 1-D data
[...]

Issue #192, opened 2017-02-08
Conventions in README?


DanHickstein:
In a discussion with @stggh, I realized that I was frequently getting confused if `IM[0]` referred to the "top" or "bottom" of an image in PyAbel. So, I was thinking that it might be nice if we listed some of the "PyAbel conventions" in the README file, and if you all had any suggestions for what might be listed there. So far I can think of:

- **Image orientation:** PyAbel adopts the "television" convention, where `IM[0,0]` refers to the **upper** left corner of the image. (This means that `plt.imshow(IM)` should display the image in the proper orientation, without the need to use the `origin='lower'` keyword.) As an example, the x,y-grid for a 5x5 image can be generated using:
    ```python
    x = np.linspace(-2,2,5)
    X,Y = np.meshgrid(x, -x) # notice the minus sign in front of the y-coordinate
    ```

- **Angle:** All angles in PyAbel are measured in radians. When an absolute angle is defined, zero-angle corresponds to the upwards, vertical direction. Positive values are on the right side, and negative values on the left side. The range of angles is from -Pi to +Pi. The polar grid for a 5x5 image can be generated (following the code above) using:
    ```python
    R = np.sqrt(X**2 + Y**2)
    THETA = np.arctan2(X, Y)
    ```
    where the usual "`(Y, X)`" convention of `arctan2` has been reversed in order to place zero-angle in the vertical direction. Consequently, to convert the angular grid back to the Cartesian grid, we use:
    ```python
    X = R*np.sin(THETA)
    Y = R*np.cos(THETA)
    ```

- **Image center:** Fundamentally, the Abel and inverse-Abel transforms in PyAbel consider the center of the image to be located in the center of a pixel. This means that, for a symmetric image, the image will have a width that is an odd number of pixels. (The center pixel is effectively "shared" between both halves of the image.) In most situations, the center is specified using the `center` keyword in `abel.Transform` (or directly using `abel.center.center_image` to find the true center of your image. This processing step takes care of locating the center of the image in the middle of the central pixel. However, if the individual Abel transforms methods are used directly, care must be taken to supply a properly centered image.

EDIT: Fixed a bunch of minor issues with the above on Feb 22, 2017.

Anything else??

(Alternatively, this information might be too specialized, and we should include it in the CONTRIBUTING.md, file.)

stggh:
[...]

Issue #159, opened 2016-04-13
sub-pixel transform alignment discussion


stggh:
`example_all_O2.py` demonstrates pretty good agreement between the transform methods, but at the pixel level there are positional differences between transform methods, as illustrated by the right figure zoom.

![o2-allx](https://cloud.githubusercontent.com/assets/10932229/14479336/b3a73efe-0160-11e6-941e-84a1ad8d0101.png)  ![o2-all-zoomx](https://cloud.githubusercontent.com/assets/10932229/14479349/cb28eb04-0160-11e6-8fe2-8c2a1658025d.png)

The offsets are sub-pixel, about `1/2`-pixel difference between the `hansenlaw` method and `three_point`.

Here is the situation for the "Ominus" sample image [[code](https://gist.github.com/stggh/c4c44c8f29c377a9f8c1ce450d75088e)]
![o-](https://cloud.githubusercontent.com/assets/10932229/14480032/356375e6-0168-11e6-8513-0d7ea6bbbb11.png)
Neither `hansenlaw` nor Dasch `three_point` align perfectly with the original image, but **NB** these are inverse Abel transforms of the **projected** sample image, the forward `hansenlaw` transform.

To attempt to track down any issues with transform method coding, I have run some 1-d function tests. 2d testing will follow.  The 1d-functions include a Gaussian function, which transforms to the same function  <math>x&sigma;&radic;&pi;</math>, where &sigma; is the Gaussian width at `1/e` height, and "curve A" from the [Hansen and Law paper](http://dx.doi.org/10.1364/JOSAA.2.000510) also listed in @DhrubajyotiDas [ipython notebook](https://gist.github.com/DhrubajyotiDas/f8f531ca2cfa3fee6345), which also has a well defined transform. [[Test code](https://gist.github.com/stggh/2d245a6c2187f4d737e978b26be4e8b5)].
1. 1d-Gaussian
   
   _forward_
   There are only two forward transforms in PyAbel, `direct` and `hansenlaw`, both projections agree well with a scaled Gaussian (left side figures below).  The inset is a zoom of the `0-50` radius peak area. 
   
   _inverse_
   I have included the Dasch `three_point` method, which has excellent agreement with the 1d-Gaussian original function. Both `hansenlaw` and `direct` diverge from the Gaussian near the image center-line (function maximum).

[...]

PR #195, opened 2017-02-10
image circularization, distortion correction

stggh:
See issue #186 
> For velocity-map imaging experiments the charged particles emanate as Newton spheres that have spherical symmetry, and thus the 2D projection should be perfectly circular. Experimental velocity-map images may have distortion in the circular charged particle energy structure, due to, for example, stray magnetic fields, or optical distortion of the camera lens that images the particle detector. The effect of distortion is to reduce the radial (or velocity or kinetic energy) resolution, since a particular energy peak will "walk" in radial position, depending on the particular angular position on the detector. Imposing a physical circular distribution of particles, improves the kinetic energy resolution, at the expense of uncertainly in the absolution kinetic-energy position of the transition.

`abel.tools.circularize.py` with `circularize_image()` function to circularize an image, removing distortion, to improve the resolution of spectral structure. Evaluates a scaling factor to align adjacent angular slice intensity profiles, either via `argmax` following maximum intensity, or `lsq` that radially scales the profiles to align. This radial scaling factor is interpolated in angle to provide the actual image grid coordinates `(Yactual, Xactual)`, that are then remapped back to grid coordinates `(Y, X)`.

`examples/example_circularize_image.py` provided. Unit test `tests/test_circularize_image.py`. Docs `circularize_image.rst`.

Works well, thanks to the great inputs from @DanHickstein:  
o-circ-eg

**Additional changes**
- Bug in the `origin` coordinate [determination](https://github.com/PyAbel/PyAbel/issues/186#issuecomment-278217784) (introduced by @stggh #134 :-( ):
   ```python
    origin = (nx//2 + nx % 2, ny//2 + ny % 2)   # % handles odd size image
    ```
   should be `origin = (nx//2, ny//2)` for odd size images.
   Fixed all occurrences. Predominantly in `tools/center.py`, `tools/polar.py`, and Abel transform `direct.py`. 
   Slight tweaks to some unit tests.

- added directories (empty) `examples/bases`, `examples/output` as file holders for the outputs generated by the examples code. Modified all examples to use `basis_dir='bases'` and `savefig('output/xxx')`. NB added `**kwargs` to `hansenlaw.py`, `direct.py`, `onion_bordas.py` to allow `basis_dir` parameter. Verified all examples. 

- added @DanHickstein's `flower_distort()` to `abel.tools.analytical.py`

- added utility function `abel.tools.vmi.toPES()`to provide conversion of speed profile to photoelectron spectrum, as for the figure above. The tricky part is the intensity scaling.

(Edited to fix typos, correct meaning)


DanHickstein:
@Stggh, thanks for doing this! I think that it's a great addition to PyAbel. I'd like to take a close look at all of the new code - hopefully will have time this weekend.

stggh:
No worries, take your time.

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+from scipy.interpolate import UnivariateSpline
+from scipy.optimize import leastsq
+
+import abel
+
+# see https://github.com/PyAbel/PyAbel/issues/186 for discussion
+
+
+def circularize_image(IM, method="lsq", center=None, radial_range=None,
+                      zoom=1, smooth=0, nslices=32, ref_angle=None,
```
DanHickstein: Have you found the smooth keyword to be useful? To me, it seems like a good idea to pass smooth=0 to the spline to allow the curve to run through all the points. But, I haven't thought about this too much.

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+    ----------
+    IM : numpy 2D array
+
+    method: str
+        method to align slice profiles,
+        "argmax" - compare intensity-profile.argmax() of each radial slice.
+        "lsq" - determine radial scaling factor by minimizing the
+                intensity-prfoile with an adjacent slice intensity-profile.
+
+    center: str, or None
```
DanHickstein: include a cross-reference to the `center_image` function. I think that this is how you would do it, but check that this works in sphinx:

```python
pre-center image using :func:`abel.tools.center.center_image`
```

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+        pre-center image using PyAbel centering methods,
+        "com", "convolution", "gaussian", "image_center", "slice"
+
+    radial_range: tuple, or None
```
DanHickstein: Explain what the units of radial-range are. Degrees? Radians?

Hmmm, the units must be the same as whatever is returned by the `reproject_image_to_polar` function, which also doesn't specify which units are returned. Can you check and also include in the `reproject_image_to_polar` docstring?

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+        limit slice comparison to the radial range (rmin, rmax).
+
+    zoom: float
```
DanHickstein: Interesting... Did you have problems with this? Alternatively, you can just pass `reproject_image_to_polar` a smaller value for `dr` in order to improve this. This might be a case where the default `r=1` of `reproject_image_to_polar` is not sufficient.

You made some nice plots about 1 year ago (https://github.com/PyAbel/PyAbel/issues/39#issuecomment-168664008), which show that dr=0.5 looks much better than dr=1.0. Actually, looking at those plots makes it seem like, if we are really trying to line up the profiles, then we may want to use dr=0.1.

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+        nmimage.zoom image before analysis. This may help improve the low
+        radius slice profile.
+
+    smooth: float
+        smoothing for spline interpolation of the determined radial scaling
```
DanHickstein: "This value is passed to the `scipy.interpolate import UnivariateSpline` function and controls how smooth the spline interpolation is. A value of zero corresponds to a spline that runs through all of the points, and higher values correspond to a smoother spline function. Typically this can remain zero, but use this if you are having trouble with the circularization algorithm."

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+        factor vs angle
+
+    nslices: int
+        divide the image into nslices. Ideally, this should be a divisor of
```
DanHickstein: What happens if it's not an even divisor? Should we throw a warning?

DanHickstein: Also, maybe include something like: "Larger values allow for the correction of distortions that change rapidly as a function of angle, but smaller values allow the algorithm to complete more quickly and are more robust to noise."

DanHickstein: I think that this keyword should be placed earlier in the list, since I think that it may require the most user intervention.

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+        the image column width.
+
+    ref_angle: None or float
+        reference angle for which radial coordinate is unchanged.
+        None uses numpy.mean(radial scale factors)
```
DanHickstein: append: "..., which attempts to maintain the same average radial scaling. This approximation is likely valid, unless you know for certain that a specific angle of your image corresponds to an undistorted image."

abel/tools/circularize.py
DanHickstein: Oh, also specify that zero angle is in the upwards vertical direction, where IM[0] corresponds to the top of the image, and the angle runs from -Pi to Pi

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+import abel
+
+# see https://github.com/PyAbel/PyAbel/issues/186 for discussion
+
+
+def circularize_image(IM, method="lsq", center=None, radial_range=None,
+                      zoom=1, smooth=0, nslices=32, ref_angle=None, 
+                      inverse=False, return_correction=False):
+    """
+    Remove radial distortion from a velocity-map-image through radial scaling
```
DanHickstein: I don't think that anything about the image circularization is specific to VMI images. Instead, this circularization can be applied to any image that has features that exist at a constant radius. So, perhaps the summary is something like: "This function corrects for radial distortion in an image that consists of features that exist at a constant radius. In other words, the image must consist of "circle-like" structures. This function attempts to correct for any distortion and ensures that the circle-like structures are perfect circles. 

This function is especially useful for correcting the image obtained with a velocity-map-imaging spectrometer with imperfect electro-static lenses, allowing a high-resolution 1D photoelectron distribution to be extracted. 

The algorithm is to split the image into "slices" at many different angles (set by `nslices`) and compare adjacent slices. A scaling factor is found which aligns each slice with the previous slice. The image is then corrected using a spline function that smoothly connects the discrete scaling factors as a continuous function of angle."

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+        divide the image into nslices. Ideally, this should be a divisor of
+        the image column width.
+
+    ref_angle: None or float
+        reference angle for which radial coordinate is unchanged.
+        None uses numpy.mean(radial scale factors)
+
+    inverse: bool
+        inverse Abel transform the *polar* image, to remove the background
+        intensity. This helps follow a given peak intensityr, for all angles,
```
DanHickstein: I'm not sure that this would *only* help when the anisotropy parameter is large. It seems like it might improve the SNR in many situations. 

Also, worth mentioning: "Note that this step is only for the purposes of allowing the algorithm to better follow peaks in the image. It does not affect the final image that is returned, expect for (hopefully) slightly improving the precision of the distortion correction." 

"intensityr" -> "intensity"

abel/hansenlaw.py
```diff
@@ -170,6 +170,10 @@ def igammagt(Nm, lam, n):
     # center pixel column
     AIM[:, 0] = AIM[:, 1]
 
+    # for some reason shift by -0.3 pixel aligns? - Fix me!
+    if direction == 'inverse':
+        AIM = shift(AIM, (0, -0.30))
```
DanHickstein: Weird...

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+    correction to the whole image grid. The image is remapped to the
+    correct cartesian grid.
+
+    Parameters
+    ----------
+    IM : numpy 2D array
+
+    method: str
+        method to align slice profiles,
+        "argmax" - compare intensity-profile.argmax() of each radial slice.
```
DanHickstein: I think that the least-squares method is likely to produce good results in a broad range of situations. However, the argmax method relies on the image having an obvious sharp peak, right? I think that we should make this clear:

"a very-fast but very-naive alignment method that aligns the adjacent slices based on the most intense pixel in each slice. This method may provide good results for images that have one sharp, intense radial peak. Otherwise it will not produce correct results." 

I also think that there would be a similar algorithm that would be essentially a combination of both methods:

calculate the least-squares error for a range of scaling factors (say, `linspace(0.9,1.1,20)`)
now, find the argmax of this array. Or, better yet, up-sample a bit and find the best scaling factor from a spline fit. 
Basically, this a nearly identical approach to what we are doing for `lsq`, except that we can try a bunch of scaling factors in parallel. It's probably much faster than `lsq`, and unlikely to catastophically fail like `least_squares` can sometimes do. But, it will be less precise.

abel/tools/circularize.py
DanHickstein: Alternative, a float center (x,y) can be used, right?

DanHickstein: Also, perhaps good to emphasize: "The circularization algorithm should be applied to a well-centered image..."

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+
+    (if return_correction is True)
+    slice_angles: numpy 1D array
+        Mid-point angle (radians) of image slice
+
+    radial_corrections: numpy 1D array
+        radial correction scale factors, at each angle
+
+    spline_function: numpy func()
+        spline function used to evaluate the radial correction at each angle
```
DanHickstein: "each" -> "any"

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+
+    """
+
+    if zoom > 1:
```
DanHickstein: I'd love to change this to just passing a dr to `reproject_image_into_polar` as mentioned above.

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+    # limit radial range of polar image, if selected
+    radial = radial_coord[:, 0]
+    if radial_range is not None:
+        subr = np.logical_and(radial > radial_range[0]*zoom,
+                              radial < radial_range[1]*zoom)
+        polarIM = polarIM[subr]
+        radial = radial[subr]
+
+    # split image into n-slices
+    slices = np.array_split(polarIM, nslices, axis=1)
```
DanHickstein: Would it be more straightforward to just utilize the `dt` keyword of `reproject_image_into_polar` to generate an image with the appropriate number of slices?  It would probably shorten the code here quite a bit.

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+    # evaluate radial scaling factor for each slice
+    radcorr = correction(slice_angles, slices, radial, method=method)
+
+    # spline radial scaling vs angle
+    radcorrspl = UnivariateSpline(slice_angles, radcorr, s=smooth, ext=3)
+
+    # apply the correction
+    IMcirc = circularize(IM, radcorrspl, ref_angle=ref_angle)
+
+    if zoom > 1:
```
DanHickstein: Even if we are going to use the "zoom" concept (which I think can be eliminated through `dr` as discussed above), then we could simply apply the correction to the un-zoomed image and eliminate this line here.

abel/tools/analytical.py
```diff
@@ -305,3 +306,52 @@ def sym_abel_step_1d(r, A0, r0, r1):
...
 
     return d
+
+
+def scaling(angle, amp=0.1, phase=2):
+    # define a simple scaling that will squish a circle into a "flower"
+    return 1 + amp*(np.sin(phase*angle)**4)
+
+
+def flower_distort(IM, amp=0.1, phase=2):
```
DanHickstein: So, I think that this function is equivalent to:

```python
def flower_scaling(theta, freq=2, amp=0.1):
    return amp*np.sin(freq*theta)**4

abel.tools.circularize.circularize(IM, radcorrspl=flower_scaling)
```

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+        # return to original image size
+        IMcirc = ndimage.zoom(IMcirc, 1/zoom)
+
+    if return_correction:
+        return IMcirc, slice_angles, radcorr, radcorrspl
+    else:
+        return IMcirc
+
+
+def circularize(IM, radcorrspl, ref_angle=None):
```
DanHickstein: "radcorrspl" should be "radial_correction_function" to make it clear that the function can be any function and does not need to be a spline.

abel/tools/circularize.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+    Remap image from its distorted grid to the true cartesian grid.
+
+    Parameters
+    ----------
+    IM : numpy 2D array
+        original image
+
+    radcorrspl: numpy func(theta)
```
DanHickstein: doesn't need to be a spline. But, the function does need to be able to accept a numpy array.

abel/hansenlaw.py
```diff
@@ -170,9 +170,9 @@ def igammagt(Nm, lam, n):
     # center pixel column
     AIM[:, 0] = AIM[:, 1]
 
-    # for some reason shift by 1/2 pixel aligns? - Fix me!
+    # for some reason shift by -0.3 pixel aligns? - Fix me!
     if direction == 'inverse':
-        AIM = shift(AIM, (0, -1/2))
+        AIM = shift(AIM, (0, -0.30))
```
stggh: @DanHickstein thanks for the comments. I will work through them.

Yes, an only on the `hansenlaw` inverse Abel transform. This is still an open issue #159.

abel/direct.py
```diff
@@ -74,7 +74,7 @@ def _construct_r_grid(n, dr=None, r=None):
 def direct_transform(fr, dr=None, r=None, direction='inverse',
                      derivative=gradient,
                      int_func=simpson_rule_wrong,
-                     correction=True, backend='C'):
+                     correction=True, backend='C', **kwargs):
```
stggh: Yes. Purely for the parameter `basis_dir=`

abel/tests/test_hansenlaw.py
```diff
@@ -67,7 +67,7 @@ def test_hansenlaw_inverse_transform_gaussian():
     tr = np.tile(ref.abel[None, :], (n, 1)) # make a 2D array from 1D
 
     recon = abel.hansenlaw.hansenlaw_transform(tr, ref.dr, direction='inverse')
-    recon1d = recon[n//2 + n%2]  # centre row
+    recon1d = recon[n//2]  # centre row
```
stggh: Agreed!

abel/tests/test_tools.py
```diff
@@ -65,20 +63,22 @@ def test_centering_function():
...
+
 def test_anisotropy_parameter():
     # anisotropy parameter from test image (not transformed)
     IM = abel.tools.analytical.sample_image(name='dribinski')
     
     Beta, Amp, Rmid, Ivstheta, theta = abel.tools.vmi.radial_integration(IM,
                                          radial_ranges=([(0, 33), (92, 108)]))
 
-    assert_almost_equal(-0.13, Beta[0][0], decimal=2)
+    assert_almost_equal(-0.14, Beta[0][0], decimal=2)
```
stggh: I believe this is related to a more correctly defined center, that then yields a more negative anisotropy parameter.  beta = -0.14 cf beta = -0.13 is not a big concern.

abel/tools/analytical.py
```diff
@@ -305,3 +306,52 @@ def sym_abel_step_1d(r, A0, r0, r1):
         d[:, mask] = abel_step_analytical(np.abs(r[mask]), A0, r0, r1)
 
     return d
+
+
+def scaling(angle, amp=0.1, phase=2):
```
stggh: Agreed

abel/tools/analytical.py
stggh: @DanHickstein that is pretty profound!

Not quite `1 + amp*np.sin(freq*theta)**4` ;-)

Removed `def flower_distort` from `abel/tools/analytical.py` and added your function suggestion to `examples/example_circularize.py`, `tests/test_circularize.py`.

abel/tools/analytical.py
```diff
@@ -305,3 +306,52 @@ def sym_abel_step_1d(r, A0, r0, r1):
...
+    """ Distort image with a sinusoidal radial scaling.  Intensity `*R**2`
+        Squish a circle into a "flower".
+
+    Parameters
+    ----------
+    IM : numpy 2D array
+        image
+    amp: float
+        amplitude of the sinusoidal radial perturbation
+    phase: float
```
stggh: fixed using your neat function definition

abel/tools/analytical.py
```diff
@@ -305,3 +306,52 @@ def sym_abel_step_1d(r, A0, r0, r1):
...
+
+    IM *= R**2 # making the sample image more beautiful...
+
+    R_rescaled = R*scaling(T, amp, phase) # angle-dependent scaling
+
+    # now, convert our squished angular grid back to a squished X, Y grid
+    X_rescaled = R_rescaled*np.cos(T)
+    Y_rescaled = R_rescaled*np.sin(T)
+
+    Z_rescaled = scipy.interpolate.griddata( (X_rescaled.ravel(),
```
stggh: Moot. Function now removed.

abel/tools/circularize.py
stggh: `smooth` is essential. Noisy data will cause the spline to go haywire. Conversely, too much smoothing will not fully undistort the image. It is essential that the user checks the evaluated correction points and the (smoothed) spline.

abel/tools/circularize.py
stggh: Good. Changed, verbatim.

abel/tools/circularize.py
stggh: For accurate results both `lsq` and `argmax` require the `radial_range=(rmin, rmax)` to be restricted. The main issue with `argmax` is the pixel resolution. Ideally, the peak should be subpixel sampled. Your, alternative method has merit, and is worth investigating. (later)

stggh: I will `git push` at this point. More to follow later. May be tomorrow, given the particular day of the year.

abel/tools/circularize.py
stggh: This parameter is purely a radial limit to restrict the intensity profile to the important part. There is no angular range specified.

abel/tools/circularize.py
stggh: This is an interesting point. I will investigate.

abel/tools/circularize.py
DanHickstein: Oh, whoops, read "radial" but though "angular". I guess the units are "pixels"...

abel/tools/circularize.py
stggh: Hmm, this could be neat. I will try it.

stggh:
@DanHickstein I often think that you should be writing **all** the (new) `PyAbel` code. The use of `abel.tools.polar.reproject_image_into_polar()` to generate the angular slices, [above],(https://github.com/PyAbel/PyAbel/pull/195#discussion_r100858559) is brilliant, and to employ the `dr, dt` parameters for zooming, clever.

I have implemented the method, removing the code associated with `numpy.array_split()`. 

There is still some tidying up of the code to do, to address your other suggestions, next.

abel/tools/circularize.py
stggh: changed

abel/tools/circularize.py
stggh: removed in place of `dt` of `abel.tools.polar.reproject_image_into_polar()`

abel/tools/circularize.py
stggh: Working. Great call!

abel/tools/circularize.py
stggh: done

DanHickstein:
> I often think that you should be writing all the (new) PyAbel code

Not true! A team is always best :).

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+    allowing a high-resolution 1D photoelectron distribution to be extracted.
+
+    The algorithm splits the image into "slices" at many different angles
+    (set by `dt`) and compares the radial intensity profile of adjacent slices.
+    A scaling factor is found which aligns each slice profile with the previous
+    slice. The image is then corrected using a spline function that smoothly 
+    connects the discrete scaling factors as a continuous function of angle.
+
+    This circularization algorithm should only be applied to a well-centered
+    image.
```
DanHickstein: add: "If your image is not already centered, use the `center` keyword (described below) to center it"

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+    Parameters
+    ----------
+    IM : numpy 2D array
+
+    method : str
+        method to align slice profiles,
+        "argmax" - compare intensity-profile.argmax() of each radial slice.
```
DanHickstein: I think that we should give some additional information about when users should select each method. 

 argmax "this method is quick and reliable, but it assumes that the radial profile has an obvious maximum. Also, it only centers each slice to within the nearest pixel."

lsq: "This method is slower and can fail to converge, but it may be applied to images with any (circular) structure. Also, it aligns the slices with sub-pixel precision."

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+        pixels, from the image center. Use to determine the distortion 
+        correction associated with particular peaks. 
+
+    dr : float
+        Radial grid size for the polar coordinate image, default = 0.5 pixel.
+        This is passed to :func:`abel.tools.polar.reproject_image_into_polar`.
+
+        Small values may improve the distortion correction, which is often of
+        sub-pixel dimensions, at the cost of reduced signal to noise for the
+        slice intensity profile.
```
DanHickstein: "As a general rule, dr should be significantly smaller than the "feature size" in the image."

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+
+    center : str, float tuple, or None
+        pre-center image using :func:`abel.tools.center.center_image`
+        "com", "convolution", "gaussian", "image_center", "slice", or
+        float tuple center (y, x).
+
+    radial_range : tuple, or None
+        limit slice comparison to the radial range tuple (rmin, rmax), in
+        pixels, from the image center. Use to determine the distortion 
+        correction associated with particular peaks.
```
DanHickstein: "It is recommended to select a region of your image where the signal-to-noise is highest, and where there are sharp features which are easy for the algorithm to follow."

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+        image that is returned, expect for (hopefully) slightly improving the
+        precision of the distortion correction.
+
+    return_correction : bool
+        Additional outputs, as describe below.
+
+    Returns
+    -------
+    IMcirc : numpy 2D array, same size as input
+        Circularized imput image
```
DanHickstein: "Circularized version of the input image"
(or, at least change "imput" to "input")

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+
+    (if return_correction is True)
```
DanHickstein: "the following values are returned only if return_correction is True"

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+    slice_angles : numpy 1D array
+        Mid-point angle (radians) of image slice
+
+    radial_corrections : numpy 1D array
+        radial correction scale factors, at any angle
```
DanHickstein: "radial correction scale factors for each angle in slice_angles"

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+
+    function : numpy function that accepts numpy.array
+        function used to evaluate the radial correction at each angle
```
DanHickstein: "function used to evaluate the radial correction at any angle"

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+
+    """
```
DanHickstein: I love this docstring! So much info!

abel/tools/circularize.py
```diff
@@ -0,0 +1,291 @@
...
+
+        for ang, aslice in zip(angles[1:], polarIMTrans[1:]):
+            profile = aslice
+
+            result = leastsq(_residual, fitpar, args=(radial, profile,
+                             previous))
+
+            radcorr.append(result[0][0])  # radial scale factor direct from lsq
+
+            previous += _residual(result[0], radial, profile, previous)
```
DanHickstein: add comment:
```
# This "previous" slice corresponds to the previous slice that has been re-scaled. Thus, if the next slice is identical, it will be assigned a scale factor of 1.0
```

abel/tools/polar.py
```diff
@@ -56,19 +56,19 @@ def reproject_image_into_polar(data, origin=None, Jacobian=False,
...
-    nr = np.round((r.max()-r.min())/dr)
+    nr = np.int(np.ceil((r.max()-r.min())/dr))
 
     if dt is None:
         nt = max(nx, ny)
     else:
-        # dt in degrees
-        nt = np.round((theta.max()-theta.min())/(np.pi*dt/180))
+        # dt in radians
+        nt = np.int(np.ceil((theta.max()-theta.min())/dt))
```
DanHickstein: The docstring for `reproject_image_into_polar` says that `dt` is in degrees. I don't mind changing this, and, in fact, I think that we should declare that PyAbel works in radians, not degrees (as I now have mentioned in #192.). 

So, can you change this in the docstring, and also change the docstring of `abel.tool.vmi.angular_integration()`?

DanHickstein:
Awesome work @stggh! I really like the improved docstring for circularize_image and the other changes.

I still need to review the new function in vmi.py, the documentation, and the example. Sorry to take so long to do this. It's been a little crazy this week.

stggh:
Thanks for all the inputs @DanHickstein. Sorry for all of my typos (new Mac = odd keyboard). I did pick up a few of my legacy typos in `vmi.py`, that have no keyboard excuse :-(, just a `vim` operator error.   

`Abel.tools.vmi.toPES()` may require an example. I will look at modifying one in the `examples/..` directory.

I think that we are asymptoting to a merge. AFAIK the only remaining issue is your suggested circumvent around using`leastsq()`, by evaluating the residual image intensity for various (linear) radial correction factors, interpolating the minimum difference to give the optimum correction factor. I have not tried this method, but my feeling now is that `script.optimize`, should be able to do the best job. Perhaps, `least_squares()` with variable constraints, or a another routine may be better?

abel/tools/polar.py
stggh: I agree with the choice of radians. There is a little more work, because `linbasex` uses degrees. I will look at.

stggh:
- `PyAbel`'s degrees are now in radians. The biggest hiccup was `linbasex` which used `proj_angles` in degrees, and the associated basis files which are labeled with them, now as % of &pi;.

- Changed `examples/examples_hansenlaw.py` to plot a photoelectron spectrum, as an example of the use of `abel.tools.vmi.toPES()`.

- Other minor typo fixes.

- added `.gitignore` to the new empty directories `examples/output` and `examples/bases` used to store the outputs of the example scripts. Needed for `git` to track an empty directory.

- AFAIK missing only the possibility of [replacing](https://github.com/PyAbel/PyAbel/pull/195#issuecomment-280503091) `leastsq` , which I have not tested.

DanHickstein:
Awesome work @stggh! 

I will get to proofreading the doc and example tomorrow.

I was not suggesting the we replace the leastsquares method, but simply offer another method that works slightly differently.

stggh:
No worries. All good.

examples/bases/.gitignore
```diff
@@ -0,0 +1,5 @@
+# empty directory to hold basis set files
```
DanHickstein: So, you got tired of all of the ugly basis sets running around? :)

examples/example_O2_PES_PAD.py
```diff
@@ -117,7 +117,7 @@ def PAD(theta, beta, amp):
                     wspace=0.35, hspace=0.37)
 
 # Save a image of the plot
-plt.savefig("example_O2_PES_PAD.png", dpi=100)
+plt.savefig("output/example_O2_PES_PAD.png", dpi=100)
```
DanHickstein: While I kinda like the idea of hiding the basis sets from the user in the "bases" folder, I am not sure if I see why we would want to hide the example output? Shouldn't we want this to be obvious and easy to find?

examples/example_circularize_image.py
```diff
@@ -0,0 +1,80 @@
...
+IM = abel.tools.analytical.sample_image(n=511, name='Ominus', sigma=2)
+
+# forward transform == what is measured
+IMf = abel.Transform(IM, method='hansenlaw', direction="forward").transform
+
+# flower image distortion
+def flower_scaling(theta, freq=2, amp=0.1):
+    return 1 + amp*np.sin(freq*theta)**4
+
+IMdist = abel.tools.circularize.circularize(IMf,
```
DanHickstein: include comment: "# distort the image"

examples/example_linbasex.py
```diff
@@ -15,13 +15,15 @@
 
 # Load image as a numpy array - numpy handles .gz, .bz2 
 IM = np.loadtxt("data/O2-ANU1024.txt.bz2")[::2,::2]
```
DanHickstein: Ah, sorry to have lowered your resolution :).  I'm sure you all worked hard to get the instrument up working so well, and it kills you to see it downsampled. I have a possible solutions:

In `setup.py`, we automatically detect if we are running on rtd. You could employ a similar strategy here, only downsampling if we are on rtd. I think you could try something like:
```python
import os
IM = np.loadtxt("data/O2-ANU1024.txt.bz2")
if os.environ.get('READTHEDOCS', None) == 'True':
    IM = IM[::2, ::2] # downsample image to avoid memory limitation of readthedocs.org
...
```

examples/output/.gitignore
```diff
@@ -0,0 +1,5 @@
+# empty directory to hold example output images
```
DanHickstein: Not sure if we should add a folder (as mentioned above).

abel/tools/circularize.py
```diff
@@ -0,0 +1,315 @@
...
+
+    The algorithm splits the image into "slices" at many different angles
+    (set by `dt`) and compares the radial intensity profile of adjacent slices.
+    A scaling factor is found which aligns each slice profile with the previous
+    slice. The image is then corrected using a spline function that smoothly 
+    connects the discrete scaling factors as a continuous function of angle.
+
+    This circularization algorithm should only be applied to a well-centered
+    image, otherwise use the `center` keyword (described below) to center it.
+
```
stggh: Neat. Thanks for this info. At the moment, something is not quite right with the code auto doc generation. I will check further tomorrow.

abel/tools/vmi.py
```diff
@@ -243,3 +243,70 @@ def PAD(theta, beta, amplitude):
...
     return (beta, error_beta), (amplitude, error_amplitude)
+
+
+def toPES(radial, intensity, wavelength, Vrep, R2E, zoom=1, energy_form='eBE',
+          units='cm'):
+    """
+    Convert speed radial coordinate into electron binding/kinetic energy.
+    Return the photoelectron spectrum (PES).
+
+    NB requires experiment parameters, measurement `wavelength`,
```
stggh: 285-287 has:
>             # This "previous" slice corresponds to the previous slice intensity
            # profile that has # been re-scaled. Thus, if the next slice is 
            # identical, it will be assigned a scale factor of 1.0

abel/tools/vmi.py
```diff
@@ -243,3 +243,70 @@ def PAD(theta, beta, amplitude):
...
+    energy_form: str
+         ``eBE`` or ``eKE``, electron binding energy or electron kinetic energy
+
+    units: str
+         ``cm-1`` or ``eV``
+          
+    """
+
+    # transform radius to electron binding energy
+    eBE = 1.0e7/wavelength - radial**2*(np.abs(Vrep)/1000)*R2E/100/zoom**2
```
stggh: Some good points. 

The experimental parameters are the wavelength, and repeller voltage. If all were fair in the experimental world the calibration factor `R2E` would be a constant. In the real world, `R2E` is approximately constant. Combining these three factors into one results in something that changes with every experiment = not very useful. 

I am not sure what you mean by photon energy vs wavelength `Ephoton(cm-1) = 1.0e7/wavelength(nm)`. The experiment uses laser wavelength.

abel/tools/vmi.py
```diff
@@ -243,3 +243,70 @@ def PAD(theta, beta, amplitude):
...
+        eBKE = 1.0e7/wavelength - eBE   # in cm-1
+    else:
+        eBKE = eBE
+
+    if units == 'eV':
+        eBKE /= 8065.541  # convert cm-1 to eV
+
+    # Jacobian correction to intensity, radius has been squared
+    # Not sure why divide, but this is same as for basex?
+    intensity[1:] /= radial[1:]  # 1: to exclude R = 0
```
stggh: This is interesting. In general the intensity is arbitrary, so I have not given this much thought. I do know that the how the relative spectral peak intensities should behave, and that has been my criterion for the Jacobian.  However, an absolute intensity could be useful.

examples/example_O2_PES_PAD.py
stggh: Because of the proliferation of files that are generated. The poor user may not know what to run next. I often use the example scripts to determine what parameters to use. They become hidden. An alternative may be to rename the output plots to plot_example_xxx.png. 

I may be alone in this opinion.

examples/example_linbasex.py
stggh: Yes, I would refer `2048x` images ;-)

Neat! Clever! Thanks!

examples/output/.gitignore
stggh: I am in favour, but am open to popular opinion. @PyAbel/developers

stggh:
Thanks again @DanHickstein, excellent and well considered comments. I will give some more thought to `toPES` (and whether there is an issue with the doc generation. My version of `livehtml` has gone bad with the auto generated code docs).

Apologies, for my rushed reply and corrections. I had a day in Sydney and have just returned.

examples/example_O2_PES_PAD.py
DanHickstein: Fine with me to rename the output files to "plot_..."

stggh: Ok, agreed. I will remove the "output" directory, and rename the example plots.

stggh:
I have:
- removed the `examples/output` directory. All plots are saved to `plot_example_xxx.png`
- changed `abel.tools.vmi.toPES()` to accept a single energy calibration factor to give an eKE output scale, and optional parameters `wavelength`, `Vrep` the which define the eBE output scale. `per_energy_scaling` parameter added.  
  I have removed the hidden factors of 1/1000 for the repeller voltage, and 1/100 for the old `R2E` factor. Now `energy_cal_factor`~1.15e-5 instead of ~1.15.
- pep8 checks for `abel.tools.circularize.py`, `abel.tools.vmi.py`

@DanHickstein I am not sure about using `:ref:` does it only apply to the doc files `.rst` or can it be used to reference a label placed in a function comment?

stggh:
Previous commit statement, not quite correct the renaming is:`abel.tests.test_circularize.py` -> `abel.tests.test_tools_circularize.py`.


@DanHickstein I think this PR is ready for a merge. There is a minor doc item, to reference a function using a label, but this may be addressed later.

DanHickstein:
Regarding the `:ref:` thing, according to the Sphinx docs, you can put a label (using `.. _my-reference-label:`) in any text that Sphinx formats (either the .rst files or the docstrings), and then use 
```
:ref:` The Reference <my-reference-label>` to include a link to it. 

http://www.sphinx-doc.org/en/stable/markup/inline.html#ref-role

But, I think that we can work on this later. 

I'll have a look at the improved toPES function now...

File: abel/tools/circularize.py [hunk 46]
```diff
@@ -0,0 +1,320 @@
...
+#
+# February 2017
+#########################################################################
+
+
+def circularize_image(IM, method="lsq", center=None, radial_range=None,
+                      dr=0.5, dt=0.5, smooth=0, ref_angle=None,
+                      inverse=False, return_correction=False):
+    """
+    Corrects image distortion on the basis that the structure (Newton spheres)
```
DanHickstein: Can you move the mention of "Newton Spheres" to the next paragraph, in order to keep the first paragraph general, and start into the specifics of VMI in the next paragraph?

abel/tools/vmi.py
```diff
@@ -243,3 +243,85 @@ def PAD(theta, beta, amplitude):
...
+        correspond to an "intensity per eV" or "intensity per cm-1".
+        If `False`, the returned intensities correspond to an "intensity per
+        pixel"
+
+    Optional:
+     wavelength: float
+         measurement wavelength in nm
+         e.g. `812.51 nm`, for "examples/data/O-ANU1024.txt"
+
+     Vrep: float
```
DanHickstein: I think that the Vrep should be removed since it's just a multiplicative scaling of the energy_cal_factor.

abel/tools/vmi.py
```diff
@@ -243,3 +243,85 @@ def PAD(theta, beta, amplitude):
...
+
+    Parameters
+    ----------
+    radial : numpy 1D array
+        radial coordinates
+
+    intensity : numpy 1D array
+        intensity values, at the radial array
+
+    energy_cal_factor: float
```
DanHickstein: We should mention that the units of the energy_cal_factor will affect the units of the output. Basically, if the user inputs an energy_cal_factor in eV/pixel^2, then the units will be in eV. Or, they can use J/pixel^2, or cm-1/pixel^2, whatever.

abel/tools/vmi.py
```diff
@@ -243,3 +243,85 @@ def PAD(theta, beta, amplitude):
...
+        e.g. `1.148427e-5` for "examples/data/O-ANU1024.txt"
+
+    per_energy_scaling : bool
+        sets the intensity Jacobian. If `True`, the returned intensities
+        correspond to an "intensity per eV" or "intensity per cm-1".
+        If `False`, the returned intensities correspond to an "intensity per
+        pixel"
+
+    Optional:
+     wavelength: float
```
DanHickstein: I think that we should have the user input the photon energy, rather than the wavelength, in whatever units they want. Currently, we are basically demanding that the user inputs the energy_scaling_factor in cm-1/pixel^2, and I suspect that many users won't like this. (For instance, I always worked in eV or J).

abel/tools/vmi.py
```diff
@@ -243,3 +243,85 @@ def PAD(theta, beta, amplitude):
...
+        eBKE = 1.0e7/wavelength - eKE
+    else:
+        eBKE = eKE
+
+    # Jacobian correction to intensity, radius has been squared
+    # We have E = c1 - c2 * r**2, where c1 and c2 are constants. To get thei
+    # Jacobian, we find dE/dr = 2c2r. Since the coordinates are getting
+    # stretched at high E and "squished" at low E, we know that we need to
+    # divide by this factor.
+    intensity[1:] /= (2*radial[1:])  # 1: to exclude R = 0
```
DanHickstein: Oh, I didn't realize that you were not actually interpolating to a new grid that is evenly spaced in energy. If you are not doing this, then it's unclear if you really need to apply the Jacobian. Basically, when you were trying to compare your data to existing data and found disagreement, it was likely because you were leaving your units in intensity/pixel while the data you were comparing to was in intensity/eV, which distorts the spectrum.

Anyway, my rule-of-thumb is that intensities should be expressed in "per the units of the x-axis of the plot". So, if I am plotting a photoelectron spectrum, I will plot it in counts/eV and use a dataset that has evenly spaced x-values. In this case, the x-values are not evenly spaced. So, I'm not sure if a user would expect the units to be in intensity/pixel, intensity, or intensity/eV.

I guess what we are doing is probably fine...

DanHickstein:
I think that this is looking pretty good! I made a few more comments on the toPES function. I'm wondering if it would be better to have the user input the photon energy (rather than the wavelength), since this would allow us to operate in a "unit agnostic" fashion, allowing the user to input whatever units for `energy_cal_factor` and `photon_energy` that they like (as long as they agree), and receive their units in the output. It takes care of J, cm-1, eV all at the same time.

abel/tools/vmi.py
stggh: This is an optional physical property of the VMI lens experimental parameter. It allows `energy_cal_factor` to remain constant between measurements (with different repeller voltages ages). IMO it is essential.

abel/tools/vmi.py
stggh: Added. I realised there is a missing `Returns` section of the documentation. Now added.

abel/tools/vmi.py
stggh: The function just sets `eKE = radial**2*energy_cal_factor`, so the units are user provided.

Optional `Vrep` includes the VMI physical factor.

Optional `wavelength` outputs electron-binding-energy (cm-1).

I think this covers all cases.

I have added a little more explanation to the optional parameters.

abel/tools/vmi.py
stggh: It is great having having your experience, and shared confusion on this ;-)

A very good point about the non-uniform grid. I added a comment to the `Returns` parameter doc.

stggh: > I guess what we are doing is probably fine...

Yes, it gives the correct result for experimental spectra of O<sup>-</sup> and O<sub>2</sub><sup>-</sup> PES.

stggh:
`toPES` is causing us a great deal of angst, due to our different uses (or expected uses). The single factor that you requested covers all cases, to generate photoelectron spectrum vs electron-kinetic-energy in user specified (`energy_cal_factor`) units. The other parameters, are optional, to make it easier to obtain a photoelectron spectrum vs electron binding energy.

It is likely that I have missed your point.  I have made several doc changes, please take another look at the function. Let me know if it still does not do what you might expect.

Thanks.

abel/tools/vmi.py
stggh: My statement should be correct "relative" intensity result.

abel/tools/vmi.py
```diff
@@ -287,16 +289,28 @@ def toPES(radial, intensity, energy_cal_factor, per_energy_scaling=True,
...
-         measurement wavelength in nm
+         measurement wavelength in nm,=. The output energy scale
+         is then set to electron-binding-energy (cm-1).
          e.g. `812.51 nm`, for "examples/data/O-ANU1024.txt"
 
      Vrep: float
-         repeller voltage
+         repeller voltage. Convenience parameter provided to allow the
+         `energy_cal_factor` to remain constant, for different VMI lens
+         repeller voltages.
```
DanHickstein: Include: "Defaults to None, in which case no further scaling is applied."

abel/tools/vmi.py
```diff
@@ -276,8 +276,10 @@ def toPES(radial, intensity, energy_cal_factor, per_energy_scaling=True,
         intensity values, at the radial array
 
     energy_cal_factor: float
-        energy calibration factor that will convert radius squared into energy
-        e.g. `1.148427e-5` for "examples/data/O-ANU1024.txt"
+        energy calibration factor that will convert radius squared into energy.
+        The units affect the units of the output. e.g. inputs in  eV/pixel^2,
+        will give output energy units in eV.
+        A value of  `1.148427e-5` applies for "examples/data/O-ANU1024.txt"
```
DanHickstein: "A value of 1.148427e-5 cm-1/pixel^2 ..."

abel/tools/vmi.py
```diff
@@ -287,16 +289,28 @@ def toPES(radial, intensity, energy_cal_factor, per_energy_scaling=True,
 
     Optional:
      wavelength: float
-         measurement wavelength in nm
+         measurement wavelength in nm,=. The output energy scale
```
DanHickstein: typo? 

Also mention that "if the wavelength is specified, then the units of energy_cal_factor *must* be in cm-1/pixel^2"

DanHickstein:
Thanks for making the changes, I think that they do clarify things.

I still don't understand why we would want to lock the binding energy spectrum to units of wavenumbers. Currently, a user can retrieve the KE spectrum in whatever units they want (which is great!), but as soon as they want to get the binding energy spectrum, then they must use wavenumbers. If we simply change the `wavelength=` argument to a `photon_energy=` keyword argument, then we can easily allow the user to use whatever units they want, as long as the units for `photon_energy` match the units of `energy_cal_factor`. We can include in the docstring that, if they want to use wavenumbers and enter the wavelength in nm, then they can call the function with `photon_energy = 1 / wavelength_in_nm * 1e7`

stggh:
Sounds good, `photon_energy` it is. Pretty amazing docs for a 2 line function ;-)  The function is very useful and a worthwhile addition to the `PyAbel` stable. 

Updated docs with markup to improve the look of `readthedocs` output, made `example_hansenlaw.py` use `photon_energy`, pep8 checks on `abel.tools.vmi.py` and `abel.tools.circularize.py`.

Thanks @DanHickstein for your great inputs and clever code improvements.

DanHickstein:
> Pretty amazing docs for a 2 line function ;-) 

Ha! True. 

But, it reflects the reality of the pixels -> energy conversion: the actual math is pretty simple, but doing it correctly is more complicated and confusing that it seems at first glance. It's great that we now give instructions to guide people through it.

I'd say we should just go ahead and hit the merge button on this one. But we should give @rth and @DhrubajyotiDas "Last Call" to chime in with and comments or concerns.

DanHickstein:
@stggh, Are the docs building alright? If so, I'd say that you can go ahead and press the big green button.

stggh:
Good question, and the answer is no. The docs build fine on my system. How do we track this down?

DanHickstein:
They are building fine on your system? Why do you say "the answer is no"? Are you also having readthedocs.org build the docs for your fork?

stggh:
Opps!  Too early in the morning.  Need the merge for the new docs.

Merging.

DanHickstein:
Whoops, the docs failed to build on readthedocs again: https://readthedocs.org/projects/pyabel/builds/5081882/

Same issue with memory on the example_linbasex.py.

We need to change[ line 19](https://github.com/PyAbel/PyAbel/blob/master/examples/example_linbasex.py#L19) from
```python
if os.environ.get('READTHEDOCS', None) == True:
```
to
```python
os.environ.get('READTHEDOCS', None) == 'True'
```
(The `os.environ.get` returns a string instead of a boolean. Tricky.)

stggh:
Would you like to do that directly, or should I create another PR?

DanHickstein:
I think that PRs are the way to go. Easier to keep track of changes. And we are close to breaking 200 Issues/PRs :)

stggh:
Closing comment.  This PR provides:

 - `abel.tools.circularize` functions to circularization a velocity-map image, through the radial alignment of adjacent slice intensity profiles. This improves the radial resolution (kinetic energy) of the angular integrated integrated spectrum, and allows co-averaging of quadrants to improve signal-to-noise, `symmetry_axis=(0, 1)` in `abel.Transform()`.

- `abel.tools.vmi.toPES()` function to convert the radial pixel scale to an energy calibrated scale e.g. into electron kinetic energy or electron binding energy, and so easily generate a photoelectron spectrum, from the radial intensity.

 - a bug-fix for `abel.tools.center` that +1 pixel off for odd-size images.

PR #197, opened 2017-03-03
Center image along one axis

stggh:
This PR:
- adds a new parameter `axes` to `abel.tools.center.set_center()` and its calling function `abel.tools.center_image()`, to apply the center correction only for the axis specified. It provides for automatic centering  (`com`, `convolution`, `slice`) if the image is symmetric along only one axis.

- Bug fix: `crop` parameter was in the docs of `center_image()`but not listed as an argument, or passed to `set_center()`.

I anticipate a discussion on whether the new parameter `axes` should be the axis along which the centering is applied (at the moment), or the axis of symmetry ...

### DIFF
--- a/abel/tests/test_tools_center.py
+++ b/abel/tests/test_tools_center.py
@@ -5,7 +5,7 @@ from __future__ import print_function
 import os.path
 
 import numpy as np
-import numpy.testing as npt
+from numpy.testing import assert_allclose
 
 import abel
 
@@ -25,12 +25,12 @@ def test_center_image():
     # find_center using 'slice' method
     center = abel.tools.center.find_center(IMx, center="slice")
 
-    npt.assert_almost_equal(center, true_center, decimal=0)
+    assert_allclose(center, true_center, atol=1)
 
     # find_center using 'com' method
     center = abel.tools.center.find_center(IMx, center="com")
 
-    npt.assert_almost_equal(center, true_center, decimal=0)
+    assert_allclose(center, true_center, atol=1)
 
     # check single axis - vertical
     # center shifted image IMx in the vertical direction only
@@ -38,14 +38,14 @@ def test_center_image():
     # determine the center
     center = abel.tools.center.find_center(IMc, center="com")
 
-    npt.assert_almost_equal(center, (179, 180), decimal=0)
+    assert_allclose(center, (179, 180), atol=1)
 
     # check single axis - horizontal
     # center shifted image IMx in the horizontal direction only
     IMc = abel.tools.center.center_image(IMx, center="com", axes=0)
     center = abel.tools.center.find_center(IMc, center="com")
 
-    npt.assert_almost_equal(center, (180, 182), decimal=0)
+    assert_allclose(center, (180, 182), atol=1)
 
     # check even image size returns odd
     # drop off one column, to make an even column image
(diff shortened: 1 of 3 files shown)


abel/tools/center.py
```diff
@@ -104,6 +104,9 @@ def center_image(IM, center='com', odd_size=True, square=False, verbose=False,
             the image will be padded with zeros such that none of the original
             image will be cropped.
 
+    axes : int or tuple
+        center image with respect to axes ``0``, ``1``, or both ``(0, 1)`` (default)
```
DanHickstein: Can we mention that 0 is "vertical" and 1 is "horizontal"?

abel/tools/center.py
```diff
@@ -172,6 +176,9 @@ def set_center(data, center, crop='maintain_size', verbose=False):
         ``maintain_data``
             the image will be padded with zeros such that none of the original
             image will be cropped. 
+
+    axes : int or tuple
+        center with respect to axis ``0``, ``1``, or both axes `(0, 1)` default.
```
DanHickstein: also "horizontal" and "vertical" here...

abel/tools/center.py
```diff
@@ -214,7 +221,13 @@ def set_center(data, center, crop='maintain_size', verbose=False):
     if verbose:
         print("delta = ({0}, {1})".format(delta0, delta1))
 
-    centered_data = shift(data, (delta0, delta1))
+    if isinstance(axes, int):
+        if axes in [0, 1]:
+            centered_data = shift(data, (delta0*(1-axes), delta1*axes))
```
DanHickstein: This is clever, but I think that the code would be more readable as just: 
```python
 if isinstance(axes, int):
    if axes == 0:
        centered_data = shift(data, (delta0, 0))
    elif axes == 1:
        centered_data = shift(data, (0, delta1))
    else:
        raise ValueError("axes value must be 0, or 1.")
else:
    centered_data = shift(data, (delta0, delta1))
```

stggh:
Good suggestions. All done, together with pep8 corrections.

stggh:
- Added unit test for single axis centering in `abel.tests.test_tools_center.py`. There was a parameter `axis=(0,1)`, originally passed, quietly, but not implemented in `find_center()` . IMO the use of  ``**kwargs` is bad, since parameter checking is not done.

- Switched testing in `abel.tests.test_center.py` to `numpy.testing.assert_almostequal()` since this uses a `decimal=#` comparison, which IMO is more clear than `atol= , rtol= ` of `np.allclose()`. However, I note the API for `testing.assert_almostequal` has the warning:
   > It is recommended to use one of assert_allclose, assert_array_almost_equal_nulp or assert_array_max_ulp instead of this function for more consistent floating point comparisons.
 
  so I am open to advice on the better assertion check.

- pep8 checks on `abel.tests.test_tools_center.py` and `abel.tools.center.py`

abel/tests/test_tools_center.py
```diff
@@ -5,39 +5,57 @@
 import os.path
 
 import numpy as np
-from numpy.testing import assert_allclose
+import numpy.testing as npt
```
DanHickstein: We can already access `assert_allclose` via `np.testing.assert_allclose`. The following is slightly more verbose, but I think it's more clear.

stggh:
I have no particular objection either. I have changed `assert_almost_equal()` to `assert_allclose()` in both `abel.tests.test_tools_center.py` and `abel.tests.test_tools_circularize.py` to be consistent.

PR #200, opened 2017-12-16
Included "dr" for onion_bordas

DanHickstein:
Hi everyone!

I was trying to make a figure to compare the results of each method with the analytical solution for a Gaussian, and I noticed that the `onion_bordas` method doesn't apply the `dr` factor. While I don't think that we need to spend much time maintaining this method, since the dasch onion peeling method seems to outperform the onion_bordas in all respects, this seems like an easy fix. 

All we need to do is include the `1/dr` term in the onion_bordas method.

I also fixed the docstring in `dasch.py` to correctly indicate that `dr` is applied for all of the dasch methods.

Here is what it looked like before the fix ([code here](https://gist.github.com/DanHickstein/68aa7d3891691dda4c6655392aa37819)):
figure_1


DanHickstein:
The TravisCI build failed for Py3.4 for some silly reason. I was looking into fixing it, but then I realized that Py3.6 is out, so maybe we should just test Py3.5 and 3.6 and not be testing Py3.4 anyway. (Py2.7 testing is still there.)

Thoughts?

stggh:
@DanHickstein good find! I presume the corresponding `tests/test_onion_bordas.py` will require a  tweak as well? 

For the travis-ci, there is now [no need to use Miniconda:](https://gist.github.com/dan-blanchard/7045057). Add the following to `.travis.yml` and delete the `miniconda` stuff.
```
pip install --upgrade pip setuptools wheel
pip install --only-binary=numpy,scipy numpy scipy
```

IMO we should not even bother to support `python2.7` it finishes in 2020. There is no reason to stay with `2.7`. Pass (old) `2.7` code through `2to3 -w`.

Sorry, for the lack of noise on our `PyAbel` paper. I have been sidetracked.

stggh:
Hmm, I should really push more often. Held up by issues with `abel/fourier_expansion.py` (too slow) and `abel/fourier_hankel.py` (transfrom pairs fail from some grid sizes). My current fork has `dasch.py`:
```python
    dr : float
        sampling size (=1 for pixel images), used for Jacobian scaling.
```
So, I had seen the error ;-). No matter @DanHickstein's fix is good.

While we are tidying up `abel/dasch.py`  
```python
def get_bs_cached(
 :
``` 
lines `260:` onward may be deleted. The (common) code is in `tools/basis.py` and so no longer required in `dasch.py`.

rth:
> For the travis-ci, there is now no need to use Miniconda:.

Well, even though numpy and scipy now have pre-build wheels, conda allows to standardize a bit the setup on different OSs. Although the current .travis conda setup is a bit convoluted and I agree that it might be worth either switching to pypi or possibly using [asropy ci-helpers](https://github.com/astropy/ci-helpers)...

> IMO we should not even bother to support python2.7 it finishes in 2020

+ 1
 although from experience some people still run Python 2.7 so this might reduce the number of users who will be able to try PyAbel...

stggh:
There is a `travis-ci` build issue with some versions of `miniconda` and various `scipy` packages.  I forget the full detail, but it forced a switch to `pip` for my [PyDiatomic](https://github.com/stggh/PyDiatomic) code.

`2.7` - even [micropython](https://micropython.org) (used for IoT devices) has switched ;-)

DanHickstein:
> I presume the corresponding tests/test_onion_bordas.py will require a tweak as well?
Funny enough, the 1D gaussian test still passed, because the test was set to:
```python
assert_allclose(ratio, 1.0, rtol=1, atol=0)
```
and the rtol=1 was pretty forgiving. There was also a `/2` factor that shouldn't have been there.

> lines 260 onward may be deleted. The (common) code is in tools/basis.py and so no longer required in dasch.py.

Good point. I changed the function call to `abel.tools.basis.get_bs_cached` and removed the vestigial code.

Regarding the changes to the `.travis.yml`, it sounds like @stggh is in favor of changing to `pip` while @rth likes the flexibility of miniconda? I don't have strong feelings either way, but maybe we can merge this and open a separate issue for `.travis.yml` edits.

stggh:
> open a separate issue for .travis.yml edits.

Yes, a separate issue is fine.

DanHickstein:
Alright, I'm going to merge this thing!

Thanks for noting the extra code in `dasch.py`. Adding code is good, but deleting code is even better :)

stggh:
Sorry, probably my fault, something that I no doubt intended to get back to. In the early days I was not too concerned with the transform intensity. Now I know better.

PR #201, opened 2017-12-21
Analytical transform pair functions

stggh:
This PR:
1. revamps `abel/tools/analytical.py` returning the structure to the original Class (`BaseAnalytical`, `StepAnaytical`, and `GaussianAnalytical`) with methods. 
2. Adds a new `class TransformPair` (and associated file `abel/tools/transform_pairs.py`) containing the 7 analytical transform pair functions of Table 1, G. C.-Y Chan and G. M. Hieftje *Spectrochimica Acta B* 61, 31-41 (2006), [doi:10.1016/j.sab.2005.11.009](http://www.sciencedirect.com/science/article/pii/S0584854705003344?via%3Dihub). 

      The code mirrors @DhrubajyotiDas [gist](https://gist.github.com/DhrubajyotiDas/f8f531ca2cfa3fee6345), but makes it available to `PyAbel`.

   As [noted](https://github.com/PyAbel/PyAbel/issues/19#issuecomment-158244527) by @DhrubajyotiDas profile4 has typographical errors that are not trivial to resolve.

3. The function `abel.tools.analytical.sample_image()` function has been made into a Class, to be consistent with the original `analytical.py` class concepts,  with attributes `image` and `name`.
4. Changed `abel/tests` and `examples/` files to use the attribute (and name change): `IM = abel.tools.anaytical.SampleImage(n).image`. 

The new `TransformPair` class provides easy access to the 7 tabulated profiles e.g.
```python
f = abel.tools.analytical.TransformPair(n=101, profile=5)
```
together with the attributes `r, dr, func, abel, label` providing for fast `PyAbel` testing.

Note: I realise merging a new PR at Xmas may be a recipe for disaster. I am happy to let this one sit over the holiday break. However, I find the availability of the 7 profiles within `PyAbel` useful for testing.


DanHickstein:
Thanks @stggh! At first glance, it looks great. But, it's a lot of code, so give me a few more days to go over it.

stggh:
Suits me, I will be away from easy internet access, until mid-next week.

abel/linbasex.py
```diff
@@ -235,7 +238,7 @@ def _linbasex_transform_with_basis(IM, Basis, proj_angles=[0, np.pi/2],
 
     radial = np.linspace(clip, cols//2, len(Beta[0]))
 
-    return inv_IM, radial, Beta, QLz
+    return inv_IM*2/cols, radial, Beta, QLz
```
DanHickstein: I'm not sure if this is really the correct scaling factor after all. Perhaps add a comment that we are not yet sure of the correct scaling factor.

stggh:
Update: This PR incorporates PR #204, and raises the even image size error to fix #202. The changes are quite minimal, adding the `abel/tools/transform_pairs.py` functions `profile1 ... profile8`, and modifying `abel/tools/analytical.py` to follow the original Class methods of `BaseAnalytical`. Previous function calls to `sample_image` are replaced with class method calls.

abel/linbasex.py
stggh: Ok, checking this. Note, the image scaling will have no effect on the radial intensity.

stggh: You are correct. `2/cols` works for `O2-ANU1024.txt` and `O2-ANU2048.txt`, but not `Sample.Image(n)`:
linvs3pt
[code gist](https://gist.github.com/stggh/2f898afb902867f26cdf29ba48c9d0f7)

I have removed the factor and added a comment referencing issue #202

abel/tools/transform_pairs.py
```diff
@@ -0,0 +1,146 @@
...
+#   see abel/tools/analytical.py for Class attributes
+# 
+##############################################################################
+
+def a(n, x):
+    return np.sqrt(n*n - x*x)
+
+
+def profile1(r):
+    # Cremers and Birkebak App. Opt. 5, 1057-1064 (1966) Eq(13)
```
DanHickstein: Can we change these comments to docstrings, so that they will be included in the documentation? I think it would also be nice to have a one sentence summary of that the function looks like. Something as simple as "Starts at 1 at r=0 and decays slowly to 0 at r -> infinity."

I think that you are assuming that people won't want to use these directly, since it's probably better to use them through abel.tools.analytical.TransformPair, but it seems possible that people may want to use these directly. 

You may want to add the line about "call this function using abel.tools.analytical.TransformPair" to each docstring. I suppose I'm advocating to append an identical few-line docstring to each of thees functions, with a unique sentence for each.

abel/tools/analytical.py
```diff
@@ -250,59 +208,177 @@ def __init__(self, n, r_max, sigma=1.0, A0=1.0,
...
+            curve B Hansen and Law JOSA 2, 510 (1985)
+            doi:10.1364/JOSAA.2.000510
+
+        Parameters
+        ----------
+        n : int
+            number of points along the r axis
+
+        profile: int
+            the profile number 1-8, see 'abel/tools/transform_pairs.py'
```
DanHickstein: note the profile 4 is broken.

abel/tools/analytical.py
```diff
@@ -250,59 +208,177 @@ def __init__(self, n, r_max, sigma=1.0, A0=1.0,
...
+
+            mask_valid : boolean array
+                set all True. Used for unit tests
+
+
+        """
+
+        super(TransformPair, self).__init__(n, r_max=1, symmetric=False)
+
+        # BaseAnalytical creates self.r = [0, 1]
```
DanHickstein: self.r is a linspace from 0->1, right? Not just a list of 0 and 1. (Math notation versus python notation).

abel/tools/analytical.py
```diff
@@ -250,59 +208,177 @@ def __init__(self, n, r_max, sigma=1.0, A0=1.0,
...
+        # prevent divide by zero and NaN for r = 0, or 1, 
+        # slightly offset these values
```
DanHickstein: Move these comments to label each line separately?

examples/example_direct_gaussian.py
```diff
@@ -21,36 +21,36 @@
...
-recon = direct_transform(ref.abel, dr=ref.dr, direction="inverse",
-                         correction=True)
-
-ax[1].plot(ref.r, recon , '--o', c='red', label='direct')
-
-recon = direct_transform(ref.abel, dr=ref.dr, direction="inverse",
+# forward Abel transform
+reconC = direct_transform(ref.func, dr=ref.dr, direction="forward",
+                          correction=True, backend='C')
+reconP = direct_transform(ref.func, dr=ref.dr, direction="forward",
```
DanHickstein: We are now pretty confident that the Direct-C and Direct-P methods produce identical results (it's a unit test now). So, perhaps it's more interesting to plot one of the Direct methods with and without the correction?

abel/linbasex.py
DanHickstein: Okay, fine to leave this issue for another day. Can you open an Issue saying that we still need to figure out the scaling for Lin-BASEX?

abel/tools/transform_pairs.py
stggh: Good comments. 
1. Transform pairs may now be accessed directly via `abel.tools.transform_pairs.profile#(r)` where `# = 1-8`.
2. Profile functions will now accept a single value for r.
3. Added r range check to ensure  `0 <= r <=1`, although a few of the functions require 0 < r <= 1 or 0 < r < 1. This is checked, but the docs may need improvement to address this subtly. 
Items 1-3 are not required for the `abel.tools.analytical.TransformPair` class, since it generates `r=np.linspace(0, 1, n)` and slightly, offsets the limits to prevent NaNs.
4. Added docstrings to each profile and ascii plots(*) of the source and projection functions, as a picture is worth a 1000 words (and these profiles are not easy to describe).

(*) ascii plots from the [AP package](https://github.com/mfouesneau/asciiplot)

abel/tools/analytical.py
stggh: Done. However, the source function has only a slight discontinuity, and so in principle we may generate and least-squares fit our own projection function.

abel/tools/analytical.py
stggh: That was an unfortunate math notation [0, 1] inclusive, (0, 1] exclude 0, (0, 1) exclude 0, 1. Now removed.

abel/tools/analytical.py
stggh: done

examples/example_direct_gaussian.py
stggh: Ok. Why use `correction=False`?

plot_example_direct_gaussian

stggh:
Hold off with the doc critique until delete the dumb `.replace()` = work in progress

stggh:
OK, this PR should be pretty close for review. 

For the docs I cannot understand why `abel/transform.py` creates an `__init__()` description, but `abel/tools/analytical.py` does not. The autodoc generation does not recognise the Class doc keyword `Attributes` and so I have replaced (both cases) with `Returns`, to make `readthedocs` look better.

Also, I do not know how to read an `.rst` file into `doc/abel.rst`. At the moment there is text below `abel.tools.transform_pairs module`.

I will see whether i can add axis scales to the plots. The ascii plotting package [AP](https://github.com/mfouesneau/asciiplot) does not round the axis values, which IMO makes the graphics confusing.

Call method for the transform pair functions:
```python
In [8]: func = abel.tools.analytical.TransformPair(n=101, profile=8)

In [9]: func.
              abel       label      profile()  
              dr         mask_valid r          
              func       n          r_max      

```
or function direct, accepts single float for `r` or an array of floats:
```python
In [12]: x = abel.tools.transform_pairs.profile8(r=0.5)

In [13]: x
Out[13]: (array([ 1.02859087]), array([ 1.24304308]))
```

DanHickstein:
Thanks for making these changes @stggh! The ASCII plots are really amusing :)

I'm out this weekend, but I will try to look at the new changes on Mon or Tues. Sorry for the delay!

DhrubajyotiDas:
Never seen this before in a docstring - this is pretty cool!

image

stggh:
Thanks @DanHickstein, @DhrubajyotiDas, enjoy your record big chill, while this weekend Australia has its (record?) heatwave

DhrubajyotiDas:
What big chill? :)

image

stggh:
**Profile4** via least squares fitting

*Source function* is shown in Fig. 5 of [Alvarez et al.](http://www.sciencedirect.com/science/article/pii/S0584854702000873?via%3Dihub), 

may be digitized using [engauge](http://markummitchell.github.io/engauge-digitizer/) and fitted to the profile 4 polynomial expression, with the coefficients as variables. It turns out that I mistyped one coefficient, `155.56`, and that the published coefficients are correct.

*Projection* function may be obtained by a forward transform (`hansenlaw`) of profile4 applying a least-squares fit. Here, the `r <= 0.7` expression breaks for the last 3 coefficients, which should be ~`64.17, -17.22, -13.58`, and for `r > 0.7` the 2nd coefficient should be `196.26`, not `193`. 

See `abel/tools/transform_pairs.py` code (updated).

Not perfect, but close(r):
profile4plt

code used for plot:
```python
# profile4
p4 = abel.tools.analytical.TransformPair(101, 4)

# forward Abel transform
hlf = abel.hansenlaw.hansenlaw_transform(p4.func, dr=p4.dr,
                                         direction='forward')

# inverse Abel transform
tpi = abel.dasch.three_point_transform(p4.abel, dr=p4.dr)
```

[fit code](https://gist.github.com/stggh/6efbb024678e48e1a0873b70ad7d6c92) for reference.

DhrubajyotiDas:
Thanks for pointing me towards Engauge - was not aware of it before!

Can you comment on what, if any, were the typos in the formulation of profile4 in the original source? You said that the published coefficients are correct - my understanding was that that was not the case - maybe I'm a bit confused here as to what the coefficients are referring to.

stggh:
table1profile4

> Can you comment on what, if any, were the typos in the formulation of profile4 in the original source

No typos, the source function formula is correct.

> coefficients are referring to

The coefficients of the source  polynomial `0.1`, `5.51`, `-5.25` for `r <= 0.7` amd `-40.74`, `155.56`, `-88.89` and `74.07`. 

I have repeated the whole procedure just to be sure:
1. digitize Fig5 curve. To be more accurate with the axis points `(0,0), (1,0)` and `(0, 1)` have been marked these points in red using gimp:
fig5
2. Python [lmfit](https://lmfit.github.io/lmfit-py/) allows easy turning on/off parameters `l0, l1, l2` and `r0, r1, r2, r3` the above coefficients. The calculated profile is "good" for the source parameters as specified (above):
profile4fix

Allowing parameter  `r1` to vary, gives a fit value close to the original. Note, here I am only playing with the right size ` r> 0.71` of the function. The same procedure applies to the left side, for the coefficients `l0, l1, l2`:
```python
pars.add('l0', value=0.1, vary=False)
pars.add('l1', value=5.51, vary=False)
pars.add('l2', value=5.25, vary=False)

pars.add('r0', value=40.74, vary=False)
pars.add('r1', value=155.56, vary=True)
pars.add('r2', value=188.89, vary=False)
pars.add('r3', value=74.07, vary=False)
```
Screen output:
```
[[Variables]]
    l0:   0.1 (fixed)
    l1:   5.51 (fixed)
    l2:   5.25 (fixed)
    r0:   40.74 (fixed)
    r1:   155.554464 +/- 0.004352 (0.00%) (init= 155.56)
    r2:   188.89 (fixed)
    r3:   74.07 (fixed)
```
similarly for `r2` `    r2:   188.895436 +/- 0.004976 (0.00%) (init= 188.89)` etc.

If all the coefficients are allowed to vary then a different set of coefficients results, because of the digitizing error, **but**, we are looking for publication typos, and may assume most of the coefficients are correct.

[gist code and data](https://gist.github.com/stggh/6bf40d2c1f55134c5ed860ded411c357)

stggh:
Hmm! On a finer scale the left (`r <= 0.7`) and right (`r > 0.7`) sides are discontinuous, for the published coefficients (of profile4), so a small tweak of the published coefficients is required:
break

DanHickstein:
@stggh, I like the addition of the equations and the ASCII plots to the docstrings for the functions. 

How close are we to being able to merge this PR? You mentioned that you were having some problems with getting the readthedocs to build - have these been solved, or are there still lingering issues? Also, it seems like there was some controversy with profile4 - is there a conclusion on this? It's fine with me to leave this one incomplete as along as it's inaccuracies are clearly described in the docstring, and perhaps a warning is raised.

stggh:
Profile4 is problematic. I modified the source and projection profiles to align the left side `r <= 0.7` to the respective right side `r > 0.7`, by adding the "small" difference at `r=0.7` to the right side profile. This looks *great* for `n=101`, but not so good for `n=1001`:
profile4plt

I am now of the opinion that profile4 should use the published incorrect constants of [Table1 above](https://github.com/PyAbel/PyAbel/pull/201#issuecomment-356743172) and issue some form of warning message. 

> inaccuracies are clearly described in the docstring, and perhaps a warning is raised.

I will work on that now.

> How close are we to being able to merge this PR? 

Should be ready once I fix the above.

> readthedocs to build - have these been solved

No issues AFAIK

stggh:
I have returned `profile4` to the published coefficients, and the corresponding ascii plot, added a warning message (which I am happy to tweak), and renamed some of variables, in the split functions, so that the profile construction code is clearer (I hope).

Ready to merge.

stggh:
Ok to merge?

stggh:
@DanHickstein ok to merge this?

DanHickstein:
Yes! You can go ahead and merge! Thanks for all of your hard work on this!

Issue #58, opened 2015-12-16
Ensure that Python and Cython versions of abel.direct produce identical results


DanHickstein:
As mentioned in PR https://github.com/PyAbel/PyAbel/pull/52 and https://github.com/PyAbel/PyAbel/pull/52#issuecomment-164826567, the results of the Python and Cython implementations of the direct fabel/iabel transforms are similar, but not identical. At some point, we should track down the cause of this and figure out how to make them exactly the same.

rth:
At this points, it appears clear that I won't have the bandwidth to finish this, so if anyone is motivated to take up this issue they can feel free to do so.

DanHickstein:
It seems like the simpson's rule integration might be leading us astray. Perhaps if we switch both methods to a `trapz` integration, then the issue will be easier to diagnose. @rth, if you have a moment, can you demonstrate how to switch the Cython code to a trapz integration?

stggh:
With the `O2-ANU1024.txt` image `direct-C` is similar to `direct-python`. This is a plot of `row=512` of the transformed image:
![o2-directpc](https://user-images.githubusercontent.com/10932229/34464890-e9844fd6-eee8-11e7-8147-ffc6585e234b.png)
```python
im = np.loadtxt('O2-ANU1024.txt.bz2')

directC = abel.Transform(im, method='direct', center='convolution',
                         transform_options=dict(backend='c', dr=1)).transform

directP = abel.Transform(im, method='direct', center='convolution',
                         transform_options=dict(backend='python', dr=1)).transform
```

However, for the *1D* analytical functions there is a substantial difference, which may point to some truncation error in the true python version:
![ck](https://user-images.githubusercontent.com/10932229/34464892-0269e7cc-eee9-11e7-9967-9ded112ae6b7.png)

Interestingly, `direct-c` mirrors `hansenlaw`. As usual `three_point` is the most accurate.

Transform code:
```python
import numpy as np
import abel
import matplotlib.pyplot as plt

n = 101
f = abel.tools.analytical.TransformPair(n, profile=8)

hlf = abel.hansenlaw.hansenlaw_transform(f.func, dr=f.dr, direction='forward')
hli = abel.hansenlaw.hansenlaw_transform(f.abel, dr=f.dr)

tpi = abel.dasch.three_point_transform(f.abel, dr=f.dr, basis_dir=None)
dic = abel.direct.direct_transform(f.abel, dr=f.dr, backend='c')
[...]

PR #204, opened 2018-01-01
Making direct-C and direct-Python agree

DanHickstein:
This addresses https://github.com/PyAbel/PyAbel/issues/58. This is still a work in progress, but I'm starting a PR to facilitate comparison of the changes.


abel/direct.py
```diff
@@ -187,17 +188,24 @@ def _pyabel_direct_integral(f, r, correction, int_func=simpson_rule_wrong):
...
+    mask = (II < JJ) 
 
     I_sqrt = np.zeros(R.shape)
     I_sqrt[mask] = np.sqrt((Y**2 - R**2)[mask])
 
     I_isqrt = np.zeros(R.shape)
     I_isqrt[mask] = 1./I_sqrt[mask]
+    
+    mask2 = ((II>JJ-2) & (II < JJ+1))
+    print (mask2)
```
stggh: delete legacy trace print

abel/direct.py
```diff
@@ -73,7 +74,7 @@ def _construct_r_grid(n, dr=None, r=None):
 
 def direct_transform(fr, dr=None, r=None, direction='inverse',
                      derivative=gradient,
-                     int_func=simpson_rule_wrong,
+                     int_func=np.trapz,
```
stggh: Perhaps, the change to `int_func` should occur at the source level?
`def _pyabel_direct_integral(f, r, correction, int_func=simpson_rule_wrong):` and delete the redundant `simpson_rule_wrong()`. This code may be placed in the `PyAbelLegacy` repo

stggh:
Wahoo! A good start to 2018 @DanHickstein. Your change to `trapz` appears to work fine. 

Should the legacy `simpson_rule_wrong`and references be deleted?

Also, update the inline docs

stggh:
travis-ci build failing due to a disk space issue. Note for #201 I switched to `pip wheel` to bulid 2.7. However. the disk space error may be unrelated: `.travis.yml`:
```
install:
- pip install --upgrade pip setuptools wheel
- pip install --only-binary=numpy,scipy numpy scipy
- pip install coveralls
- python setup.py install
script:
```

abel/direct.py
DanHickstein: I think that it's fine to delete the simpson_rule_wrong function. `scipy.integrate.simps` can also handle both even and odd sizes, so I'm unsure exactly why this function is needed.

abel/direct.py
```diff
@@ -107,6 +89,13 @@ def direct_transform(fr, dr=None, r=None, direction='inverse',
         if True in addition to the integration with the Simpson rule,
         integration near the singular value is done analytically,
         assuming a piecewise linear data.
+    backend : string
```
stggh: missed the `f` -> `r` on line 77:
>f : 1D ndarray
        the spatial mesh (optional)

DanHickstein: ah, okay, got it now.

DanHickstein:
I think you are correct about `reflect_array()`. I'll go ahead and get rid of it.

DanHickstein:
Same thing for `abel_sym` too it seems.

stggh:
Agreed.

stggh:
`tests/test_direct.py` remove `test_simps_wrong()`.

DanHickstein:
Yeah, I was just removing that, and then it sent me spiraling down the hole of trying to get the C/Python agreement test to work. Seems like I'm still off by one pixel on the high-r-value side (which is the side that typically doesn't matter much, but I'd still like to get it to work...)

DanHickstein:
Phew! I emerged from the rabbit hole! 

Direct-C just needed a small tweak to match Direct-Python for the edge case of `r=N`.

I unleashed the previously disabled unit tests. We now have agreement between the Direct-C and Direct-Python methods better than `1e-9`, and the Direct methods reproduce a 501-point gaussian to within 2%.

stggh:
Impressive!  

The build fails due to a missing miniconda package (for 2.7):
```
0.28s$ conda config --set always_yes yes --set changeps1 no
  File "/home/travis/miniconda2/bin/conda", line 11, in <module>
    sys.exit(main())
  File "/home/travis/miniconda2/lib/python2.7/site-packages/conda/cli/main.py", line 125, in main
    from ..exceptions import conda_exception_handler
  File "/home/travis/miniconda2/lib/python2.7/site-packages/conda/exceptions.py", line 17, in <module>
    from .common.io import timeout
  File "/home/travis/miniconda2/lib/python2.7/site-packages/conda/common/io.py", line 5, in <module>
    from concurrent.futures import ThreadPoolExecutor
ImportError: No module named concurrent.futures
```

which may be "fixed" by using the [pip wheel above](https://github.com/PyAbel/PyAbel/pull/204#issuecomment-354680980)  ;-)

DanHickstein:
I'll give that a shot now...

DanHickstein:
Hmmm, when I switch to your code in travis.yml, then it seems that Cython is not installed.

stggh:
One very minor English spelling correction `abel/tests/test_direc.py`  correspondance -> correspondence, although sometimes it is nice to keep the author spelling (@rth) to showcase the multinational nature of `PyAbel`

DanHickstein:
Okay, I made that change. And now I included cython in the pip install, which seems to work.

Now, the Cython extensions are built, but for some reason Travis thinks that the Direct-C method is not available and skips those tests.

abel/direct.py
rth: I agree that it's better to use the trapz/simps functions from scipy if possible.

abel/lib/direct.pyx
```diff
@@ -46,48 +46,38 @@ cpdef _cabel_direct_integral(double [:, ::1] f, double [::1] r, int correction):
...
+        
+                s = 0
+            
+                # simple left Reimann sum:
+                # for k in range(j+1, N1): # inner loop over elements such as r < y
+                #   s = s + f[i,k] * I_isqrt[j,k] * dr
+            
+                # Trapezoidal rule integration, skipping r == y
+                for k in range(j+1, N1-1): # inner loop over elements such as r < y
+                    s = s + f[i,k] * I_isqrt[j,k] * 0.5 * dr  + f[i,k+1] * I_isqrt[j,k+1] * 0.5 * dr
```
rth: maybe use inplace add and factorize multiplications?
```py
s += (f[..] * I_isqrt[..] + f[..] * I_isqrt[..]) * 0.5 * dr
```

abel/tests/test_direct.py
```diff
@@ -8,109 +8,77 @@
...
-    recon = abel.direct.direct_transform(x, direction="inverse")
-
+    recon = abel.direct.direct_transform(x, direction='inverse')
     assert recon.shape == (n, n)
 
 
 def test_direct_zeros():
     # just a sanity check
-    if not abel.direct.cython_ext:
-        raise SkipTest
```
rth: This was intended to skip this test if the cython implemented failed to compile. We don't want this anymore?

abel/direct.py
```diff
@@ -224,45 +222,3 @@ def is_uniform_sampling(r):
...
-    global_assumptions.add(Q.is_true(r2 > y))
-    global_assumptions.add(Q.is_true(r2 > r1))
-    P = c0 + (r-y)*c_r  # + (r-r0)**2*c_rr
-    K_d = 1/sqrt(r**2-y**2)
-    res = integrate(P*K_d, (r, y, r1))
-    sres = simplify(res)
-    print(sres)
-
-
-def reflect_array(x, axis=1, kind='even'):
```
rth: I think we still have code depending on this function in my former lab, so I would rather keep it if possible to avoid breaking things.

stggh: Perhaps, the function may be added to `abel/tools/symmetry.py` ?

stggh:
@DanHickstein once addressing @rth's concerns, this PR is ready to merge. Feel free to do so, and I will rebase PR #201, the code changes in which are not really that much.

abel/lib/direct.pyx
stggh: Very clever!

abel/direct.py
rth: @stggh Sounds good to me.

abel/lib/direct.pyx
DanHickstein: Sounds good to me!

abel/direct.py
DanHickstein: I made the move to `tools.symmetry.reflect_array`. But basically this is just a wrapper for np.flipud and np.fliplr?

If you really don't want to break the code in your previous lab, should we leave the function exactly where it is and include a depreciation warning?

DanHickstein:
Hmmm, the Cython extensions seem to build just fine, but then, when running on Travis, it seems like they cannot be found. I'm not quite sure what's going on. I'll try switching back to the previous Travis.yml

DanHickstein:
Well, I think that the issue on Travis has something to do with where the Cython extensions are put after they are built.

Switching `python setup.py install` to `python setup.py develop` fixed the problem. 

Also, using 
```
python setup.py build_ext -i
python setup.py install
```
works. The first line is to "build extension inplace".

Probably we need to change something about how the extensions are built in the setup.py to ensure that they are properly installed on linux with just `python setup.py install`.

abel/direct.py
rth: OK, I checked, a copy is still available in the [`luli/hedp`](https://github.com/luli/hedp/blob/master/hedp/math/abel.py#L107) repo, so feel free to either remove it or move it under `tools.symmery`, as you prefer. Thanks for fixing this issue, in general!

rth:
> Probably we need to change something about how the extensions are built in the setup.py to ensure that they are properly installed on linux with just python setup.py install

I'll try to have a look once this PR is merged..

abel/direct.py
DanHickstein: Okay, I think that I will remove the function. Thanks!

DanHickstein:
Thanks for the input @rth and @stggh! Are there further suggestions, or should I go ahead and merge later today?

stggh:
Looks good to me.

BTW, on my Linux system `python3 setup.py install --user` builds and finds the cython stuff.
```
nosetests  abel/tests/  --verbosity=2  --with-coverage --cover-package=abel
  :
Ran 40 tests in 3.742s

OK
```

and `direct` `backend='c'` works, so the installation [method issue](https://github.com/PyAbel/PyAbel/pull/204#issuecomment-354845103) may be fine.

rth:
Merged. @DanHickstein thanks again for fixing this and @stggh for the review!

DanHickstein:
Thanks guys!

@stggh, does the install still work if you omit the `--user` bit?

stggh:
Yes

DanHickstein:
Hmm. Okay, maybe that issue is unique to Travis. That's weird. But I suppose it means that we don't need to fix it.

PR #206, opened 2018-01-23
hansenlaw -0.3 pixel offset

stggh:
This (simple) PR addresses issue #159 sub-pixel alignment for the `hansenlaw` method:

The offset was due to an incorrect definition of the normalized radius <math>(N-n)/(N-n-1)</math>

Discovered during a code clean up for execution speed enhancement, namely, the replacement of `np.power()` by the faster `**`

I have also changed:
1) `np.gradient` call with the `axis=-1` parameter, eliminating the image shape test for 2D
2) a couple variable names to match those used in the Hansen & Law paper: `Nm -> N`, `nn -> n`
3) removed the -0.3 pixel grid shift/interpolation

Bonus, inverse Abel transform now takes the same execution time as the forward transform

### DIFF
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -5,8 +5,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import numpy as np
-from scipy.ndimage.interpolation import shift
-from math import exp, log, pow, pi
 
 #############################################################################
 # hansenlaw - a recursive method forward/inverse Abel transform algorithm
@@ -103,33 +101,28 @@ def hansenlaw_transform(IM, dr=1, direction="inverse", **kwargs):
     """
 
     IM = np.atleast_2d(IM)
-    N = np.shape(IM)         # shape of input quadrant (half)
-    AIM = np.zeros(N)        # forward/inverse Abel transform image
-
-    rows, cols = N
+    rows, cols = np.shape(IM)      # shape of input quadrant (half)
+    AIM = np.zeros_like(IM)        # forward/inverse Abel transform image
 
     # Two alternative Gamma functions for forward/inverse transform
     # Eq. (16c) used for the forward transform
-    def fgamma(Nm, lam, n):
-        return 2*n*(1-np.power(Nm, lam+1))/(lam+1)
+    def fgamma(N, lam, n):
+        lam += 1
+        return 2*n[::-1]*(1 - N**lam)/lam
 
     # Eq. (18) used for the inverse transform
-    def igammalt(Nm, lam, n):
-        return (1-np.power(Nm, lam))/(pi*lam)
+    def igammalt(N, lam, n):
+        return (1 - N**lam)/(np.pi*lam)
 
-    def igammagt(Nm, lam, n):
-        return -np.log(Nm)/pi
+    def igammagt(N, lam, n):
+        return -np.log(N)/np.pi
 
     if direction == "inverse":   # inverse transform
         gammagt = igammagt   # special case lam = 0.0
         gammalt = igammalt   # lam < 0.0
 
         # g' - derivative of the intensity profile
-        if rows > 1:
-            gp = np.gradient(IM)[1]
-            # second element is gradient along the columns
-        else:  # If there is only one row
-            gp = np.atleast_2d(np.gradient(IM[0]))
+        gp = np.gradient(IM, axis=-1)
 
     else:  # forward transform
         gammagt = gammalt = fgamma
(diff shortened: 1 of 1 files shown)


abel/hansenlaw.py
```diff
@@ -103,33 +101,28 @@ class :class:``abel.Transform``.
...
-        return 2*n*(1-np.power(Nm, lam+1))/(lam+1)
+    def fgamma(N, lam, n):
+        lam += 1
+        return 2*n[::-1]*(1 - N**lam)/lam
 
     # Eq. (18) used for the inverse transform
-    def igammalt(Nm, lam, n):
-        return (1-np.power(Nm, lam))/(pi*lam)
+    def igammalt(N, lam, n):
+        return (1 - N**lam)/(np.pi*lam)
```
DanHickstein: Woah, it turns out that `numpy.power` can be significantly slower than `**`. Good to know!

https://stackoverflow.com/questions/25870923/how-to-square-or-raise-to-a-power-elementwise-a-2d-numpy-array

DhrubajyotiDas: What about np.square(x) as one of the comments in the SO link states?

stggh:
`lam` is not integer 

@DhrubajyotiDas my apologies for not waiting for your input. I figured that you might be busy.

DhrubajyotiDas:
No worries - if @DanHickstein is okay with it, so am I. As is, Hansenlaw is outside my wheelhouse.

stggh:
Sadly, I think the `(0, -0.3)` shift may still be needed. Investigating ...

stggh:
Monkey writing Shakespeare ...

The Hansen & Law paper states:
  
  image

For `hansenlaw` if the number of columns is set to `0, ..., N-1`, instead of `N-2`, the inverse transform agrees with the `direct` inverse method, and better with the source function curveA = `profile3`. No `-0.3` pixel shift required.
Before (lines 141-148, note slight change for some variable names):
```python
    n = np.arange(N-1)  # n = 0, ..., N-2

    Nratio = (N - n)/(N - n - 1)  # R0/R

    X = np.zeros((K, rows))
    Gamma = np.zeros((N-1, K, 1))
    Phi = np.zeros((N-1, K, K))

```
curvea_n-1
After:
```python
    n = np.arange(N)  # n = 0, ..., N-1

    n[-1] = n[-2] # to prevent divide by zero in ratio
    Nratio = (N - n)/(N - n - 1)  # R0/R

    X = np.zeros((K, rows))
    Gamma = np.zeros((N, K, 1))  # size increased by 1
    Phi = np.zeros((N, K, K))  # size increased
```

 curvea_n

For `n = np.arange(1, N+1)` `hansenlaw` is above the projection function. 

The plots above are for `N=101` for `N=1001` both `direct` and `hansenlaw` have better agreement.
curvea_n

This may be close enough. However, both `hansenlaw` and `direct` may benefit from a closer examination of the `R0/R` grid.

More playing ...

stggh:
The offset, by 1 column improvement, above rests with the gradient function (~ line 164)
```python
X = np.dot(Phi[i], X) + Gamma[i]*gp[:, ni]
```
where for the inverse transform `gp = np.gradient(IM, axis=-1)` (line 126)

If the gradient image is shifted across by `-0.43` pixels `gp = shift(gp, (0, -0.43))` it improves the agreement. This is `n=101`, `hansenlaw` orange dashes:
curvea

I believe `direct` method uses its own gradient function, `abel.tools.math.gradient()` which has a parameter `x`:
```python
def gradient(f, x=None, dx=1, axis=-1):
    :
    x : array_like, optional
       Points where the function f is evaluated. It must be of the same
       length as ``f.shape[axis]``.
       If None, regular sampling is assumed (see dx)
```
Unfortunately, AFAICT `x` does not affect the evaluation of the derivative, if the spacing is unchanged. Otherwise, it might be useful to offset the gradient (phase) slightly, for both the `hansenlaw` and `direct` methods @rth?

stggh:
Actually, the shift should be easy using linear interpolation, something like: `(gp[:, 1:]+gp[:, :-1])/2` for `-0.5` shift.

stggh:
Still looking at this. For the `direct` method it is all a matter of degree, the improvement for small radius is offset by a poorer fit for high `r`. From what I can tell the `direct` code should stay as it is.

For `hansenlaw` starting the inverse transform one pixel higher improves the agreement for `curveA` == `profile3`, and this becomes better if the gradient function is further shifted by a fraction of a pixel ~0.5. In general the `hansenlaw` inverse transform profile is displaced from all the other methods (by ~-0.3 pixels, which I thought I had "fixed" with the last PR). The gradient shift results in an equivalent image (grid) shift, without the time cost of interpolating the final image. 

I have run a large number of tests. The following plots include the gradient shift for `hansenlaw`, the `direct` code is unchanged:
1) Curve A (profile3):
curvea
2) Step
step
3) `profile5` which `hansenlaw` does not handle very well, and the gradient displacement makes worse.  `three_point`, of course, does very well:
profile5
4) All the transform pair functions (inverse). Note, here the plot is the % difference from the source profile. In general, the disagreement is small:
transformpairs
5) All the transform pair functions (forward):
transformpairs-forward
6) O<sub>2</sub><sup>-</sup> photodetachment:
o2
7) Simulated O<sup>-</sup> spectrum:
o
In this case I am not sure why the source spectrum does not overlap with the transform methods. It may be the initial forward transform, or the angular integration process (investigating).

Hansen & Law Fig 3. indicates that the `curve A` inverse transform is exact:
image

This is only achieved in the current code if the gradient is slightly shifted (linear interpolation):
```python
        # shift gradient curve (0, -phase) pixels. Improves agreement with
        # analytical transform pairs, see PR #206
        phase = 0.43
        gp[:, :-1] = gp[:, 1:]*phase + gp[:, :-1]*(1-phase)
```
I think this may be good enough ...

Given the excellent transform from `three_point` there is little to be gained in spending a great deal more time trying to perfect `hansenlaw`. The transform pair difference functions indicate both `hansenlaw` and `direct` inverse transforms are close, but not as perfect as `three_point`. Something, possibly, associated with their respective gradient functions. However, they are the only methods with a `forward` direction [apart from the upcoming `fourier_hankel` method, the forward transform is working].  

I will play a little more, before submitting a PR to apply the `hansenlaw` gradient shift.

DanHickstein:
Thanks for these extensive tests and tinkering! It seems like some progress has been made. Most people won't be concerned with a sub-pixels shift, but there are always perfectionists out there :)

While three-point has shown really robust performance in our recent tests, hansenlaw has some advantages, including a forward transform, and (especially if you include the pre-computation (basis-set generation) for three point) better scaling for large image sizes.

stggh:
It turns out that it is pretty easy for me to make mistakes with indexing, even for such a simple Hansen&Law algorithm, and my analysis was not helped by believing that the forward transform is 100% correct. A comparison with the output from `direct` and source->projection->source indicated a subpixel displacement in the output of `hansenlaw`, for the forward transform. Part of this was due to using `(N-n)` and not `(N-n-1)` in the first part of the forward gamma function, Eq. (16c).
image

One "fix" is to apply a `-1/2` pixel shift to the gradient (inverse transform) or image (forward transform). This then makes both offsets (either for the gradient or for the image, depending on transform direction) the same value, producing excellent results, with even `profile5` better:
profile5
Curve A (`profile3`):
curvea
O2 and O look great:
o2
o

I think this "fix" is enough. I have a lurking feeling that a correct assignment of grid parameters would solve this issue, without the need to displace the source function(s), but I have not found a way. Possibly something similar applies to `direct` (see the curve A and `profile5` plots).

The main change, apart from correcting `fgamma` (line 111) is to shift `gp` (gradient for the inverse transform, or image for the forward transform) by -1/2 a pixel, before applying the Abel transform code:
```python    # Abel transform, Eq. (15) forward, or (17) inverse
    for i, col in enumerate(nr):
        X = np.dot(Phi[i], X) + Gamma[i]*gp[:, col]
        AIM[:, col] = X.sum(axis=0)
```

stggh:
I have rewritten the `hansenlaw` transform to use a single grid stepping to the left from the right side, instead of stepping out(`i`) and in(`col`), used above:
```python
    # Abel transform ---- Eq. (15) forward, or (17) inverse
    for col in n[::-1]:  # outer (right) edge to inner (left) edge
        X = np.dot(Phi[col], X) + Gamma[col]*gp[:, col]
        AIM[:, col] = X.sum(axis=0)
```
I now find that the forward &gamma;<sub>n</sub> Eq. (16c) needs the prefix <math>2(n+1)</math> rather than <math>2(n-1)</math>. Discovered, using a transform loop  source-(forward)->projection-(inverse)->source and comparing the two sources.
recycle2
The  image (forward) or gradient (inverse) -1/2 pixel offset is still better aligns the transform. All of which hints that my implementation of the grid may still be not quite right. However, this `hansenlaw` in my [HansenLaw-fix fork](https://github.com/stggh/PyAbel/tree/HansenLaw-fix) is better and `2x` faster than the current version in `PyAbel`.

Issue #12, opened 2015-11-17
Change name to avoid conflict with other PyBasex


DanHickstein:
pyBasex is already a pip-installable [python package](https://pypi.python.org/pypi/pyBaseX/v0.1.0-alpha.3). So I am regretting choosing that name in the first place. If someone types "pip install pybasex", then a different package installs, which could be a source of confusion. "Basex" is also already taken, though not as a python package.

So, it might be advisable for us to change the name to something that's not already taken. Thoughts?

rth:
Good point. I'm not sure what is the optimal way of handling this...

 As far as I understand, there are two sides to this problem:
 - conflict on pypi, so `pip` would install the wrong package
 - googling for "basex"  could produce the wrong package
by itself conflict on github is not an issue (or maybe it is case sensitive) since [`lucalianas/pyBaseX`](https://github.com/lucalianas/pyBaseX) exist already anyway. 

### Apporach 1

Regarding pypi, what could be done is to leave the package pyBASEX on github, and upload a module called `basex` un pypi (to take that slot). The module can then be installed as `pip install basex` (since we are using pip (i.e. python) the `py` prefix in the module name is not really useful) . Similar approach is for instance for `pytables` that is installed as `pip install tables`. Ok there might be some confusion about it, but since these 2 modules address a very different audience it shouldn't be that bad. In some ways keeping the word basex in the module name is good, since that's what people would be  searching for I suppose. 

### Approach 2

Or, change the name completely to avoid confusion, with the constraints of the PEP style guide:
> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged. 

 Possible module names I can think of are for instance,  `basex_transform`, `iabel_basex`,  `basis_set_exp`  although none of that is as good as simply `basex`... 

Not sure what would be the best way. Other ideas?

DanHickstein:
Hmmm, one approach would be to just make the name more general, like ``iabel``, ``abel``, or ``pyabel``. These do not seem to be taken. The additional advantage would be that we could potentially implement other version of the inverse abel transform in the future within the same package. For example, the Polar Onion Peeling algorithm is popular, and I would like to see a python implementation of this for my own research. There is a nice [matlab POP implementation by Adi Natan](http://www.mathworks.com/matlabcentral/fileexchange/41064-polar-onion-peeling), so porting could be pretty easily if he was willing to let us adapt his code.

[...]

PR #208, opened 2018-01-23
linbasex: remove the basis calculation

stggh:
@DanHickstein  PyAbelPaper [#12](https://github.com/DanHickstein/PyAbelPaper/issues/12) (developer access only)
> In our benchmarks, the computation time for the Lin-BASEX transform is significantly longer than the      basis set generation time. So, this leads me to believe that Lin-BASEX is not a method that benefits from pre-computation

>[@stggh] Yes, in fact, it may be better to drop the pre-calculation altogether, since the "basis" changes with the slightest parameter change, generating very long filenames, more easily leading to error.

This PR removes the basis calculation from `linbasex`, which included the call to `Basis = abel.tools.basis.get_bs_cached()`, the `linbasex` specific code in `abel.tools.basis.py`, and the basis timing calculation in `benchmark.py`. 

Good call @DanHickstein,  goodbye to obnoxious/obscure  `linbasex` basis filenames `linbasex_basis_1023_1023_02_050_1_0.npy` etc

plot_example_linbasex

### DIFF
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -54,8 +54,7 @@ class AbelTiming(object):
             'direct_Python': direct.direct_transform,
             'direct_C': direct.direct_transform,
             'hansenlaw': hansenlaw.hansenlaw_transform,
-            'linbasex': linbasex._linbasex_transform_with_basis,
-            'linbasex_bs': linbasex._bs_linbasex,
+            'linbasex': linbasex.linbasex_transform,
             'onion_bordas': onion_bordas.onion_bordas_transform,
             'onion_peeling': dasch.dasch_transform,
             'onion_peeling_bs': dasch._bs_onion_peeling,
@@ -67,7 +66,7 @@ class AbelTiming(object):
 
         # result dicts
         res = {}
-        res['bs'] = {'basex_bs': [], 'linbasex_bs': [], 'onion_peeling_bs': [], 
+        res['bs'] = {'basex_bs': [], 'onion_peeling_bs': [], 
                      'two_point_bs': [], 'three_point_bs': []}
         res['forward'] = {'direct_Python': [], 'hansenlaw': []}
         res['inverse'] = {'basex': [], 'direct_Python': [], 'hansenlaw': [],
--- a/abel/linbasex.py
+++ b/abel/linbasex.py
@@ -177,12 +177,10 @@ def linbasex_transform_full(IM, proj_angles=[0, np.pi/2],
         raise ValueError('image has shape ({}, {}), '.format(rows, cols) +
                          'must be square for a "linbasex" transform')
 
-    # generate basis or read from file if available
-    Basis = abel.tools.basis.get_bs_cached("linbasex", cols,
-                  basis_dir=basis_dir,
-                  basis_options=dict(proj_angles=proj_angles,
-                  legendre_orders=legendre_orders, radial_step=radial_step,
-                  clip=clip, verbose=verbose))
+    # generate basis
+    Basis =  _bs_linbasex(cols, proj_angles=proj_angles,
+                          legendre_orders=legendre_orders,
+                          radial_step=radial_step, clip=clip)
 
     return _linbasex_transform_with_basis(IM, Basis, proj_angles=proj_angles,
                                     legendre_orders=legendre_orders,
--- a/abel/tests/test_linbasex.py
+++ b/abel/tests/test_linbasex.py
@@ -16,7 +16,7 @@ def test_linbasex_shape():
     n = 21
     x = np.ones((n, n), dtype='float32')
 
-    recon = abel.linbasex.linbasex_transform_full(x, basis_dir=None)
+    recon = abel.linbasex.linbasex_transform_full(x)
 
     assert recon[0].shape == (n+1, n+1)   # NB shape+1
 
--- a/abel/tools/basis.py
+++ b/abel/tools/basis.py
@@ -14,21 +14,14 @@ def get_bs_cached(method, cols, basis_dir='.', basis_options=dict(),
 
     Checks whether file ``{method}_basis_{cols}_{cols}*.npy`` is present in 
     `basis_dir` 
-    (*) special case for ``linbasex``
-         _{legendre_orders}_{proj_angles}_{radial_step}_{clip} 
-        where {legendre_orders} = str of the list elements, typically '02'
-              (proj_angles} = str of the list elements, typically '04590135'
-              {radial_step} = pixel grid size, usually 1
-              {clip} = clipping size, usually 0
 
     Either, read basis array or generate basis, saving it to the file.
-        
 
     Parameters
     ----------
     method : str
-        Abel transform method, currently ``linbasex``, ``onion_peeling``,
-        ``three_point``, and ``two_point``
+        Abel transform method, currently ``onion_peeling``, ``three_point``,
+        and ``two_point``
     cols : int
         width of image
     basis_dir : str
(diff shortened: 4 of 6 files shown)


DanHickstein:
@stggh, thanks for looking into this! 

I was originally thinking that you were just going to change the default behavior of linbasex so that the basis sets would not be saved to disk, but that the option would still be available if necessary. Here, it seems that you are removing the capability to save the basis sets altogether.

In our tests, it seemed that the basis set generation was a small fraction of the total transform time. Are we fairly confident that is always the case? If so, then it seems okay to go ahead with this.

DhrubajyotiDas:
Would it be possible to make the default implementation for linbasex be such that the basis set is not calculated and/or saved instead of "amputating" the feature from the codebase altogether?

DanHickstein:
@DhrubajyotiDas and I are thinking along the same lines.

However, if the feature really adds nothing, then it should be amputated.

stggh:
I was never happy with the number of parameters required to uniquely specify the `linbasex` basis, generating many odd filenames:
```
{method}_basis_{cols}_{cols}_{legendre_orders}_{proj_angles}_{radial_step}_{clip}.npy
```
The projection angles are converted to % of &pi;, all very awkward e.g. see `examples/example_linbasex.py` where `proj_angles = np.arange(0, 2*np.pi, np.pi/20)` which creates the basis file (if basis_dir changed from None): `linbasex_basis_1023_1023_02_05101420252935404550555965707580859095100105110115119124130135140145150155160164170174180185190195_1_0.npy`
and possibly not unique, so it was highly likely that a wrong basis file would be read-back, and the user not know.

However! `linbasex` basis storage:
```

Implementation      n = 301           n = 501       
----------------------------------------------------------
        basex_bs    6119.3           19379.0         
     linbasex_bs       5.1              13.5         
onion_peeling_bs       2.5               5.7         
  three_point_bs       9.2              24.4         
    two_point_bs       3.3               8.1         


Implementation      n = 301           n = 501       
----------------------------------------------------------
        direct_C      18.1              77.7         
   direct_Python     187.7             845.8         
       hansenlaw       5.8              15.3         


Implementation      n = 301           n = 501       
----------------------------------------------------------
           basex       2.1              19.8         
        direct_C      19.3              80.2         
   direct_Python     190.0             874.9         
       hansenlaw       6.3              16.2         
        linbasex      35.3              95.4         
    onion_bordas     288.5             908.0         
   onion_peeling       0.6               1.8         
     three_point       0.4               2.0         
       two_point       0.4               1.6     
```

After, no `linbasex` basis stored:

```

Implementation      n = 301           n = 501       
----------------------------------------------------------
        basex_bs    6284.7           19181.0         
onion_peeling_bs       2.7               6.7         
  three_point_bs       9.8              24.8         
    two_point_bs       3.5               9.0         


Implementation      n = 301           n = 501       
----------------------------------------------------------
        direct_C      18.1              77.8         
   direct_Python     277.7             852.5         
       hansenlaw       5.8              15.2         


Implementation      n = 301           n = 501       
----------------------------------------------------------
           basex       2.4              19.8         
        direct_C      18.4              79.6         
   direct_Python     280.9             853.9         
       hansenlaw       6.3              16.3         
        linbasex     161.2             489.6         
    onion_bordas     287.7             899.2         
   onion_peeling       0.5               1.8         
     three_point       0.4               1.5         
       two_point       0.4               1.5         
```

Hmm, you have a point!  The 14% basis calculation cost gets x4 because of the full image transformation.  I will implement your suggestion. The `linbasex` basis file just needs to be used with caution.

DanHickstein:
That doesn't seem to add up. Before, the basis set calculation took 13 msec (for the 501 case) and the transform took 95 msec. After, the transfom takes 500 msec? That seems like a 40x penalty...

DhrubajyotiDas:
>`linbasex_basis_1023_1023_02_05101420252935404550555965707580859095100105110115119124130135140145150155160164170174180185190195_1_0.npy`

I had no idea it was this bad - with plain old basex these parameters are true integers, so the file name was relatively easily resolvable.

stggh:
> I had no idea it was this bad - with plain old basex these parameters are true integers

`linbasex` has many more parameters:

 `{method}_basis_{cols}_{cols}_{legendre_orders}_{proj_angles}_{radial_step}_{clip}.npy`


> That doesn't seem to add up

No basis storage:
```
abel.benchmark.AbelTiming(n=[1023], select=['linbasex'])
Implementation      n = 1023      
        linbasex    4188.4  

In [x]: IM = np.loadtxt("O2-ANU1024.txt.bz2")
In [x]: IMc = abel.tools.center.center_image(IM, center='convolution', square=True)
In [8]: IMc.shape
Out[8]: (1023, 1023)

In [9]: %timeit abel.Transform(IMc, method='linbasex')
557 ms ± 15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Basis storage:
```
In [1]: import abel

In [2]: abel.benchmark.AbelTiming(n=[1023], select=['linbasex'])
Out[2]: 

Implementation      n = 1023      
     linbasex_bs      66.8         

Implementation      n = 1023      
        linbasex     555.1         

In [3]: import numpy as np

In [4]: IM = np.loadtxt("O2-ANU1024.txt.bz2")

In [5]: IMc = abel.tools.center.center_image(IM, center='convolution', square=Tr
   ...: ue)

In [6]: %timeit abel.Transform(IMc, method='linbasex')
556 ms ± 4.75 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [7]: %timeit abel.Transform(IMc, method='linbasex', transform_options=dict(ba
   ...: sis_dir='.'))
505 ms ± 3.61 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I agree it does not add up, however, the presence of the basis file does improve execution time. Academic, further investigation required.

I will keep the basis storage, and make the default for `linbasex` `basis_dir=None`

stggh:
Investigating the timing difference ...

stggh:
Ok, the timing issue was in `benchmark.py` (only), where I had changed `'linbasex': linbasex._linbasex_transform_with_basis` to `'linbasex': linbasex.linbasex_transform` the latter constructs a full (4x) image from the input and hence the `benchmark` timing blow-out.

`linbasex` requires a full image to fit the basis expansion, hence a quadrant transform is reassembled into a full image before applying the `linbasex` method. `benchmark` "cheats" by passing a quadrant which is treated as a full image by `linbasex`. In this sense, `basex` is penalized since it internally generates a full image:  `full_image = np.hstack((np.fliplr(data), data[:, 1:]))` (line 119).

This issue comes back to @DanHickstein's discussion on what `benchmark` should measure. Perhaps, it should pass each method a full image?

stggh:
Just to clarify my thoughts on the `benchmark.py` timing discussion:

`abel.method.method_transform()` transforms the right-side of an image. `basex` and `linbasex` construct a full-image before evaluating the Abel transform.  Hence, these methods are disadvantaged by `benchmark.py`, because they are performing 2x-4x the operations.

`abel.Transform()` process a full-image according to the `symmetry_axis`, if  `=None` 4-image quadrants are transformed, and the image reassembled. `symmetry_axis=(0, 1)` results in only a single quadrant transform, but still a full-image assembly. There is a special test case for `linbasex` to use the `full_image` transform if a symmetry axis is present. `basex` could use something similar. Using `abel.Transform()` for timing, in the current form, would disadvantage `basex` and AFAIK there is no easy way to extract the basis calculation timing.

`benchmark.py` could  x0.5 (`basex`) and x0.25 (`linbasex`) times to provide relative timings. All timings could be x2 to provide "real" image (size `2nx2n`) transform times.

Issue #207, opened 2018-01-23
Benchmarks should operate on square whole images


DanHickstein:
Currently, abel.benchmark.AbelTiming passes, for example, a `101x101` pixel *half-image* to each transform method. This corresponds to a `101x201` pixel *whole-image* when we apply the cylindrical (left-right) symmetry.

In most experimental situations, people are recording both sides of the projection (using a 512x512 pixel camera, for example), and would want benchmarks for the *square* whole-image. Therefore, I suggest that we modify benchmark.py to pass a rectangular half-image to each transform method. 

Specifically, I believe that modifying line 107 of abel/benchmarks.py from:
```python
x = np.random.randn(ni, ni)
```
to:
```python
x = np.random.randn(ni, np.floor(ni * 0.5) + 1)
```
should do the trick.

This has the added advantage of making all of the benchmarks seem a factor-of-two faster :). But, I think that most previous benchmarks use this convention. For example, when [Rallis et al](http://krex.k-state.edu/dspace/bitstream/handle/2097/18907/BenItzhakRevSciInst2014.pdf;sequence=1) say that they can transform a 1024x1024 image in 0.6 seconds, they are referring to the whole image. In PyAbel, we would be passing a 1024x512 half-image to the individual transform method. 

One could argue that, for the photoelectron distributions in Rallis et al, up/down symmetry could additionally be applied, and the transform could be performed on a 512x512 pixel image. However, this symmetry is not generic to the Abel transform, which only assumes cylindrical symmetry. Additionally, one could argue that, since the default behavior of `abel.Transform` is to allow for asymmetric transforms, where each quadrant is transformed separately, then the result of `abel.benchmark.AbelTiming` with n=101, actually corresponds to 25% of the time it would take to perform the transform on a 201x201 pixel image (which is not necessarily equal to the time it takes for a 101x101 image). 

Thoughts?

stggh:
You are opening a can of worms:
1)  `linbasex` does require a square image, however, it is a special case in terms of the methods. 
2) I believe the basis methods form a whole image for processing a single quadrant in any case(*)

In general symmetry considerations improve the execution time by the symmetry divisor. Here are some `3pt` timings for a `8192x8192` VMI (don't worry about the even size "camera" image)
```python
In [1]: import numpy as np

In [2]: import abel

In [3]: IM = np.loadtxt("O2-ANU8192.txt")

In [4]: Q = abel.tools.symmetry.get_image_quadrants(IM)

[...]

PR #209, opened 2018-01-31
Changing the benchmarks to operate on square whole images

DanHickstein:
As discussed in https://github.com/PyAbel/PyAbel/issues/207, I think that it makes the most sense for the benchmarks to operate on square whole-images that possess left/right symmetry. Prior to this PR, `abel.benchmark.AbelTiming(n=[101])` completes a 101x101 transform, which actually corresponds to a transform of a left/right symmetric 101x201 image.

This PR changes that behavior, and completes a transform on the 101x51 half-image. It should make most of the benchmarks faster, since they are operating on a smaller image. BASEX and LinBasex operate on whole images, so they are passed the complete image and maintain their previous benchmark speeds.

### DIFF
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -104,27 +104,39 @@ class AbelTiming(object):
         # ---- timing tests for various image sizes nxn
         for ni in n:
             ni = int(ni)
-            x = np.random.randn(ni, ni)
-           
+            h, w = ni, ni//2 + 1
+            # We transform a rectangular image, since we are  making the assumption 
+            # that we are transforming just the "right side" of a square image.
+            # see: https://github.com/PyAbel/PyAbel/issues/207
+            
+            half_image  = np.random.randn(h, w)    
+            whole_image = np.random.randn(h, h)
             # basis set evaluation --------------
             basis = {}
             for method in res['bs'].keys():
-                if method[:-3] == 'basex':  # special case
-                    if ni <= n_max_bs:
+                if ni <= n_max_bs:
+                
+                    if method[:-3] == 'basex':  # special case
                         # calculate and store basex basis matrix
                         t = time.time()
-                        basis[method[:-3]] = transform[method](ni, ni,
-                                                       basis_dir=None)
+                        basis[method[:-3]] = transform[method](ni, ni, basis_dir=None)
+                        res['bs'][method].append((time.time()-t)*1000)
+                    
+                    elif method[:-3] == 'linbasex':  # special case
+                        t = time.time()
+                        basis[method[:-3]] = transform[method](ni)
                         res['bs'][method].append((time.time()-t)*1000)
                     else:
-                        basis[method[:-3]] = None,
-                        res['bs'][method].append(np.nan)
+                        # calculate and store basis matrix
+                        t = time.time()
+                        # store basis calculation. NB a tuple to accomodate basex
+                        basis[method[:-3]] = transform[method](w), 
+                        res['bs'][method].append((time.time()-t)*1000)
+                        
                 else:
-                    # calculate and store basis matrix
-                    t = time.time()
-                    # store basis calculation. NB a tuple to accomodate basex
-                    basis[method[:-3]] = transform[method](ni), 
-                    res['bs'][method].append((time.time()-t)*1000)
+                    basis[method[:-3]] = None,
+                    res['bs'][method].append(np.nan)
+               
 
             # Abel transforms ---------------
             for cal in ["forward", "inverse"]:
@@ -132,10 +144,26 @@ class AbelTiming(object):
                     if method in basis.keys():
                         if basis[method][0] is not None:
                             # have basis calculation
-                            res[cal][method].append(Timer(
-                               lambda: transform[method](x, *basis[method])).
-                               timeit(number=transform_repeat)*1000/
-                               transform_repeat)
+                            if method == 'basex':
+                                # pass a whole image to BASEX
+                                res[cal][method].append(Timer(
+                                   lambda: transform[method](whole_image, *basis[method])).
+                                   timeit(number=transform_repeat)*1000/
+                                   transform_repeat)
+                             
+                            elif method == 'linbasex':
+                                 # pass a whole image to linbasex
+                                 # also, don't unpack the basis set
+                                 res[cal][method].append(Timer(
+                                    lambda: transform[method](whole_image, basis[method])).
+                                    timeit(number=transform_repeat)*1000/
+                                    transform_repeat)       
+                            else:
+                                
+                                res[cal][method].append(Timer(
+                                   lambda: transform[method](half_image, *basis[method])).
+                                   timeit(number=transform_repeat)*1000/
+                                   transform_repeat)
                         else:
                             # no calculation available
                             res[cal][method].append(np.nan)
@@ -144,14 +172,14 @@ class AbelTiming(object):
                             res[cal][method].append(np.nan)
                         else:     
                             res[cal][method].append(Timer(
-                               lambda: transform[method](x, backend=method[7:],
+                               lambda: transform[method](half_image, backend=method[7:],
                                direction=cal)).
                                timeit(number=transform_repeat)*1000/
                                transform_repeat)
                     else:
                         # full calculation for everything else
                         res[cal][method].append(Timer(
-                           lambda: transform[method](x, direction=cal)).
+                           lambda: transform[method](half_image, direction=cal)).
                            timeit(number=transform_repeat)*1000/
                            transform_repeat)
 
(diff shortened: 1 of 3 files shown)


DhrubajyotiDas:
I am OK with this strategy. I believe the reason BASEX does the whole image and not two halves is it essentially averages both sides simultaneously by trying to fit a symmetric exponential*ish* curve to the corresponding points on either side of the centerline. I believe some clever coding changes could make it such that we are fitting a half-exponential to a two-sided average of the raw image (pretend the image is folded on the centerline first and then a half exponential is fit to the one side), and that *might* boost BASEX speed.

Or that's what's already happening now and it would make no difference. I'm not sure.

abel/benchmark.py
```diff
@@ -104,38 +104,66 @@ def __init__(self, n=[301, 501], select=["all", ], n_max_bs=700,
         # ---- timing tests for various image sizes nxn
         for ni in n:
             ni = int(ni)
-            x = np.random.randn(ni, ni)
-           
+            h, w = ni, ni//2 + 1
```
stggh: This is good, a 1/2 sized odd-image is even, but the correct column size is not printed in the benchmark output
```
Implementation      n = 501       
```

DanHickstein: So, for n=501, we are passing `three_point` (for example) a `h=501, w=251` half-image. This is what most users would employ if they had a `501x501` image from their CCD. So, I'm calling this the "501x501" image. Does that seem reasonable?

stggh:
Good work @DanHickstein:

1) In principle the `basex` and `linbasex` time should be divided by ~2.
2) The benchmark listing uses `n`, when it should be `hxw`.
3) The following compares the old and new. They look as expected.
```
In [2]: abel.benchmark.AbelTiming(n=[501])                      In [2]: abel.benchmark.AbelTiming(n=[501])
Out[2]:                                                         Out[2]:
PyAbel benchmark run on x86_64                                  PyAbel benchmark run on x86_64

time in milliseconds                                            time in milliseconds

=========    Basis Abel implementations ==========              =========    Basis Abel implementations ==========

Implementation      n = 501                                     Implementation      n = 501
----------------------------------------                        ----------------------------------------
        basex_bs   19335.7                                              basex_bs   19750.6
     linbasex_bs      12.8                                           linbasex_bs      13.1
onion_peeling_bs       6.9                                      onion_peeling_bs       1.7
  three_point_bs      25.3                                        three_point_bs       6.2
    two_point_bs       8.4                                          two_point_bs       2.1


=========  Forward Abel implementations ==========              =========  Forward Abel implementations ==========

Implementation      n = 501                                     Implementation      n = 501
----------------------------------------                        ----------------------------------------
        direct_C      77.7                                              direct_C      22.5
   direct_Python     919.7                                         direct_Python     229.0
       hansenlaw      14.6                                             hansenlaw       7.2


=========  Inverse Abel implementations ==========              =========  Inverse Abel implementations ==========

Implementation      n = 501                                     Implementation      n = 501
----------------------------------------                        ----------------------------------------
           basex      22.4                                                 basex      22.4
        direct_C      80.3                                              direct_C      22.3
   direct_Python     937.6                                         direct_Python     236.8
       hansenlaw      16.7                                             hansenlaw       7.5
        linbasex      98.5                                              linbasex     103.5
    onion_bordas     912.0                                          onion_bordas     241.3
   onion_peeling       2.9                                         onion_peeling       0.9
     three_point       2.7                                           three_point       0.7
       two_point       2.7                                             two_point       0.7
```
and the results look consistent with individual timings:
```python
In [3]: realimg = abel.tools.analytical.SampleImage(501)

In [4]: proj = abel.Transform(realimg.image, method='hansenlaw', direction='forw
   ...: ard')

In [5]: rows, cols = proj.transform.shape
In [18]: half = proj.transform[:, cols//2:]

In [19]: half.shape
Out[19]: (501, 251)

In [20]: %timeit abel.hansenlaw.hansenlaw_transform(half)
8.81 ms ± 17.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [21]: %timeit abel.direct.direct_transform(half, back_end='C')
22 ms ± 24 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [29]: %timeit abel.linbasex.linbasex_transform_full(proj.transform)
109 ms ± 131 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
All good.

4) My only "beef" (nothing to do with this PR) with `AbelTiming()` is that it makes no attempt to verify the transform, the various sub-components could mangle the transform, and happily, return a fast time. Perhaps, at a later stage we could add a verify method.

My only suggestion for this PR is item 2., that the output table heading be changed from `Implementation      n = 501` to `501x251` e.g. for `n=501`.

abel/benchmark.py
stggh: Only, that it is not the image size that is timed. Perhaps, I am being too pedantic.

stggh:
@DanHickstein feel free to merge, so that I may rebase and PR another `hansenlaw` fix

DanHickstein:
Hmmm, so, maybe we should just clarity the "definition" of the benchmark in the docstring for `AbelTiming` to specify that a n=501 image corresponds to passing a 501x251 image to the individual transform methods.

What I'm attempting to do there is to make our benchmarks consistent with the (few) existing benchmarks from the photoelectron community. [Rallis et al](http://krex.k-state.edu/dspace/bitstream/handle/2097/18907/BenItzhakRevSciInst2014.pdf;sequence=1) gave us this table: 
image

If we want to compare our results to these, then we should not pass a 661x661 pixel image to `three_point`. This would correspond to a 661x1322 image. We should transform a 661x331 pixel half-image to `three_point`, and this corresponds to a 661x661 full image.

So, labeling the table with h=501, w=251 is probably just as confusing as labeling it with just n=501. Maybe we can include some text in the `AbelTiming` docstring, to the extent of:

"The benchmarks assume a left/right symmetric image, and perform the transform on just the right half of the image. For example, or n=501, a h=501, w=251 image is passed to the transform methods. For methods that only support whole image transforms (BASEX, LinBASEX), a whole image is passed."

stggh:
Ok, agreed, the docstring change is sufficient.

We have a rough factor of ~25x speed up:

```python
In [4]: half.shape
Out[4]: (661, 331)

In [5]: %timeit abel.dasch.onion_peeling_transform(half)
1.72 ms ± 40.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [6]: %timeit abel.dasch.three_point_transform(half)
1.74 ms ± 52.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

DanHickstein:
Things sped up by a factor of 25??

I thought that we were seeing a dependence of n^2.5 for most of the transform methods, where n is the width and height. Here, we have cut the width in half. So, I would expect that the speedup factor should be less than 2^2.5 = 5.5. If I had to guess, I would expect that it would be 2^1.5 = 3 times faster.

In a quick test, I see a speedup of about 2.6x for three_point, which seems reasonable:

```python
In [4]: half = np.random.random((501,501))

In [5]: whole = np.random.random((501,501))

In [6]: half = np.random.random((501,251))

In [7]: %timeit abel.dasch.three_point_transform(half)
The slowest run took 36.65 times longer than the fastest. This could mean that an intermediate result is being cached.
100 loops, best of 3: 2.17 ms per loop

In [8]: %timeit abel.dasch.three_point_transform(whole)
The slowest run took 10.12 times longer than the fastest. This could mean that an intermediate result is being cached.
100 loops, best of 3: 5.81 ms per loop
```

Oh, perhaps you meant to type 2.5x?

stggh:
No, I meant relative to the Rallis Table I. onion peeling time of 50(5) ms cf ~2 ms (`PyAbel`) on similar CPU speed system, with same image size `n=661`.

Good size analysis on your part. What I like is that `abel.Transform()` does its job efficiently too.
```python
[etalon:/home/stg121]$ ipython3
Python 3.6.4 |Anaconda custom (64-bit)| (default, Jan 16 2018, 18:10:19) 
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import abel

In [2]: whole = abel.tools.analytical.SampleImage(661).image

In [3]: half = whole[:, 330:]

In [4]: %timeit abel.dasch.three_point_transform(half)
                        
1.66 ms ± 3.21 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [5]: 

In [5]: %timeit abel.dasch.three_point_transform(whole)
4.79 ms ± 3.02 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: %timeit abel.Transform(whole, method='three_point')
8.54 ms ± 5.09 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [7]: 4.79/1.66
Out[7]: 2.885542168674699

In [8]: 8.54/4.79
Out[8]: 1.7828810020876824
```

Issue #210, opened 2018-02-19
Circularization reference


DanHickstein:
@stggh, I received an e-mail from a PyAbel user, saying that they found your circularization algorithm really useful for their research. So, good work on that! 

Am I correct in thinking that the method is basically what you describe in this paper: http://aip.scitation.org/doi/abs/10.1063/1.4981024

If so, you may want to include a reference to that paper in the `abel.tools.circularize_image` docstring. Both to get some citations for your paper, but also to let people know where they can get more information on the method/approach.

stggh:
I appreciate the positive feedback, while also acknowledging the team effort to make the method available.

The algorithm is a simplified version, linear radial scaling vs angle, of what is described in the paper, however, the citation is relevant, and I will include it in the docs.

DanHickstein:
This will be addressed by #211, so closing this issue.

PR #211, opened 2018-02-27
hansenlaw.py code re-write/re-structure

stggh:
This PR that follows on from PR #206, rewriting the `hansenlaw` method code! It is now 20-30% faster and uses less memory.

The original aim was to ensure the pixel grid indices are consistent with the definitions in the Hansen & Law paper:
image
in an effort to track down the source of the pixel alignment offset/shift of the transform

**Outcomes**
1. *consistent index* - mapped index arrays to use a single index, right-side --> left-side 
2. *transform direction* - removed `Gamma` and `Phi` functions, separating the definition of the Gamma arrays according to the transform direction
3. *smaller dimension arrays* - `Phi[cols-1, K, K]` is diagonal for each pixel. Used [numpy broadcasting](https://mail.scipy.org/pipermail/numpy-discussion/2007-March/026809.html) to multiply a smaller 2D diagonal array. Similarly, `Gamma` is reduced to `[cols-1, K]`. Note, I examined storing a pre-calculated "basis" `(Phi, Gamma)`, but there was little to gain in execution speed.
4. *bad* - image pixel shift **still** necessary to align O<sub>2</sub><sup>-</sup> transform with the `three_point` method:

o2-0
o2-0 35

It is not clear what is going on here, I believe `hansenlaw` code/indexing is correct. In the end the `hansenlaw` inverse Abel transform does not produce as exact a result (from the `TransformPairs()`) as `three_point`, so the specifics of this minor pixel shift correction may not be important. I have made it an optional parameter in `hansenlaw`. It is less important for the forward transform which, from a `TransformPair()` comparison, is more accurate.

Some timings:
```
In [1]: import abel

In [2]: abel.benchmark.AbelTiming(n=[501])
Out[2]: 

time in milliseconds

Implementation      n = 501       
        basex_bs   19367.9         
     linbasex_bs      13.0         
onion_peeling_bs       1.6         
  three_point_bs       6.2         
    two_point_bs       2.1         


Implementation      n = 501       
        direct_C      19.9         
   direct_Python     229.3         
       hansenlaw       5.6         


Implementation      n = 501       
           basex      19.4         
        direct_C      21.7         
   direct_Python     232.2         
       hansenlaw       6.0         
        linbasex      94.4         
    onion_bordas     234.3         
   onion_peeling       0.6         
     three_point       0.4         
       two_point       0.4      
```
cf second column of [timings in PR #209](https://github.com/PyAbel/PyAbel/pull/209#issuecomment-362112843).    7.5ms -> 6ms == 20% improvement.

`hansenlaw` is slightly faster than `three_point` (when its basis calculation time is included)

and
```python
[etalon:.+tg121/Test-Abel]$ ipython3
Python 3.6.4 |Anaconda custom (64-bit)| (default, Jan 16 2018, 18:10:19) 
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import abel

In [2]: whole = abel.tools.analytical.SampleImage(501).image

In [3]: half = whole[:, 250:]

In [4]: half.shape
Out[4]: (501, 251)

In [5]: %timeit abel.hansenlaw.hansenlaw_transform(half)
5.88 ms ± 170 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: %timeit abel.dasch.three_point_transform(half)
1.43 ms ± 87.7 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [8]: (8.81-5.88)*100/8.81
Out[8]: 33.257661748013625    # % improvement
```
cf bottom result of the same link given above.

Benchmark figure ( i7-6700 CPU @ 3.40GHz, Ubuntu 17.10, Python  3.6.4 :: Anaconda)
benchmarks

[benchmarks_4.13.0-25-genericx86_64Linux_2018-02-26_23h47m05s.csv](https://gist.github.com/stggh/87d1dc0c48e4a40229e85d9d07712801)


DanHickstein:
Nice work @stggh! Do I sense that there is still a competitive element, trying to get HL to be the fastest transform method? :) If you manage it, I'll bet @DhrubajyotiDas will recruit some of his Intel buddies to speedup the linear algebra functions in MKL so that he can reclaim the crown. 

(An aside, but have you seen that Intel now has [it's own Python distribution](https://software.intel.com/en-us/distribution-for-python)? I was playing with it the other day for some FFTs, but it wasn't any faster for my application.)

Does the 0.35 pixel shift still appear when you are comparing with the analytical result? It's possible that HL is providing an appropriate result and the other methods are shifted for some reason.

Also, how confident are we that the `h` and `lam` values are correct? It seems possible that a typo in the manuscript is causing the issue. Or, perhaps those values were calculated using a grid that was off-by-one. Also, the length of those arrays is arbitrary, right? We should get a more accurate transform if we used more values of `h` and `lam`, yes? Potentially we should shorten (or lengthen) those arrays and make sure that the transform result is converged with respect to this approximation.

Skimming quickly, your code looks pretty good. This might be a nice opportunity to add a few more comments to the transform method.

DhrubajyotiDas:
@DanHickstein I didn't even know we had a special python distribution - I guess I'll have to migrate to that now for professional ethics reasons :)

BTW, this Anaconda claims a 97X (!!) improvement in speed on the Intel distribution (https://www.continuum.io/sites/default/files/AnacondaIntelFAQFINAL.pdf). Maybe you need the right Intel hardware (i.e., latest greatest) to see a measurable difference between this and vanilla Anaconda? 

I never quite managed to decipher the Hansen Law algorithm, so I'll trust yours and @stggh's judgment on the validity of the code. More comments in the code are mostly always good.

stggh:
> HL to be the fastest transform method?

Yes(!) and that the inverse Abel transform results for for some of the transform-pair functions are not a perfect match, as is it appears to be for `three_point`. This is important because the two/only  `forward` transform methods have what appear to be slightly inaccurate `inverse` transforms.

Curve A (profile 3) Fig 3: of Hansen & Law indicates that `hansenlaw` should do better:
image

curvea

`basex` speed would benefit from some python magic, similar to what was applied to the Dasch methods, without the need for fast linear algebra libraries, or chip based python implementations.

> Does the 0.35 pixel shift still appear when you are comparing with the analytical result?

This is not clear cut. Let's say there is not a consistent shift for all the transform pairs.

> HL is providing an appropriate result and the other methods are shifted for some reason

 `hansenlaw` and `direct` have a similar transform behaviour, however, the Dasch methods do so well it is hard to argue that they could have a problem.

> how confident are we that the h and lam values are correct

Good point, that I had also considered. I tried a least-square fit of the parameters, but the power terms e.g. Eq. (16a), make a fit possible for a variety of (extreme) parameter values. A single parameter adjustment, homes in to its own value. The `h`, and `lam` tabulated values are close.

> should shorten (or lengthen) those arrays

Interesting point. Truncating the size to 5, behaves in a way that you expected:
curvea-trunk


> add a few more comments to the transform method

You assume that I understand the method enough to make intelligent comments ... ;-)

stggh:
I have dropped the pixel `shift` parameter from `hansenlaw`, as it may cause confusion, and I no longer believe in its universal applicabilty. The pixel image shift only appears necessary to align the methods for (my) O2-photoelectron spectra, and not for the transform-pairs (as I once thought). The correction may just as easily be applied to the radial profile, as a small radial correction (`x1/1.0008`) e.g see `example_all_O2.py`. This figure (not identical to the example) is for comparison of those shown above (just uses an energy scale):

o2

I have added a few more comments to the `hansenlaw` doc string, but actually have little background in the "state-variable method approach" :-(   The coding is simply the equations given at [the top](https://github.com/PyAbel/PyAbel/pull/211#issue-171551556) of this PR.

I have also added the `circularize_image` J. Chem. Phys. reference to the function doc string and associated .rst file, as requested in #210 

Not sure about the windows travis-ci build problem?

DanHickstein:
Thanks for this analysis @stggh!

Can you make the "Curve A, N=101" figure, so that we may make a direct comparison to Fig. 3 of the HL paper. Also, can you calculate the mean-squared error, so that we can compare to the 1.33-e6 value listed in the HL paper? It seems that we are somehow falling short of the performance shown in their manuscript, which points to some annoying bug.

I am curious if we can calculate our own h and lam values, since the paper says that they are only calclulated to an error of 0.001. Perhaps we can do better, and make a larger expansion. However, I found that paragraph of the paper (regarding the calculation of h and lamda) to be very confusing.

stggh:
@DanHickstein  thanks for the guiding, it has helped identify issues that I thought I had visited.


`N=101` and mse (good idea). 

Interesting:
(1) with the `hansenlaw.py` code as above, poor result for `hansenlaw` (orange curve), `three_point` does very well, mse=<math>10<sup>-9</sup></math>
curvea-mse0 0


(2) driving function (gradient or image) shifted by 1 pixel left. Better agreement between `hansenlaw` and `direct`:
```python
    # 1-pixel shift better aligns 'curve A' transform pair
    drive[:, :-1] = drive[:, 1:]
```
curvea-mse-1 0

(3) 2 pixels, `drive[:, :-2] = drive[:, 2:]`, mse worsens, a shift too far
curvea-mse-2 0


(4) -4/3 shift, using `drive = scipy.ndimage.interpolation.shift(drive, (0, -4/3))`, mse still not = H&L
curvea-mse-1 3

No pixel shift can achieve a mse=1.33E-6 shown in H & L.

A 1-pixel offset may be due to relative indices for transform arrays cf the driving array, the `n+1` index:
Hansen and Law (index 0 is right side of image):
image
and corresponding code:
```python
    for col in n:  # right image edge to left edge
        X = Phi[col][:, None]*X + Gamma[col][:, None]*drive[:, col]
        AIM[:, col] = X.sum(axis=0)
```

Similarly, for Curve B (profile8):
curveb-mse0 0
curveb-mse-1 0
curveb-mse-1 3

It is clear that there should at least be a -1 pixel offset/shit. A fractional shift, I am not so sure.

I have temporarily returned the `shift` parameter to `hansenlaw`, and here is the test code  [Curve A code](https://gist.github.com/stggh/fa310551246cccba553b21e21d4cd442 )

stggh:
Ok, it turns out that the `col` index, for the transform code line,  was not aligned with the `Phi` and `Gamma` arrays. This gave a 1-pixel offset. Now curve A `mse=5.77e-5`. The inverse transform is very similar to the `direct` method result. However, (once again!) a `-1/2` pixel shift gives `mse=1.07e-6`, which is better/close to the Hansen & Law value of `1.33E-6`.

In the `hansenlaw` code I am now, using 'N' (for `cols`) in the `hansenlaw` code to better compare with the H&L variable names.
Note that the weightings  `(N-n)/(N-n-1)` need to be the numpy index order. This has always been the case, but points to an imperfect description in H & L.
```
   +----+----+                    +----+----+
   |  0 |  1  |                   | N-2 | N-1 |    numpy
   +----+---- +       ...         +----+----+
   | N-1 | N-2 |                  |  1 |  0 |     H & L
   +----+----+                    +----+----+
  (N-n)/(N-n-1), (N-n-1, N-n-2) ...(2/1)  (und.)
```

curvea-mse0 0
curvea-mse-0 5
Curve A zoomed:
`shift=0` tranfrom similar to `direct`
curvea-0zoom
`shift= -1/2`, better agreement with transform pair
curvea-0 5zoom

Curve B:
curveb-mse0 0
curveb-mse-0 5


In conclusion, it is easy to make an index error. The `-1/2` pixel shift may also apply to the `direct` method. This may have something to do with grid alignment to image pixels #39

stggh:
**Python weirdness**

The "good" mse of `1.0e-7` for `hansenlaw` depended on whether a forward transform was calculated, *anywhere* in the code. All very odd!

This code produces a different result, `9.24e-07`, from `1.52e-04` if the forward transform is called.
```python
import numpy as np
import abel

shift = -1/2

f = abel.tools.analytical.TransformPair(101, 3)  # curve A

hl = abel.hansenlaw.hansenlaw_transform(f.abel, dr=f.dr, shift=shift)

# this line changes mse hl from 1.52e-04 to 9.24e-07
corrupt = abel.hansenlaw.hansenlaw_transform(f.func, dr=f.dr, direction='forward', shift=shift)

msehl = np.square(hl-f.func).mean()
print('hansenlaw mse = {:4.2e}'.format(msehl))
```
The forward transform has, line 143: `drive = IM`, which somehow changed/corrupted the input `(f.func)`. I have set `drive = IM.copy()` to give a consistent result.

Unfortunately, the optimum, least-squares fit, is now no better than `1.25-04`, for `shift=-0.16`, which has got me wondering why the "corruption" produced such a perfect result ...

corruptedfunc

stggh:
This PR is turning into a blog of my incompetence.
(1) Python's pass by value will pass back the array change. Something that I "should" be familiar with :-(: 

```python
In [1]: import numpy as np

In [2]: def func(x):
   ...:     y = x
   ...:     y[2] = 10
   ...:     return y
   ...: 

In [3]: x = np.arange(9)

In [4]: x
Out[4]: array([0, 1, 2, 3, 4, 5, 6, 7, 8])

In [5]: yy = func(x)

In [6]: x
Out[6]: array([ 0,  1, 10,  3,  4,  5,  6,  7,  8])
 ```
(2) Final `hansenlaw.py` code:
   **1-pixel offset** of `Phi` and `Gamma` arrays relative to the input image (forward transform) or gradient (inverse transform), `drive`, 'AIM` (transform). Coded as:
```python
    for col  in n:  # right image edge to left edge
        X = Phi[col][:, None]*X + Gamma[col][:, None]*drive[:, col+1]
        AIM[:, col+1] = X.sum(axis=0)
```
This form executes slightly faster:
```python
    n1 = n + 1 # Phi, Gamma, 1-pixel offset relative to image
    for col, col1  in zip(n, n1):  # right image edge to left edge
        X = Phi[col][:, None]*X + Gamma[col][:, None]*drive[:, col1]
        AIM[:, col1] = X.sum(axis=0)
```
For curveA (`profile3`) and curveB (`profile8`) `hansenlaw` produces very similar results to `direct`:
curvea-zoom
cruveb-zoom
```
[etalon:.+tg121/Test-Abel]$ python3 CurveA-mse.py 

  forward method     mean-squared error
  --------------     ------------------
    hansenlaw             5.77e-05
    direct                2.99e-05

  inverse method     mean-squared error
  --------------     ------------------
    hansenlaw             5.77e-05
    direct                5.27e-05
    three_point           5.68e-09
[etalon:.+tg121/Test-Abel]$ python3 CurveB-mse.py 

  forward method     mean-squared error
  --------------     ------------------
    hansenlaw             1.30e-04
    direct                6.52e-05

  inverse method     mean-squared error
  --------------     ------------------
    hansenlaw             9.07e-05
    direct                2.92e-05
    three_point           6.76e-08
```
**additional -1/2 pixel offset**
`hansenlaw` (and possibly `direct`) can do better if the input is shifted left by 1/2-pixel:
curvea shift-zoom
curveb shift-zoom

```
[etalon:.+tg121/Test-Abel]$ python3 CurveA-mse.py 

shift = -1/2

  forward method     mean-squared error
  --------------     ------------------
    hansenlaw             1.55e-06
    direct                2.99e-05

  inverse method     mean-squared error
  --------------     ------------------
    hansenlaw             1.07e-06
    direct                5.27e-05
    three_point           5.68e-09
[etalon:.+tg121/Test-Abel]$ python3 CurveB-mse.py 

shift = -1/2

  forward method     mean-squared error
  --------------     ------------------
    hansenlaw             5.63e-06
    direct                6.52e-05

  inverse method     mean-squared error
  --------------     ------------------
    hansenlaw             3.06e-06
    direct                2.92e-05
    three_point           6.76e-08
```
Here are the inverse transform pair % different plots  (= (method-f.func)*100):
transformpairs-inverse-diff
and forward transform:
transformpairs-forward-diff

I conclude that the `hansenlaw`, and possibly `direct`, methods apply to an even width image, see the grid discussion in #39.  The additional 
```python    # 1/2 pixel shift improves transform pair agreement, see #206, #211
    drive = interpolation.shift(drive, (0, -1/2))
```
increases the transform time, although, it is still comparable `three_point` (including the basis calculation). It may be better to simply check whether the image is odd, and apply the grid shift if required.

I believe (*hope*!) that the `hansenlaw` code is correct, and the transform alignment issue is now solved.

My apologies for all the toing and froing.

DanHickstein:
> Python's pass by value will pass back the array change.

I think that it's actually "pass-by-reference-like" behavior that is getting us into trouble. I actually think that Python is fairly confusing in this regard, and it sometimes seems to flip back and forth between passing by reference and passing by value. (This opinion is probably because I don't really understand what a "mutable" variable is, see [here](https://stackoverflow.com/questions/986006/how-do-i-pass-a-variable-by-reference).)

This behavior has caused problems for me in the past. I think that we should modify the functions in PyAbel such that they don't modify the original image. Basically, I think that we need a np.copy() statement at the top of a lot of the transform functions. Could you open a new Issue about this, and we can maybe get @rth to weigh in, since he likely has an informed opinion on this. 

It looks like you've made good progress on the sub-pixel shift though! Do I understand correctly that you have a MSE of 1.55e-6, which the H&L paper reports 1.33e-6? That seems like reasonably good agreement. That is annoying that the HL algorithm seems to operate on an even grid. Perhaps if we could figure out how to recalculate the `h` and `lam` arrays, then we could modify it to work on an odd grid.

Are you basically ready to merge? Should I have a final look through the code (when I get some time)?

stggh:
>  "pass-by-reference-like"

You are right. 

> need a np.copy() 

`direct` does this (of course @rth!), `basex` and `linbasex` create a full image, leaving only the methods that I had a hand in, that do not copy the input image ...   
Not sure this needs discussion, the other methods simply should follow @rth's code: `np.atleast_2d(image.copy())`. I have made the change to `hasenlaw`, dasch methods (which oncludes `three_point`, `two_point`, and `onion_peeling`), and `onion_borads` in this PR.

> MSE of 1.55e-6, which the H&L paper reports 1.33e-6

Actually, `1.07e-06` for curve A.

>ready to merge

Hold off until after the W/E.

rth:
>> need a np.copy()
>
> direct does this, basex and linbasex create a full image, leaving only the methods that I had a hand in, that do not copy the input image ...
Not sure this needs discussion, the other methods simply should follow @rth's code: np.atleast_2d(image.copy()).

From what I have seen, sometimes transformers have an option to copy or do inplace operations (e.g. [sklearn.preprocessing.StandardScaler](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html)). Making a copy is safer but it does have some performance cost, which can be negligible or not depending on the application.

For instance [fft](https://github.com/numpy/numpy/blob/v1.14.0/numpy/fft/fftpack.py#L189) does call `np.asarray` (`np.atleast_2d` would preclude 1D arrays) for validation but it doesn't make a copy. I imagine that the `_raw_fft` doesn't do inplace modifications of the input array.

DanHickstein:
Thanks @rth!

So, I think that there are two issues:

1) What do users expect?

2) What is the performance cost?

Regarding the first issue, I think that most people could expect our Abel transform method to work similarly to something like `np.rot90` or `scipy.ndimage.interpolation.rotate`. These methods do not overwrite the input array, but simply return a new array with the result. ([Some quick code to test this behavior](https://gist.github.com/DanHickstein/7b5ef8e8b244b0f05832141972126fc3).)

We might not need an actual "np.copy", so long as we are using some function that makes a copy of the array. But, perhaps making a copy is the safest way.

It's funny that you bring up the issue of the fft, because `scipy.fftpack.fft` does have an option `overwrite_x`, which is False by default, but can be set to True an allows the function to overwrite the input array with the fft, which can allow for higher performance in some situations, since a copy doesn't need to be made. In any case, since the default behavior is False, this communicates to me that the "expected" behavior is to make a copy of the array, and this should only be overridden if performance is an issue.

So, that brings us to the second issue: is performance an issue in this case? I prepare a quick [gist](https://gist.github.com/DanHickstein/a0536df5d08bc3442babe94550083895) to compare the speed of `np.copy` for arrays of various sizes. Unfortunately, my quick test shows that np.copy might be comparable to the speed of the fastest transform methods:

image

We should do a more careful test to see if copying the array actually causes a noticable slowdown when incorporated into the transform methods. If so, then I suppose we can do something similar to the `scipy.fftpack.fft` where we provide the "safe and slow" behavior (copying the array) by default, and allow users to select the "fast and crazy" (overwriting the array) if they require higher performance.

stggh:
> pass-by-reference-like

 Note, I was a little too fast in requiring all algorithms to `numpy.copy()` the input image.
 Input image destruction only applies to `hansenlaw` (forward), `onion_bordas`, and `direct` (probably, only forward too).

> merge - hold off until after the W/E.

A long-weekend in Canberra (public holiday).

`hansenlaw` is definitely an even column  algorithm.  Here are the transform pair difference plots. Note, incorrect labeling of the y-axis for the pair plots above, the forward transform compares with the projection function `G(R)` (in hansenlaw labeling), and the inverse transform with the source function `f(r)`. 

Inverse Abel transform pairs, difference curves:
transformpairs-inverse-diff
Note, the profile3 (== curveA) mse of 1.13e-6 (from 1.07e-6 above, is due to the 1/2-pixel shift being applied, externally to the image, rather than, within `hansenlaw` to the gradient (not image)).

Forward Abel transform pairs:
transformpairs-forward-diff

I have pulled the `scipy.ndimage.interpolate.shift()` code from `hansenlaw.py` since:
1. It is not the role of a method code to correct an image grid/centre, although, `onion_bordas` does attempt a correction. However, IMO, it cannot be universal, without considering how the image was centred.  
2. In particular, the user would need to specify the whole-image-shape, and centre method, as separate parameters.
3. The 1/2-pixel difference between transform of an odd or even image is only noticeable on a spectroscopic scale, and in general the speed distribution scaling would "correct" the difference.

Hence, for spectroscopic accuracy, the input (whole) image should have an even column size. I should make changes to the docs to reflect this. Some, changes have been made.

The even image width for `hansenlaw` causes a few issues, as it now joins the `onion_bordas` method, with `basex`, `three_point`, `two_point`, `onion_peeling`, and `direct` being odd size. Again, the docs should be examined. 

In summary, I think a more careful look at the method codes to check whether the input image is necessarily, mangled, and so whether a `.copy()` is required. I have reverted the dasch methods. The `hansenlaw` method applies to an even width (whole) image, grid axis on the left of the pixel, as per the diagram in the discussion #39. As long as the user knows, I don't think this is a big issue. Check the docs for any inconsistencies.

rth:
>  I think a more careful look at the method codes to check whether the input image is necessarily, mangled, and so whether a .copy() is required. 

Yes, I think it might be an issue only in a few methods.

> It's funny that you bring up the issue of the fft, because scipy.fftpack.fft does have an option overwrite_x,
> Unfortunately, my quick test shows that np.copy might be comparable to the speed of the fastest transform methods:

For those where it is an issue, just having an optional 'copy=True' input parameter, might be the simplest way to get around this. This way it could be on by default to be safe, and carefully disabled when performance matters.

DanHickstein:
@stggh, thanks for the nice comparison! 

So, there are two issues here: 

1) As @rth suggests, the issue of transform methods mangling arrays is fairly easy to deal with: they shouldn't mangle the input array unless the user "gives permission" for this to happen via a boolean kwarg. I would probably call this kwarg something like "overwrite_input", or something like this.

2) The issue of which grid the transform methods use is much more difficult. We are calling this the "even/odd" issue, but it's perhaps better called the "y[0]=0 versus y[0]=0.5" issue (see #39 for nice diagrams by @rth). I was hoping that this issue was effectively resolved, because all of the method (except for Bordas) worked on odd images. But, if we now think that HL works for even images, then we need to think more carefully about how we want to deal with this. Fundamentally, it makes a bit more sense to me to use odd images, since the value at y=0 is specified.

I think that the direct method should be able to work on any grid, even a non-uniform grid, in principle. However, looking at the code, I'm fairly sure that is assumes an even grid. I opened a PR (#213) to change this to an odd grid. 

@stggh, can you confirm that HL expects an even image? As you describe it, it sounds like you are just shifting the input by 0.5 pixels. But, to make a comparison with another method, I think that you would need to shift the input by 0.5 pixels, and then either shift the output back to the odd-grid, or make sure that you are plotting using an array of x-coordinates that is on the `[0.5, 1.5, 2.5, 3.5 ...]` grid.

If it does assume and even image, then we need to decide how to proceed. In general, it seems that PyAbel could provide the best comparisons if all of the methods adopted the same convention. So, one option would be to try to modify the method to use an odd image convention. It seems that [Hansen is still a Prof. at Dartmouth](https://engineering.dartmouth.edu/people/faculty/eric-hansen). I might send him an e-mail and ask him if the algorithm could be modified in this way. Also, I am curious about the values in table 1, and if there is any incentive to re-calculate them with higher accuracy.

stggh:
>  Input image mangling
 
A quick test reveals only `hansenlaw` (`direction=forward`) mangled the array. Now fixed. 
No copy flag required, unless changes are made to `direct` and/or the other methods, to reduce any copy activity.

> can you confirm that HL expects an even image
> then either shift the output back to the odd-grid, or make sure that you are plotting using an array of x-coordinates that is on the [0.5, 1.5, 2.5, 3.5 ...] grid

This is a good point. I have removed the image `ndimage.interpolation.shift(IM, (0, -1/2))` from `hansenlaw.py`, so that the method is parity independent. The grid adjustment is applied before the transform. The transform is compared directly with the unshifted-source function.


1. odd parity, image centre mid-pixel = no change to the transform pair function:
```python
n = 100
f = abel.tools.analytical.TransformPair(n=n, profile=3)  # curve A
# inverse Abel transform
hl = abel.hansenlaw.hansenlaw_transform(f.abel, dr=f.dr)
tp = abel.dasch.three_point_transform(f.abel, dr=f.dr)
dr = abel.direct.direct_transform(f.abel, dr=f.dr)

msehl = np.square(hl-f.func).mean()
msetp = np.square(tp-f.func).mean()
msedr = np.square(dr-f.func).mean()
```
```
  inverse method     mean-squared error
  --------------     ------------------
    hansenlaw             5.88e-05
    direct                5.37e-05
    three_point           5.93e-09
```
curvea-odd

**Note**  the `hansenlaw` transform is pretty similar to the result of `direct`. 
However, see new discussion #213, which intimates that direct assumes an even grid if `r` is not supplied.

2. even-parity image = apply `shift(f.abel, -1/2)
```python
fabel = interpolation.shift(f.abel, -1/2)

hl = abel.hansenlaw.hansenlaw_transform(fabel, dr=f.dr)
tp = abel.dasch.three_point_transform(f.abel, dr=f.dr)
dr = abel.direct.direct_transform(f.abel, dr=f.dr)

msehl = np.square(hl-f.func).mean()
msetp = np.square(tp-f.func).mean()
msedr = np.square(dr-f.func).mean()
```
```
  inverse method     mean-squared error
  --------------     ------------------
    hansenlaw             1.15e-06
    direct                5.37e-05
    three_point           5.93e-09
```
Here `hansenlaw` has much better agreement with the curveA source function:
curvea-even

The `mse=1.15e-06` is similar to that stated for Fig. 2 of the H&L paper, `1.33E-6`.
However, the figure caption also states 'N=101`!

Similar improvements, occur for even images (`n=101`) [code](https://gist.github.com/stggh/94c783e6e76ec819d6369eb13855934b ):

Profile      |     mse(odd)   |   mse(even -1/2 shift)
--------      |     -------------  |   --------------------------
  1            |       5.68e-5    |      1.25e-6
  2            |       5.48e-5    |      1.15e-6
  3  (curve A) |  5.77e-5    |      1.13e-6 
  4             |      (bad defn) |        -
  5             |      6.66e-3     |      4.03e-3
  6             |      9.07e-5     |      3.07e-6
  7             |     9.58e-5      |     3.28e-6
  8  (curve B) |  9.07e-5     |      3.07e-6 

> I'm pretty surprised by how sensitive these transforms are to an even/odd grid

Only in terms of a zoomed image, or `mse`.

The comparison is complicated by the fact that only `hansenlaw` and `direct` have forward transforms. I will examine `hansenlaw` in relation to an odd-`direct` before making a definitive statement on whether `hansenlaw` applies to an odd or even image width. i.e. I am still not 100% sure.

stggh:
@DanHickstein grid bug fix for `direct` #213 now provides an accurate forward transform, to compare e.g. the Dribinski sample image transform (forward <-> inverse) with `hansenlaw`. The `hansenlaw` -1/2 pixel shift, applies only to each row. So, for a 2D image this would be `scipy.ndimage.interpolate(IM, (0, -1/2))`. This is easier to implement the shift in `hansenlaw.py`, so that `abel.Transform(..., method='hansenlaw')` includes the shift for each quadrant processing. I have added a new parameter `align_grid=True` which applied the shift. For even-width images (center pixel edge) `align_grid=False` will not apply the `shift` (intensity interpolation) and also provide a faster transform.

recycle4
```
    method    mse-image(transform-source)   mse-speed(transform-source)
  hansenlaw               26.28              1.43e+10
  direct                1394.02              1.93e+10
  three_point             13.56              5.54e+09
```
[code](https://gist.github.com/stggh/c92343886c33c5cc0dbebe30e579875b)

Subject to the choice of variable name `align_grid`, and whether the docstring is sensible, I am happy with the compromise(*) that all methods should apply to odd-width (centre pixel grid) images.

(*) `scipy.ndimage.interpolate(), when applied, slows the algorithm

stggh:
> see #39 for nice diagrams by @rth 
 
Hey, I think they may have been [my diagrams](https://github.com/PyAbel/PyAbel/issues/34#issuecomment-160344809)   ;-)

rth:
> Hey, I think they may have been my diagrams ;-)

Yes, I looks I took some of your for that issue :)

DanHickstein:
> Yes, I looks I took some of your diagrams for that issue :)

Ha! Well, thanks to both of you for the nice diagrams - @stggh for making them, and @rth for reposting them :)

I wrote a quick test to investigate if HansenLaw suffers from an "even/odd" grid issue. I don't think that just shifting the input grid is quite the right way to go about this. Fortunately, with the analytical methods, we can simply generate a grid that is twice the density that we need, and pass either the even or the odd points to our transform function.

``` python
import numpy as np
import abel

n     = 41
r_max = n
sigma = n*0.25

ref  = abel.tools.analytical.GaussianAnalytical(n, r_max, sigma, symmetric=False)
func = ref.func
proj = ref.abel 

proj_even = proj[1::2]
proj_odd  = proj[0::2]
proj_shift = scipy.ndimage.interpolation.shift(proj, (-0.5))

r    = ref.r
dr   = ref.dr

r_even = r[1::2]
r_odd  = r[0::2]

fig, axs = plt.subplots(2, 1, figsize=(5,8), sharex=True, sharey=True)

axs[0].plot(r, func, label='Analytical', lw=1)
axs[1].plot(r, func, label='Analytical', lw=1)

hl_even  = abel.hansenlaw.hansenlaw_transform(np.copy(proj_even), dr=2, direction='inverse' , align_grid=False)
hl_odd   = abel.hansenlaw.hansenlaw_transform(np.copy(proj_odd ), dr=2, direction='inverse' , align_grid=False)
tp_odd   = abel.dasch.three_point_transform(proj_odd, dr=2, direction='inverse')
tp_even  = abel.dasch.three_point_transform(proj_even, dr=2, direction='inverse')

axs[0].plot(r_even, hl_even,  'o', label='HL with even grid')
axs[0].plot(r_odd,  hl_odd,   'o', label='HL with odd grid')

axs[1].plot(r_even, tp_even,  'o', label='Three-point with even grid')
axs[1].plot(r_odd,  tp_odd,   'o', label='Three-point with odd grid')

axs[0].legend(frameon=False)
axs[1].legend(frameon=False)

plt.show()
```

image

We see that Three-point has a clear preference for the "odd" grid. HL seems to be off for both grids. The fact that a 0.5 pixel shift to the input grid seems to improve the performance of the HL method, point to some "off by one error" somewhere in the code, but I'm not convinced that it's simply an issue with even/odd grids.

stggh:
@DanHickstein - neat coding!  I will investigate.

stggh:
Note, that for the odd-size image, `align_grid=True`, to apply the -1/2 pixel shift. In which case the transform is better:
dhhl

I agree, even, with `align_grid=False` should achieve the same result - so, something wrong for the even case.

stggh:
subtract `1` from `r_even`
```python
axs[0].plot(r_even-1, hl_even,  'o', label='HL with even grid')` 
```
dhhl

I have grid confusion ...

stggh:
Back to our nice `PyAbel` grid figure, even-column-image (centre is left pixel edge) vs odd-column-image (center is mid-pixel):
image
Here is my take on the -1/2 pixel shift (x = interpolated values):
dhhlo

It is not a matter of odd, or even sampling of the (Gaussian) function, or whether the image has an odd or even number of pixels. Just the location of the grid on which the image intensity is evaluated.

DanHickstein:
Hmmm, yeah, so I guess the grid can be defined in terms of the left side of the bin, the center, or the right side. But, the center seems like the most obvious choice.

I wonder if there is just some small error in the indexing in our HL algorithm, and we are just finding different ways to rationalize and compensate for the behavior. Looking again at the code, I had forgotten that the HL algorithm doesn't actually calculate he intensity at the center pixel, but instead just copies the value from the adjacent pixel.

stggh:
> I wonder if there is just some small error in the indexing in our HL algorithm

Agreed! I have been looking, over the past 2 months, but may be missing something obvious.

> HL algorithm doesn't actually calculate he intensity at the center pixel

That's because `(N-n)/(N-n-1)` is &infin; for `n=N-1` (index `n = 0` at right side). This should apply for both the forward and inverse transform, yet the H&L paper initimates there is only an issue for the inverse transform:
image

stggh:
Not sure whether this helps, but displacing the ratio matrix by -1/2, numerator and denominator:
```python
    denom = N - n - 1 - 0.5 # N-n-1 in Hansen & Law
    ratio = (N - n - 0.5)/denom  # (N-n)/(N-n-1) in Hansen & Law
```
achieves a similar effect to `shift(IM, (0, -1/2))`
a

which points to @DanHickstein's "some small error in the indexing in our HL algorithm".

DanHickstein:
Oh, very interesting!

Is your HL code adapted from code written by H&L, or was it developed by someone else from the equations in the paper? I'm thinking about e-mailing Hansen to see if he has the original code sitting around on some old floppy disk (or tape drive? I was 1 year old in 1985 and don't know what technologies were in use. Hopefully not punch cards.)

stggh:
> HL code adapted from code written by H&L

No, it is my interpretation of what is written in their paper.

A couple of things are inconsistent/uncertain:
1. <math>(N-n)/(N-n-1)</math> vs `n=0, ..., N-1`   (forward transform), which would blow up for `N-1`.
2. <math>x<sub>n+1</sub> = &Phi;<sub>n</sub> x<sub>n</sub> + &Gamma;<sub>n</sub> f<sub>n</sub> </math>
<math> g<sub>n</sub> = C x<sub>n</sub><math>
The Python code works better with <math> g<sub>n</sub> = C  x<sub>n+1</sub><math>.
3. I use <math> x_0 = 0</math> for the initial value.

In the very distant past, I recall that we (ANU) did once contact Hansen and received a code (fax?) listing. However, it was not used, as I thought my version of the algorithm was working. :-(

I will be moving offices soon, and in the process, may find it.

> 1985

Chronology, I received my PhD in 1984!  We had mainframes, Commodore PET's, and HP calculators. ;-)

I will keep playing.

stggh:
Here is a one-row version of `hansenlaw`:
```python
import numpy as np
import abel

# Hansen & Law parameters of exponential approximation, Table 1.
h = np.array([0.318, 0.19, 0.35, 0.82, 1.8, 3.9, 8.3, 19.6, 48.3])
lam = np.array([0.0, -2.1, -6.2, -22.4, -92.5, -414.5, -1889.4, -8990.9,
               -47391.1])

def hl(im, dr, shift=0):
    aim = np.zeros_like(im)
    N = len(im)

    deriv = np.gradient(im)

    K = len(h)
    gamma = np.zeros(K)
    x = np.zeros(K)

    for col in np.arange(N-1):

        ratio = (N-col-shift)/(N-col-1-shift)

        phi = ratio**lam

        gamma[0] = -h[0]*np.log(ratio)
        gamma[1:] = h[1:]*(1-phi[1:])/lam[1:]

        x = phi*x + gamma*deriv[N-col-1]
        aim[N-col-1] = x.sum()

    # missing 1st pixel
    aim[0] = aim[1]
    return aim/dr


# main -------------------------
f = abel.tools.analytical.TransformPair(101, 3)

shift = 0.45
him = hl(f.abel, f.dr, shift)

mse = np.square(f.func-him).mean()
print("shift={:.2f}, mse = {:4.2e}".format(shift, mse))

# plot
plt.plot(f.r, f.func, label=r'Curve A $f(r)$')
plt.plot(f.r, him, label='hansenlaw')
plt.legend()
plt.annotate(r'mse={:4.2e}'.format(mse), (0.7, 0.8))
plt.annotate(r'shift={:.2f}'.format(shift), (0.7, 0.75))
plt.xlabel(r'$r$')
plt.ylabel(r'$f(r)$')

plt.savefig('hansenlaw-1row.png', dpi=75)
plt.show()
```
hansenlaw-1row

DanHickstein:
> I will be moving offices soon, and in the process, may find it.

Okay, let me know if you find it, otherwise I am tempted to e-mail Hansen sometime next week and see if we can get the original code. Your code does seem quite close, so I'm sure we can keep using it, we just need to figure out exactly where this heisenbug is located.

> Here is a one-row version of hansenlaw

Awesome! For diagnosing this issue, it's probably best to work with the simplest example possible.

stggh:
**Full circle**

The ratio correction `(N-n-shift)/(N-n-1-shift)` is a furphy, it does not produce the same amount of agreement as `scipy.ndimage.interpolation.shift(IM, (0, -1/2))`.  I have returned the `(0, -1/2)` pixel shift, if `full_image_odd_cols=True` (default).  I am open to a better choice for the variable name.

I conclude (again ;-)), `hansenlaw` is an even width full image transform.

transformpairs-inverse-diff
transformpairs-forward-diff
recycle4

[test codes gist](https://gist.github.com/stggh/ee4e026d1244183acb6f3a6b717e52ff)

stggh:
This follow up June '85 paper by Hansen [IEEE Trans. Acoust. Speech. Signal Proc. 33, 666 (1985)](10.1109/TASSP.1985.1164579) has a variation on the April '85 `hansenlaw` method:
> we have found that a piecewise linear, or “first-order hold” approximation gives superior
results.

image

where, `B0`(&beta;0) and `B1`(&beta;1) replace &Gamma;(&gamma;) in the top of this discussion [formula graphic](https://github.com/PyAbel/PyAbel/pull/211#issue-171551556).

In this June paper, Hansen states the test functions have an **even** column number:
> III. Experiments ...  with both 128 and 512 samples

I am in the process of coding the "first-order hold" *forward* transform,  which may provide a useful comparison with the (accuracy of the) current method. However, the fact the `hansenlaw` method needed to be tweaked, may suggest that it will not meet the same level of accuracy we have attained using the Dasch `three_point` method.

The "first-order hold" approximation has been implement in the *fast Hankel transform*  [Matlab code](https://au.mathworks.com/matlabcentral/fileexchange/59636-fast-hankel-transform), which lead  to the above paper and this alternative `hansenlaw` algorithm (for the forward transform).

stggh:
Hanson published a later [correction](https://doi.org/10.1109/TASSP.1986.1164860) to the &beta; expressions:
screenshot-20180329091541-416x90

The algorithm slightly improves the *forward* Abel transform, **and** no '-1/2' pixel shift needed.
a

 | forward method   |  mean-squared error |
  --------------   |  ------------------
 |   hansenlaw  |           1.55e-06 |
 |   hansenlaw2 |           7.25e-07 |
 |   direct            |    2.66e-06 |

(note: the orginal `hansenlaw` applies `-1/2` pixel shift)

NB AFAICT the new method only applies to the `forward` transform.

[hansenlawv2.py](https://gist.github.com/stggh/8599ffda390cd5a9c9b7c49b1c7c2ae0)

stggh:
I should add it is possible map the same grid to each formula (see new push to my HansenLaw-fix fork), and so the `-1/2`-pixel shift  implies a even column width image for the `hansenlaw` method cf the `hansen`, above. i.e.
`hansenlaw`
```python
    for indx,col in zip(n[::-1]-1, n[1:]):  # 
        x  = phi[indx][:, None]*x + gamma[indx][:, None]*drive[:, col]
        AIM[:, col] = x.sum(axis=0)
```
`hansen`
```python
    for indx, col in zip(n[::-1]-1, n[1:]):
        x  = phi[indx][:, None]*x + B0[indx][:, None]*drive[:, col+1]\
                                  + B1[indx][:, None]*drive[:, col]
        AIM[:, col] = x.sum(axis=0)
```

DanHickstein:
Oh, so we missed the "Correction" article! The title of the article says that it's a correction to the Fast Hankel Transform, but it also applies to the Abel transform as well?

And it appears that the correction allows the forward transform to achieve good agreement with the analytical results? Very exciting! Are you passing the hl2 method points on an even or odd grid?

And now you think that it might be possible to incorporate this correction into the inverse transform as well?

stggh:
> correction to the Fast Hankel Transform, but it also applies to the Abel transform as well?

Hansen's second paper defines a more accurate forward Abel transform,  independent of the evaluation of the Hankel transform(*).  The function <math>f(&xi;)</math> previously assumed constant across each interval (pixel), in paper I, is now a linear approximation  <math>f(&xi;)=p + q&xi;</math>. This changes the recursion expression from:
image
to
image
The subsequent paper III *correction* redefined &beta;<sub>0</sub> and &beta;<sub>1</sub> expressions used to evaluate <math>B<sub>0</sup></math>, <math>B<sub>1</sup></math> of paper II.

> Are you passing the hl2 method points on an even or odd grid?

The linear approximation improves the `forward` Abel transform, over the `hansenlaw` (paper I) method, **and** does not require the source function -1/2 pixel shift. The following  plots compares all the `foward` transform methods: new `hansen` (no shift, blue), `hansenlaw` with -1/2-pixel shift (orange dash),  and 'direct' (green dots), for each of the `abel.tools.analaytical.TransformPairs()`, mean-square-errors annotated:
transformpairs-forward-diff2

All **great**, except the &beta;<sub>0</sub> and &beta;<sub>1</sub> expressions apply only for the `forward` Abel transform.

> possible to incorporate this correction into the inverse transform as well?

No, unless we have expressions  <math>&beta;<sub>0</sub></math> and <math>&beta;<sub>1</sub></math> for the `inverse` transform. We could attempt to evaluate:
image
using sympy? or query Hansen, our best bet. I will send him something tomorrow, today is the Easter Monday holiday in Oz.

For testing, I have uploaded a new file `hansen.py` into my [HansenLaw-fix fork](https://github.com/stggh/PyAbel), called in the usual manner, `abel.hansen.hansen_transform(IM, dr)`

Alternatively, [this gist](https://gist.github.com/stggh/30b6bf715f4051f2c3e0661a60942562) has a slightly more readable code, applicable for 1d images.

(*) The paper makes some interesting comments on sampling for the FFT

stggh:
Eric Hansen has kindly outlined the algebra to obtain the gamma (&gamma;) expressions Eq. (16c) and (18) of the [Hansen & Law paper](https://doi.org/10.1364/JOSAA.2.000510), from the state equation (Eq. (9) in the previous comment), so that we may obtain similar expressions for the inverse Abel transform in the first-order hold approximation, &beta;<sub>0,1</sub> Eq. (12), of the Fast Hankel [paper](https://doi.org/10.1109/TASSP.1985.1164579).  It turns out that &Phi; in state equation integral evaluates to:
image
where `a=0, 1, 2`, depending on the transform direction, and the order of the hold approximation.  The zero-order hold approximation, where the source function is assumed constant across an interval is the `hansenlaw` transform, while the first-order hold approximation, assumes a iinear variation, "gives superior results" (as quoted from Hansen's Hankel paper).

The bottom line is that it is relatively straightforward to code the first-order hold approximation for both the `forward` and `inverse` Abel transforms. The results are excellent, for the Abel transform pairs, here plotted as % difference. Profile 3 is "curveA", blue `hansen`, yellow `direct`.
Forward:
transformpairs-forward-diff
Inverse, + green `three_point` difference:
transformpairs-inverse-diff
The annotations are the mean-square error values.

`three_point` still wins for the speed and accuracy of the inverse Abel transform, but the Hansen algorithm is a winner for the `forward` transform (cf `direct`).

@DanHickstein Q: discussion concerning `PyAbel` structure: 
   Two new files have been added to this PR (my [HansenLaw-fix fork](https://github.com/stggh/PyAbel), `hansen1d.py`, a single row or 1D-function Abel transform, and `hansen.py` the 2D vectorized version.  The 1d-version can be deleted before the merge.  

  `hansen.py` replaces `hansenlaw.py`, with the parameter `hold_order=0`. There is little reason to use this approximation, given superiority of the first-order hold (default `hold_order=1`). 

  I propose that `hansenlaw.py` and associated links be dropped, to be replace with `hansen.py`.  The transform call would be: `abel.hansen.hasen_transform(image, dr=1, direction='inverse', hold_order=1)`. i.e. similar to the Dasch methods `abel.dasch.three_point_transform()`. Here, it would be the Hansen methods. Note, I realise that this would break the `PyAbel` API, but it is pointless to carry `hansenlaw.py`.  I also do not feel we should simply call the new algorithm `hansenlaw`, because Law was not part of the first-order hold approximation. If need be we could maintain a functional link `hansenlaw.hansenlaw_transform()` to `hansen.hansen_transform()` to maintain API compatibility.

The `hansen.py` code file at the moment is missing its doc string, which I will add back.  Execution speed is good, but there is always room for improvement. In particular, at the moment a global variable is used so that `phi` is not recomputed. I am open to suggestions.
```python
[etalon:/home/stg121]$ ipython3
Python 3.6.5 |Anaconda custom (64-bit)| (default, Mar 29 2018, 18:21:58) 
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import abel

In [2]: im = abel.tools.analytical.SampleImage(1001)

In [3]: %timeit abel.hansen.hansen_transform(im.image)
81.4 ms ± 1.4 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %timeit abel.hansenlaw.hansenlaw_transform(im.image)
127 ms ± 949 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [5]: %timeit abel.dasch.three_point_transform(im.image)
14.7 ms ± 248 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

@DanHickstein if it is ok to swap `hansen.py` (with doc string) for `hansenlaw.py`, I will update all the calls, the examples, and readthedocs files.

stggh:
Hmm, a 2D image is not correctly mapped in `axis=0`. I will look into my programming error ...

stggh:
All now **working**. I forgot to add the sum axis `aim[:, col] = x.sum(axis=0)` in the conversion back to 2D.

BTW, `hansen_transform(image, hold_order=0) == hansenlaw_transform(image, full_image_odd_cols=False)`. i.e. for no -1/2 pixel shift => `hansen;aw` coding was correct. The additional -1/2 pixel shift (for the zero-order hold approximation) improved the transform-pair fit, but not as much as an (unshifted) first-order hold.

stggh:
> snowballed into a fundamental better ... HL transform method

Yes, until the first-order hold approximation came along, I was becoming concerned that this time-consuming code-restructure process was gaining very little.

> "half-pixel shift" issue has effectively been solved through the "first order hold" approach

Yes.

> ask Eric 

Will do.

> fix the global variable issue

Work in progress.

@DanHickstein thanks for your great help in stepping through the intricacies of the `hansen(law)` method, with me, and in particular, initiating the contact with Eric Hansen, who kindly explained the detail of his 1985 algorithms to us.

stggh:
Eric is a fast responder:
> I am completely OK with keeping it as "hansenlaw" in the API, etc, and sharing the credit with Phaih-Lan.  Just cite both articles in your paper and in the comments for the code.   
Thanks for thinking of it.
Eric

DanHickstein:
Awesome! That saves us some re-organizing :)

DanHickstein:
@stggh, Thanks again for all of your great work on this PR!

What is the current status? Are there outstanding issues? Are you waiting for code review?

stggh:
**Summary of this PR**
This long PR and #206 started as the annoying difference between the `hansenlaw` method for the `Curve A` test case and what was published in the [Hansen and Law paper](https://doi.org/10.1364/JOSAA.2.000510):
image
image

It turned out that most of the difference was due to a [grid indexing error](https://github.com/PyAbel/PyAbel/pull/211#issuecomment-370195927).  However, the mean-square-error `mse` remained larger than stated on the Hansen and Law Figure:
curvea-mse00 00
```
  forward method     mean-squared error
  --------------     ------------------
    hansenlaw             5.89e-05
    direct                2.69e-06

  inverse method     mean-squared error
  --------------     ------------------
    hansenlaw             3.26e-06
    direct                2.59e-06
    three_point           5.93e-09
```

This difference could be ameliorated by a horizontal pixel shift of the driving function (gradient for an `inverse` transform, or the raw image, for the `forward` transform), which also shifts the transform curve. For `CurveA` (`TransformPair(profile=3)`) a pixel shift of `-0.5` gives a good fit to the transform pair:

curvea-mse0-0 50

```
python CurveA-mse.py

  forward method     mean-squared error
  --------------     ------------------
    hansenlaw             1.57e-06
    direct                2.69e-06

  inverse method     mean-squared error
  --------------     ------------------
    hansenlaw             3.26e-06
    direct                2.59e-06
    three_point           5.93e-09
```
The `-1/2` pixel shift hinted at a coding indexing error, and different sub-pixel shift values worked better for other test-curves, `CurveB` (`TransformPair(profile=8)`):
curvea-mse0-0 50

curveb-mse0-0 35

A `sub_pixel_shift` parameter was not very satisfying, because its value is not well determined.

In the mean time, a second Hansen paper was uncovered (discguised by the Hankel transform title), and Eric Hansen kindly helped us [understand it](https://github.com/PyAbel/PyAbel/pull/211#issuecomment-379648491).  The original `hansenlaw` algorithm assumed the driving function is constant across a grid interval (a pixel), called *staircase* or *zero-order hold* approximation.  The second paper, extended this to linear function, the *first-order hold* approximation.

The *first-order* algorithm does a **great** job for the transform pair test profiles:
curvea-mse10 00
mse = 7.24e-7
curveb-mse10 00
(not as good as the pixel shift case)

It turns out that the *first-order hold* algorithm does not replace the *zero-order hold* algorithm. For real image data, the *zero-hold* method, **with** sub-pixel-shift, the alignment is better:
o2-

This makes sense in that image data consists of pixels and so the *zero-order hold* for the driving function is correct. While for the transform pair curves the *first-order hold* is better.

The `hansenlaw` `PyAbel` method has been re-written to provide for the traditional *zero-order* method `hold_order=0`, with sub-pixel-shift `sub_pixel_shift=-0.35` (default) and the *first-order* method `hold_order=1`, in this case no pixel shift is applied (`sub_pixel_shift` is ignored.

In conclusion, for the `hansenlaw` method with image data, we are stuck with the *zero-order* method and a sub-pixel-shift, to better align the transform data. For non pixelated data , the *first-order* method may be better. YMMV 

In the end, the `three_point` provides the faster more accurate inverse Abel transform, however, for forward transforms one must use either `hansenlaw` or `direct`.

DanHickstein:
Great summary @stggh! And great work on this PR. 

I merged #214, so you can now incorporate that into this PR.

I will proofread the code this weekend and then we can get this merged.

stggh:
rebase, brought a great deal of code change baggage ...

Last and most useful [Summary](https://github.com/PyAbel/PyAbel/pull/211#issuecomment-407934242)

abel/hansenlaw.py
```diff
@@ -5,169 +5,201 @@
...
+        .         |   *          |    *   |  <---------- im
+        .         |  *           |     *  |
+        .         +--------      o--------+
+        .         |  *           |     *  |
+        .         |   *          |    *   |
+        .         |     *        | *      |
+        .         +--------      +--------+
+
+        In accordance with all PyAbel methods the image center ``o`` is
+        defined to be mix-pixel i.e. an odd number of columns, for the
```
DanHickstein: "mix-pixel" should be "mid-pixel"

abel/hansenlaw.py
```diff
@@ -5,169 +5,201 @@
...
+        Right-side half-image (or quadrant). See figure below.
 
     dr : float
-        sampling size (=1 for pixel images), used for Jacobian scaling
-
-    direction : string ('forward' or 'inverse')
-        ``forward`` or ``inverse`` Abel transform
+        Sampling size (=1 for pixel images), used for Jacobian scaling.
+
+    direction : string 'forward' or 'inverse'
```
DanHickstein: state that default is 'inverse'

abel/hansenlaw.py
```diff
@@ -5,169 +5,201 @@
...
+        ``forward`` or ``inverse`` Abel transform.
+
+    hold_order : int 0 or 1
+        First- or zero-order hold approximation used in the evaluation of
+        state equation integral. `0` (default) gives the same result as the
+        original implementation of the `hansenlaw` method. It is applicable
+        for pixel images, where the intensity of the driving function
```
DanHickstein: I'm not exactly sure what this sentence is trying to say...

abel/hansenlaw.py
```diff
@@ -5,169 +5,201 @@
...
+        (gradient for the inverse transform, original image for the forward
+        transform).  `1` assumes a linear intensity variation across
+        a 'pixel'. This may yield a more accurate transform for continuoulsy
+        varying functions.
```
DanHickstein: we should state that the default hold order is zero

abel/hansenlaw.py
```diff
@@ -5,169 +5,201 @@
...
+
+    sub_pixel_shift : float -0.35 default
```
DanHickstein: I think that the default value is typically listed at the end of the description of that argument.

abel/hansenlaw.py
```diff
@@ -5,169 +5,201 @@
...
+        For the zero-order hold approximation `hold_order=0`, a sub-pixel
+        left-shift of the driving function (image-forward or gradient-inverse)
+        improves the transform alignment with the other PyAbel methods,
+        and Abel transform-pair functions.  See the discussion in/conclusions
+        of `PR #211 <https://github.com/PyAbel/PyAbel/pull/211>`_
```
DanHickstein: "Default: -0.35" should go here.

doc/transform_methods/hansenlaw.rst
```diff
@@ -8,20 +8,21 @@ Hansen-Law
 Introduction
 ------------
 
-The Hansen and Law transform is the work of E. W. Hansen and P.-L. Law [1].
+The Hansen and Law transform [1, 2] is a fast (linear time) Abel transform.
```
DanHickstein: What does "linear time" mean? We know that they are "linear", in the sense of linear algebra, but do we know that they are linear with time? I think that we showed that the HL algorithm scales like N^2.x, so it's pretty close to linear with the total number of pixels, but probably not quite linear.

doc/transform_methods/hansenlaw.rst
```diff
@@ -141,13 +132,71 @@ Historical Note
...
+  \int_{r_0}^{r} \left(\frac{\epsilon}{r}\right)^\lambda_k d\epsilon = \frac{r}{r_0}\left[ 1 - \left(\frac{r}{r_0}\right)^{\lambda_k + 1}\right] = \frac{(n-1)^a}{\lambda_k + a} \left[ 1 - \left(\frac{n}{n-1}\right)^{\lambda_k+a} \right],
+
+where, the right-most-side of the equation has an additional parameter, :math:`a` to generalize the power of :math:`\lambda_k`.  For the inverse transform, there is an additional factor :math:`\frac{1}{\pi r}` in the state equation, and hence the integrand has :math:`\lambda_k` power, reduced by -1. While, for the 
+first-order hold approximation, the linear :math:`\epsilon` term increases :math:`\lambda_k` by +1. 
 
 
 Citation
 --------
-[1] `E. W. Hansen and P.-L. Law, "Recursive methods for computing the Abel transform and its inverse", J. Opt. Soc. A2, 510-520 (1985) <http://dx.doi.org/10.1364/JOSAA.2.000510>`_
+[1] `E. W. Hansen and P.-L. Law, "Recursive methods for computing the Abel transform and its inverse*, J. Opt. Soc. A2, 510-520 (1985). <http://dx.doi.org/10.1364/JOSAA.2.000510>`_
```
DanHickstein: Should the asterisk after "inverse" actually be a quotation mark?

doc/transform_methods/hansenlaw.rst
```diff
@@ -141,13 +132,71 @@ Historical Note
...
+
+[2] `E. W. Hansen, "Fast Hankel Transform", IEEE Trans. Acoust. Speech Signal Proc. 33, 666 (1985). <https://dx.doi.org/10.1109/TASSP.1985.1164579>`_
+
+[3] J. R. Gascooke, PhD Thesis: *"Energy Transfer in Polyatomic-Rare Gas Collisions and Van Der Waals Molecule Dissociation"*, Flinders University (2000).
```
DanHickstein: Should the rest of the titles also be in italics?

doc/transform_methods/hansenlaw.rst
```diff
@@ -8,20 +8,21 @@ Hansen-Law
 Introduction
 ------------
 
-The Hansen and Law transform is the work of E. W. Hansen and P.-L. Law [1].
+The Hansen and Law transform [1, 2] is a fast (linear time) Abel transform.
+ 
+From the abstract [1]:
```
DanHickstein: This doesn't read too well, since the "...new family of algorithms" isn't the start to a new sentence. Maybe we can make the transition smoother by writing something like: 
"In their words, Hansen and Law [1] present"

DanHickstein:
A quick screenshot showing how the agreement of the HL transform with the analytical result is greatly improved by this PR:

image

DanHickstein:
BTW, did you want to include the reference to the Gascooke paper in the docstring for the circularize method? (discussed in https://github.com/PyAbel/PyAbel/issues/210) https://aip.scitation.org/doi/10.1063/1.4981024

abel/hansenlaw.py
stggh: Yes, sorry, I terminated the '0' sentence too early. Now re-written

doc/transform_methods/hansenlaw.rst
stggh: A very good point. I like your wording. Thanks.

doc/transform_methods/hansenlaw.rst
stggh: The markup url quote inhibits any further markup. I have changed the Gascooke reference to link to the `https://github.com/PyAbel/abel_info` web page. It now looks similar.

stggh:
Thanks @DanHickstein, all good suggestions. 

> include the reference to the Gascooke paper

Already present in this PR in `hansenlaw.rst`
screen shot 2018-07-29 at 14 37 39

> HL transform with the analytical result

Is this for the case `hold_order=1` ?

stggh:
> include the reference ... in the docstring for circularize_image

Sorry, for my confusion. The citation is already present in `abel.tools.circularize.circularize_image`?
```python
In [1]: import abel

In [2]: abel.tools.circularize.circularize_image?

Signature: abel.tools.circularize.circularize_image(IM, method='lsq', center=Non
e, radial_range=None, dr=0.5, dt=0.5, smooth=0, ref_angle=None, inverse=False, r
eturn_correction=False)
Docstring:
Corrects image distortion on the basis that the structure should be
circular.

This is a simplified radial scaling version of the algorithm described in 
`J. R. Gascooke and S. T. Gibson and W. D. Lawrance: 'A "circularisation"
method to repair deformations and determine the centre of velocity map 
images' J. Chem. Phys. 147, 013924 (2017).
<https://dx.doi.org/10.1063/1.4981024>`_
```

> default behavior, which is hold_order = 0

Yes, the deviation, for this analytical function, should/will be less with `hold_order=1`

abel/hansenlaw.py
```diff
@@ -164,7 +164,8 @@ def I(n, lam, a):
     rows, cols = image.shape
 
     if direction == 'forward':
-        drive = -2*dr*np.pi*image.copy()  # include Jacobian
+        # copy image for the driving function, include Jacobian factor,
+        drive = -2*dr*np.pi*np.array(image)
```
DanHickstein: Does np.array make a copy just like np.copy?

stggh:
Yes `numpy.array(object, dtype=None, copy=True, order='K', subok=False, ndmin=0)`, the default is `copy=True`. `array` is slightly faster (although, I have misplaced the relevant web url).

DanHickstein:
Are we waiting on anything to merge this?

stggh:
No, just the final nod! I am happy to press the green button.

DanHickstein:
Hit it! :)

Issue #213, opened 2018-03-13
Make direct operate on a "odd" grid


DanHickstein:
While thinking about #211, where we are realizing the the Hansen-Law algorithm operates on an "even" grid (see #39 for grid explanation), I realized that the Direct method also defaults to an "even" grid if the `r` parameter is not specified.

This is because of the factor of 0.5 in [line 51](https://github.com/PyAbel/PyAbel/blob/master/abel/direct.py#L51) of direct.py:
```python
r = (np.arange(n) + 0.5)*dr
```
I think that we can change the grid be be "odd" if we just remove the factor of 0.5. Here is a comparison of the direct inverse transform to an analytical gaussian (assuming an "odd" grid). Left is the current behavior, right is the behavior with the factor of 0.5 removed, which is way better: 
![image](https://user-images.githubusercontent.com/1107796/37357151-45708ac0-26ad-11e8-84e0-e664e0c1fa18.png)

DhrubajyotiDas:
I'm pretty surprised by how sensitive these transforms are to an even/odd grid.

stggh:
@DanHickstein you are correct!

For curveA `profile3`:
```python
n = 101
f = abel.tools.analytical.TransformPair(n=n, profile=3)
f.r[0] = 1.0e-16   # profile3 requires 0 < r <= 1
 :
dr = abel.direct.direct_transform(f.abel, r=f.r)
dr[0] = dr[1]  # rig first value
```
![a](https://user-images.githubusercontent.com/10932229/37372971-d8c49436-2768-11e8-87ff-cdd85945f0e3.png)

stggh:
Apart from `direct` grid not appearing to like `r[0] = 0`
Set to `1.0e-16` in this example:

```python
n = 1001
n2 = n//2
name = 'dribinski'

source = abel.tools.analytical.SampleImage(n, name=name).image
half = source[:, n2:]

halfshift = interpolation.shift(half, (0, -1/2))

# proj = abel.hansenlaw.hansenlaw_transform(halfshift, direction='forward')
proj = abel.direct.direct_transform(half, direction='forward')

projshift = interpolation.shift(proj, (0, -1/2))

hl = abel.hansenlaw.hansenlaw_transform(projshift)
di = abel.direct.direct_transform(proj)

plt.plot(half[n2], label='source')
plt.plot(hl[n2], label='hansenlaw')
plt.plot(di[n2], label='direct')
[...]

PR #214, opened 2018-03-15
Fixed the grid for the direct method

DanHickstein:
As discussed in #213 (which followed from our discussion in #211), we decided that the `direct` method was using a "even" grid (`[0.5, 1.5, 2.5, ...]`). I switched one line so that it now operates on an "odd" grid (`[0, 1.0, 2.0, ...]`). 

I also cleaned up a few lines of comments towards the top of `direct.py`.

### DIFF
--- a/abel/direct.py
+++ b/abel/direct.py
@@ -18,6 +18,7 @@ except (ImportError, UnicodeDecodeError):
 # numerical integration
 #
 # Roman Yurchak - Laboratoire LULI, Ecole Polytechnique/CNRS/CEA, France
+# 07.2018: DH fixed the correction for the case where r[0] = 0
 # 03.2018: DH changed the default grid from 0.5, 1.5 ... to 0, 1, 2.
 # 01.2018: DH dhanged the integration method to trapz
 # 12.2015: RY Added a pure python implementation
@@ -48,7 +49,6 @@ def _construct_r_grid(n, dr=None, r=None):
         if isinstance(dr, np.ndarray):
             raise NotImplementedError
         r = (np.arange(n))*dr
-        r[0] = 1e-16 # avoids returning np.nan
     return r, dr
 
 
@@ -219,8 +219,18 @@ def _pyabel_direct_integral(f, r, correction, int_func=np.trapz):
         f_r = (f[:, 1:] - f[:, :N1-1])/np.diff(r)[None, :]
 
         for i, row in enumerate(f):  # loop over rows (z)
-            out[i, :-1] += I_sqrt[II+1 == JJ]*f_r[i] \
+            corr  = I_sqrt[II+1 == JJ]*f_r[i] \
                     + np.arccosh(r[1:]/r[:-1])*(row[:-1] - f_r[i]*r[:-1])
+
+            
+            if np.isclose(r[0], 0):  # special case for r[0] = 0 
+                corr0 = I_sqrt[II+1 == JJ][0]*f_r[i,0] \
+                        + (row[0] - f_r[i,0]*r[0])
+                out[i, 1:-1] += corr[1:]
+                out[i, 0]    += corr0 
+            else:
+                out[i, :-1] = corr
+            
 
     return out
 
--- a/abel/direct.py
+++ b/abel/direct.py
@@ -218,7 +218,7 @@ def _pyabel_direct_integral(f, r, correction, int_func=np.trapz):
         f_r = (f[:, 1:] - f[:, :-1])/np.diff(r)[None, :]
         isqrt = I_sqrt[II+1 == JJ]
 
-        if np.isclose(r[0], 0):  # special case for r[0] = 0
+        if r[0]<(r[1]*1e-8):  # special case for r[0] = 0
             ratio = np.append(np.cosh(1), r[2:]/r[1:-1])
         else:
             ratio = r[1:]/r[:-1]
--- a/abel/direct.py
+++ b/abel/direct.py
@@ -18,12 +18,11 @@ except (ImportError, UnicodeDecodeError):
 # numerical integration
 #
 # Roman Yurchak - Laboratoire LULI, Ecole Polytechnique/CNRS/CEA, France
-#
-# 01.2018: Changed the integration method to trapz
-# 12.2015: Added a pure python implementation following a dissuasion
-#                                                     with Dan Hickstein
-# 11.2015: Moved to PyAbel, added more unit tests, reorganized code base
-#    2012: First implementation in hedp.math.abel
+# 03.2018: DH changed the default grid from 0.5, 1.5 ... to 0, 1, 2.
+# 01.2018: DH dhanged the integration method to trapz
+# 12.2015: RY Added a pure python implementation
+# 11.2015: RY moved to PyAbel, added more unit tests, reorganized code base
+#    2012: RY first implementation in hedp.math.abel
 ###########################################################################
 
 
@@ -42,13 +41,13 @@ def _construct_r_grid(n, dr=None, r=None):
             raise ValueError('The input parameter r should be a 1D array'
                              'of shape = ({},), got shape = {}'.format(
                                                                 n, r.shape))
-        # not so sure about this, needs verification
+        # not so sure about this, needs verification -RY
         dr = np.gradient(r)
 
     else:
         if isinstance(dr, np.ndarray):
             raise NotImplementedError
-        r = (np.arange(n) + 0.5)*dr
+        r = (np.arange(n))*dr
     return r, dr
 
 
--- a/abel/direct.py
+++ b/abel/direct.py
@@ -48,6 +48,7 @@ def _construct_r_grid(n, dr=None, r=None):
         if isinstance(dr, np.ndarray):
             raise NotImplementedError
         r = (np.arange(n))*dr
+        r[0] = 1e-16 # avoids returning np.nan
     return r, dr
 
 
--- a/abel/direct.py
+++ b/abel/direct.py
@@ -179,7 +179,6 @@ def _pyabel_direct_integral(f, r, correction, int_func=np.trapz):
     else:
         int_opts = {'x': r}
 
-    N1 = f.shape[1]
     out = np.zeros(f.shape)
     R, Y = np.meshgrid(r, r, indexing='ij')
 
--- a/abel/direct.py
+++ b/abel/direct.py
@@ -124,11 +124,7 @@ def direct_transform(fr, dr=None, r=None, direction='inverse',
     r, dr = _construct_r_grid(f.shape[1], dr=dr, r=r)
 
     if direction == "inverse":
-        # a derivative function must be provided
-        f = derivative(f)/dr
-        # setting the derivative at the origin to 0
-        # f[:,0] = 0
-    if direction == "inverse":
+        f = derivative(f)/dr 
         f *= - 1./np.pi
     else:
         f *= 2*r[None, :]
@@ -209,28 +205,28 @@ def _pyabel_direct_integral(f, r, correction, int_func=np.trapz):
 
 
     """
-    Compute the correction
-    Pre-calculated analytical integration of the cell with the singular value
-    Assuming a piecewise linear behaviour of the data
+    Compute the correction. Here we apply an
+    analytical integration of the cell with the singular value,
+    assuming a piecewise linear behaviour of the data.
+    The analytical abel transform for this trapezoid is:
     c0*acosh(r1/y) - c_r*y*acosh(r1/y) + c_r*sqrt(r1**2 - y**2)
+    see: https://github.com/luli/hedp/blob/master/hedp/math/abel.py#L87-L104
     """
     if correction == 1:
-        # computing forward derivative of the data
-        f_r = (f[:, 1:] - f[:, :N1-1])/np.diff(r)[None, :]
+
+        # precompute a few variables outside the loop:
+        f_r = (f[:, 1:] - f[:, :-1])/np.diff(r)[None, :]
+        isqrt = I_sqrt[II+1 == JJ]
+
+        if np.isclose(r[0], 0):  # special case for r[0] = 0
+            ratio = np.append(np.cosh(1), r[2:]/r[1:-1])
+        else:
+            ratio = r[1:]/r[:-1]
+        
+        acr = np.arccosh(ratio)
 
         for i, row in enumerate(f):  # loop over rows (z)
-            corr  = I_sqrt[II+1 == JJ]*f_r[i] \
-                    + np.arccosh(r[1:]/r[:-1])*(row[:-1] - f_r[i]*r[:-1])
-
-            
-            if np.isclose(r[0], 0):  # special case for r[0] = 0 
-                corr0 = I_sqrt[II+1 == JJ][0]*f_r[i,0] \
-                        + (row[0] - f_r[i,0]*r[0])
-                out[i, 1:-1] += corr[1:]
-                out[i, 0]    += corr0 
-            else:
-                out[i, :-1] = corr
-            
+            out[i, :-1] += isqrt*f_r[i] + acr*(row[:-1] - f_r[i]*r[:-1])
 
     return out
 
--- a/README.rst
+++ b/README.rst
@@ -119,6 +119,40 @@ Documentation
 General information about the various Abel transforms available in PyAbel is available at the links above. The complete documentation for all of the methods in PyAbel is hosted at https://pyabel.readthedocs.io.
 
 
+Conventions
+-----------
+
+The PyAbel code adheres to the following conventions:
+
+- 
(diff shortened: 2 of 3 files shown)


DanHickstein:
Ah, I see what you mean now @stggh about needing to ensure that `r[0]` isn't exactly zero.

The tests all pass when I turn off the "correction", so I think that the issue is line 223 of direct.py:
```python
        for i, row in enumerate(f):  # loop over rows (z)
            out[i, :-1] += I_sqrt[II+1 == JJ]*f_r[i] \
                    + np.arccosh(r[1:]/r[:-1])*(row[:-1] - f_r[i]*r[:-1])
```
If r[0] is zero, then we have a divide-by-zero situation in the argument of the `arccosh` function. I don't quite understand the correction well enough to understand what modification is needed. Probably, we just need to avoid applying it to the first column in the event that r[0] == 0.

But, just setting r[0] to some very small number appears to be an adequate solution for the time being.

stggh:
Not quite, setting `r[0] ~ 0` gives a large value for the transform `out[0]` (tested for curveA), so we need to tidy it up, perhaps setting `out[:, 0] = out[:, 1]` as per the `hansenlaw` method, or understanding the code better ;-)

DanHickstein:
Yeah, I think that we just need to figure out what to do for the correction in the special case where `r[0] = 0`. This should be entirely possible.

@rth, do you have any notes on how the "correction" function was implemented?

In the following, the equation in the comment doesn't seem to match the equation in the code. 


```python
    """
    Compute the correction
    Pre-calculated analytical integration of the cell with the singular value
    Assuming a piecewise linear behaviour of the data
    c0*acosh(r1/y) - c_r*y*acosh(r1/y) + c_r*sqrt(r1**2 - y**2)
    """
    if correction == 1:
        # computing forward derivative of the data
        f_r = (f[:, 1:] - f[:, :N1-1])/np.diff(r)[None, :]

        for i, row in enumerate(f):  # loop over rows (z)
            out[i, :-1] += I_sqrt[II+1 == JJ]*f_r[i] \
                    + np.arccosh(r[1:]/r[:-1])*(row[:-1] - f_r[i]*r[:-1])
```

stggh:
Quick ugly fix?
1. line 51: drop +0.5
```python
        r = np.arange(n)*dr
```
2. line 221+:    If `r[0] ~0`, don't apply the correction to the first column. Make 1st column = 2nd
```python
        if r[0] > 1.0e-16:  # non-zero value grid
            for i, row in enumerate(f):  # loop over rows (z)
                out[i, :-1] += I_sqrt[II+1 == JJ]*f_r[i] \
                               + np.arccosh(r[1:]/r[:-1])\
                               *(row[:-1] - f_r[i]*r[:-1])
        else:  # no evaluation for 1st column, r=0
            for i, row in enumerate(f):  # loop over rows (z)
                out[i, 1:-1] += (I_sqrt[II+1 == JJ]*f_r[i])[1:]\
                                + np.arccosh(r[2:]/r[1:-1])\
                                *(row[:-1] - f_r[i]*r[:-1])[1:]
            out[:, 0] = out[:, 1]
```

DanHickstein:
@stggh, do you understand the math for the correction? I think that if we could re-derive it, then we could understand how to fix it properly to work in all cases.

stggh:
Agreed!  We need @rth's input or a reference to the method implementation.

rth:
> do you understand the math for the correction? I think that if we could re-derive it, then we could understand how to fix it properly to work in all cases.

The integration for the last cell was done symbolically [here](https://github.com/luli/hedp/blob/master/hedp/math/abel.py#L87-L104). There may be better ways of doing it though...

stggh:
This is neat, sympy does produce:
```python

In [6]: P = c0 + (r-y)*c_r #+ (r-r0)**2*c_rr

In [7]: K_d = 1/sqrt(r**2-y**2)

In [8]: res = integrate(P*K_d, (r,y, r1))

In [9]: sres= simplify(res)

In [10]: sres
Out[10]: c0*acosh(r1/y) - c_r*y*acosh(r1/y) + c_r*sqrt(r1**2 - y**2)
```
which is the function listed in the `direct.py` code comment, 
image
but, as @DanHickstein [noted](https://github.com/PyAbel/PyAbel/pull/214#issuecomment-373781032)  not the (clever) function coded.

It appears to be something along the lines of (where |_n means evaluated at n):
eq1

However, the code obviously works.

rth:
> In the following, the equation in the comment doesn't seem to match the equation in the code.
> but, as @DanHickstein noted not the (clever) function coded.

Hah, to be honest, I have not touched this code in years. I'm also not certain it was cleanly implemented in the first place. I do remember that `I_sqrt` computation was factorized out of the loop for performance reasons (but maybe more in the regular integration loop rather than the correction). 

Feel to re-write the whole thing if needed since the current implementation is obviously not too readable. Sorry for not being of more help..

DanHickstein:
Hmm, I'm still confused about exactly what is happening, but I can get really good agreement for the first pixel (for the gaussian case, at least) if I set the `arccosh` term to `1`. I have no idea why this works. I would have thought that setting the arccosh term to pi might have been a solution, but setting it to 1 seems strange. 

Anyway, the test are currently failing because I haven't fixed the Cython case yet. But, if we think that this is a workable solution (please test!) then I can implement it in the Cython version as well.

abel/direct.py
```diff
@@ -219,9 +219,19 @@ def _pyabel_direct_integral(f, r, correction, int_func=np.trapz):
         f_r = (f[:, 1:] - f[:, :N1-1])/np.diff(r)[None, :]
 
         for i, row in enumerate(f):  # loop over rows (z)
-            out[i, :-1] += I_sqrt[II+1 == JJ]*f_r[i] \
+            corr  = I_sqrt[II+1 == JJ]*f_r[i] \
```
stggh: This fails for the case r == 0 because of r[1:]/r[:-1], so line 21 r[0] = 1.0e-16 is still required.

stggh: This code may need to be inside the else block?

abel/direct.py
```diff
@@ -219,9 +219,19 @@ def _pyabel_direct_integral(f, r, correction, int_func=np.trapz):
...
+            if np.isclose(r[0], 0):  # special case for r[0] = 0 
+                corr0 = I_sqrt[II+1 == JJ][0]*f_r[i,0] \
+                        + (row[0] - f_r[i,0]*r[0])
+                out[i, 1:-1] += corr[1:]
+                out[i, 0]    += corr0 
+            else:
+                out[i, :-1] = corr
+            
+
     return out
```
stggh: The original code is fine, together with `r[0]=1.0e-16`
```python
    if correction == 1:
        # computing forward derivative of the data
        f_r = (f[:, 1:] - f[:, :-1])/np.diff(r)[None, :]

        for i, row in enumerate(f):  # loop over rows (z)
            out[i, :-1] += I_sqrt[II+1 == JJ]*f_r[i] \
                    + np.arccosh(r[1:]/r[:-1])*(row[:-1] - f_r[i]*r[:-1])
```

However, this is not a general solution.

abel/direct.py
DanHickstein: `r[1:]/r[:-1]` does have a divide-by-zero for the first element when `r[0] = 0`. So, the first element of the `corr` array is `np.nan` when `r[0]` is `0`. But, on line 226, we check to see if `r[0]` is close to `0` and don't use the first element of `corr` in this case.

abel/direct.py
DanHickstein: Yeah, `r[0] = 1.0e-16` avoids returning any `np.nan` values, but if you check the result with the Gaussian case, `out[0]` is pretty far from the correct value. I think that my current code provides a pretty nice result for the gaussian test case, although for poorly understood reasons.

stggh:
As I mentioned in my code edits `np.arccosh(r[1:]/r[:-1])` fails for `r[0] = 0`.

`arccosh` soon asymptotes to a pretty flat function, so the approximation `arccosh ~ 1` for small `r` may be a good enough approximation:
arccosh

Issue: `r[0] = 0` **cannot** be corrected at line 51, e.g.
```python
        r = np.arange(n)*dr
        r[0] = 1.0e-16
```
because this creates an uneven grid, which bumps the back-end back to `Python`, **and** the first pixel (column) of the inverse Abel transform is also incorrect, although, not a big deal because the image noise is dumped here is any case.

I think that we should return to the original code:
```python    if correction == 1:
        # computing forward derivative of the data
        f_r = (f[:, 1:] - f[:, :-1])/np.diff(r)[None, :]

        for i, row in enumerate(f):  # loop over rows (z)
            out[i, :-1] += I_sqrt[II+1 == JJ]*f_r[i] \
                    + np.arccosh(r[1:]/r[:-1])*(row[:-1] - f_r[i]*r[:-1])
```
exclude the first pixel (first column) calculation, and set the first column = to the 2nd column, as is the case for the `hansenlaw` method.

abel/direct.py
stggh: The evaluation, although not used, creates a warning error

abel/direct.py
stggh: Ok. I don't think the exact value of inverse_image[:, 0] is important, since the `direct` method does not provide it, and the real-world value would be noisy, in any case. So, your fix is as good as any. However, the warning message from the evaluation of `np.arccosh(r[1:]/r[:-1])` is not a good look.

stggh:
I think this code does the right thing, has @DanHickstein's `arccosh=1` for `r[0]==0`, and it should be faster, because it pre-computes `ratio = r[1:]/r[:-1]` and extracts `I_sqrt[II+1 == JJ]` from the loop:
```python
    if correction == 1:
        f_r = (f[:, 1:] - f[:, :-1])/np.diff(r)[None, :]

        isqrt = I_sqrt[II+1 == JJ]

        if np.isclose(r[0], 0):  # special case for r[0] = 0
            ratio = np.append(np.pi/2, r[2:]/r[1:-1])
        else:
            ratio = r[1:]/r[:-1]

        for i, row in enumerate(f):  # loop over rows (z)
            out[i, :-1] += isqrt*f_r[i] + np.arccosh(ratio)*\
                           (row[:-1] - f_r[i]*r[:-1])

    return out
```

The top code may be tidied too:
```python
    if direction == "inverse":
        f = derivative(f)/dr
        f *= - 1/np.pi
    else:
        f *= 2*r[None, :]
```

My issues with nans were due to the C-backend for the forward transform that was used in one of my tests.

plot_example_direct_gaussian
r3

DanHickstein:
Thanks @stggh! I like your code. Similar to what I was doing, but more elegant, since we are not unnecessarily computing the first pixel in the case when we don't need to.



I looked more carefully at the error between the analytical gaussian and the direct transform using your code. As expected, there is a systematic error that decreases as the number of points becomes large. (In the plot below, I have exaggerated the error by multiplying it by 10. It's actually only a bit worse than the error from the two-point method, which I think is awesome for such a simple method.) You can see that the first pixel doesn't quite follow a smooth curve with the rest of the error values:

image

In your code, the value of `arc` (the `arccosh` function) at the first pixel is actually `1.02322747855`. This is because we are using hyperbolic trig functions rather than normal trig functions. If I set `arc[0]=1.0` then the error at the first pixel looks like a smooth extension of the curve:

image

We can achieve this by changing
```python
ratio = np.append(np.pi/2, r[2:]/r[1:-1])
```
to 
```python
 ratio = np.append(np.cosh(1), r[2:]/r[1:-1])
```
or 
```python
ratio = np.append(1.54308063482, r[2:]/r[1:-1])
```

If we think that this approach is the way forward, then I can implement this approach in the Cython code as well.

stggh:
Good work @DanHickstein. Ready to merge once you have composed the Cython code.

As I noted above, the last image column and last row are not defined by `direct` and remain at the initialized zero values. This is due to the use of `f_r = (f[:, 1:] - f[:, :-1])/np.diff(r)[None, :]`, which reduces the array size by 1,  instead of using `np.gradient()`. However, `gradient` assumes a fixed grid interval (I think). I am pretty sure that `direct` could be written more efficiently in `Python`, without the need for `Cython,` if the grid size is assumed constant. At the moment, `direct` is a novel `PyAbel` method, in that it can handle a non-linear grid, although, this is not tested. If the non-linear grid works, it could be pretty useful. One day, I would like to implement the same for `hansenlaw`.

abel/direct.py
```diff
@@ -209,28 +205,28 @@ def _pyabel_direct_integral(f, r, correction, int_func=np.trapz):
...
+    Compute the correction. Here we apply an
+    analytical integration of the cell with the singular value,
+    assuming a piecewise linear behaviour of the data.
+    The analytical abel transform for this trapezoid is:
     c0*acosh(r1/y) - c_r*y*acosh(r1/y) + c_r*sqrt(r1**2 - y**2)
+    see: https://github.com/luli/hedp/blob/master/hedp/math/abel.py#L87-L104
     """
     if correction == 1:
-        # computing forward derivative of the data
-        f_r = (f[:, 1:] - f[:, :N1-1])/np.diff(r)[None, :]
```
stggh: No need for `N1` it and its assignment may be deleted.  `f[:, -1]` is equally as good.

abel/direct.py
```diff
@@ -209,28 +205,28 @@ def _pyabel_direct_integral(f, r, correction, int_func=np.trapz):
...
-                corr0 = I_sqrt[II+1 == JJ][0]*f_r[i,0] \
-                        + (row[0] - f_r[i,0]*r[0])
-                out[i, 1:-1] += corr[1:]
-                out[i, 0]    += corr0 
-            else:
-                out[i, :-1] = corr
-            
+            out[i, :-1] += isqrt*f_r[i] + acr*(row[:-1] - f_r[i]*r[:-1])
 
     return out
```
stggh: The assignment, line 182, `N1 = f.shape[1]` may be deleted, since we now use `[:, -1]` instead of `[:, N1-1]`

DanHickstein:
In the latest commits, I did a few things:

1. Implemented a similar `arccosh = 1` for the case of `r[0]` near zero in the Cython code.

2. Realized that we shouldn't be checking if `r[0]` is near zero in an absolute sense, because some users may simply be using very small values for their r-grid. Instead, I now check if `r[0] < 1e-5 * r[1]`. Basically, we are asking if r[0] is very small compared to the typical grid spacing.

3. I also included the "conventions" text into README.rst, since it was the last lingering issue (https://github.com/PyAbel/PyAbel/issues/192) to allow us to release version v0.8.0. (Of course, we'll wait for #211 before we actually release the new verison.)

@stggh, you are entirely correct about out[-1] remaining zero in all cases. I looked at the code for a while, and I'm not sure how to address this issue at this point. Typically, users should be supplying images that have the last pixel with a value of zero, in which case we are doing the right thing. If a user supplies an image where the edge pixel is not zero, then I guess we should be assuming that there is a pixel beyond the edge where the value drops to zero. It's a basic assumption of the numerical Abel transform that there are not pixels outside of the supplied range that have non-zero values, right?

Perhaps we can simply extend the input image by one pixel, which is set to a value of zero. Then we can clip the last pixel from the resulting output. Perhaps we should merge this PR sooner rather than later, and open a new issue to remind us to look into this "last pixel" issue? 

For the record, I think that `np.gradient` can handle an uneven grid spacing: https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.gradient.html

stggh:
@DanHickstein changes to `direct.py` and `direct.pyx` look good. Here is a curveA test, with mse values:
curvea-mse10 00

Ready to merge, but note, line 182 `N1 = f.shape[1]` is not used, and may be deleted.

> assuming that there is a pixel beyond the edge where the value drops to zero

I agree with your arguments, the outer edge pixels should be zero, and one column, row will not make much difference. `hansenlaw` does this, but it is really not much better than leaving the value zero:
```python
    # missing column at each side of image
    aim[:, 0] = aim[:, 1]
    aim[:, -1] = aim[:, -2]
```

> np.gradient can handle an uneven grid spacing
You are correct. However, 
```python
f_r = np.gradient(f[:, :-1], np.repeat([np.diff(f)], f.shape[0], axis=0))
```
(does not work, and) does not look any better than the original. However, it may help speed up the Python code, it we come back to this issue.

DanHickstein:
Okay, I removed the now necessary line 182 (apologies for missing your comment about this earlier).

I'm going to merge now, but I think that you are probably right that it's best to pad the array by one pixel in order to allow non-zero values in the last pixel. However, I'm not sure if it's best to assume that the pixel outside of the user-provides array has the same value as the adjacent pixel, or if it decays to zero. I think that the most logical assumption is that, if a user provides a non-zero edge pixel, the only logical thing PyAbel can assume is that the intensity drops to zero at the next pixel, because otherwise the Abel transform would not be valid, since it requires knowledge of all non-zero pixel at a larger y (or r) value.

Along these same lines, when you are generating your MSE plots using a Gaussian function, you may want to use a smaller sigma for the gaussian, to ensure that the function has fully decayed to near zero at the edge of the array. I am curious if the MSE would drop in your plots if the gaussian was closer to zero at the edge of the plot.

PR #219, opened 2018-08-08
Incrementing version to 0.8.0 and including py2 bz2 support

DanHickstein:
Here I'm incrementing the version to 0.8.0, in anticipation of a new release (we closed all of the issues for v0.8, hurray!).

Also, I've changed the way that the `O2-ANU1024.txt.bz2` file is loaded. I'm not too excited about how this makes the examples more complicated, but I'm not sure what the easier solution is. Perhaps we can just use a text file, but then the file becomes several MB. Alternatively, we could make the file a PNG, but then @stggh would need to accept some degradation of his beautiful image. 

Any other ideas?


stggh:
Your solution is fine. The best solution is to drop support for Python2.7 which has EOL 2020. ;-)

It would save a great deal of baggage, and allowing the use of some very neat Python3.6+ features, such as [f-strings](https://www.python.org/dev/peps/pep-0498/)

DanHickstein:
That's interesting about the f-strings! Being a loyal Py2 user, I hadn't heard of them :).

But you are right, Py2 is supposed to be phased out soon. Perhaps we should switch to Py2-only in v0.9 or v1.0. In the meanwhile, I'll open a new Issue where we can keep track of all of simplification that we can make when we switch to Py3 only.

DanHickstein:
Alright, if there are no objections, I'll go ahead and merge and make the release!

PR #228, opened 2018-09-23
Migrate to pytest

rth:
nose is no longer maintained (and hasn't been in the last 3 years), this migrates testing to its successor [pytest](https://docs.pytest.org/en/latest/).

It's mostly sufficient to install pytest instead of nose for this to work: tests themselves don't have to be changed.

Also this removes `abel/tests/run.py` this was some old code by me to run tests, but it's really no longer necessary with pytest.


DanHickstein:
This makes sense to me! Let's merge after we merge #227

Issue #231, opened 2018-10-02
dasch cached basis implementation


stggh:
An implementation of @MikhailRyazanov's `basex` caching and slicing, for the `dasch` methods is in my PyAbel fork [dasch-cached-memory-basis](https://github.com/stggh/PyAbel/tree/dasch-cached-memory-basis)

This issue provides discussion point on the best PyAbel coding practices.

1) Implementation
   @MikhailRyazanov's clever ideas:
      a) caching the basis numpy array, for faster repeat transforms (rather than reading the basis file)
      and 
      b) slicing the basis from a larger basis (cached or file)

   Are easy to implement:
      a) add global variables `_basis = None`, as per `basex`, and a  `_method = None` to identify the Dasch method, to `abel.dasch.py`
     and 
      b) add slicing to the cached basis  `_basis[:cols, :cols]`, to extract from any larger basis.

      ```python
      # cache basis
      _basis = None
      _method = None
         :
      _basis = abel.tools.basis.get_bs_cached(method, cols, cols,
                                            basis_dir=basis_dir,
                                            cached_basis=(_basis, _method),
                                            verbose=verbose)
     _method = method
     ```

      `get_bs_cached()` is common to all the basis methods, and was extracted to `abel.tools.basis.py` during coding of the `abel.dasch.py` methods. 

     `abel.tools.basis.py` became somewhat untidy when the `linbasex` method was included, since that method broke the simple `method_basis_{cols}_{nbf}.npy` naming scheme, as uniqueness required additional variables.  The use of `abel.tools.basis.py` requires the `dasch` global variables to be passed to `abel.tools.get_bs_cached(), in this fork, as a tuple`

     ```python
     if cached_basis is not None and cached_basis[0] is not None:
         _basis, _method = cached_basis
         if _basis.shape[0] >= cols and _method == method:
             if verbose:
                 print('Using memory cached basis')
             return _basis
        :
        basis_files = glob.glob(path_to_basis_files)
        for bf in basis_files:
            if int(bf.split('_')[-2]) >= cols:  # relies on file order
                if verbose:
                    print("Loading {:s} basis {:s}".format(method, bf))
                D = np.load(bf)
                # trim to size
                return D[:cols, :nbf]
      ```   

   **Q1:** Given the mess caused by `linbasex` is it better to return `get_bs_cached()` into `abel.dasch.py`, as is the case for `abel.basex.py`?

   **Q2:** Any alternative suggestions?

[...]

Issue #230, opened 2018-09-26
Investigate basis issues in BASEX


MikhailRyazanov:
There are several problems in the `basex` method related to the underlying basis sets.
1. `nbf` ≠ `n` is currently not implemented. The utility of `nbf` < `n` or `nbf` > `n` is questionable, but the original BASEX had this capability, and it worked.
2. The mystery of the `MAGIC_NUMBER`, see #4. Currently, it is some hard-coded heuristic scaling factor, although it should have some reasonable mathematical origin and an analytical expression (probably, with basis-width dependence).
3. Something strange happens near _r_ = 0. See https://github.com/PyAbel/PyAbel/issues/225#issuecomment-422992673:
![profile8](https://user-images.githubusercontent.com/10932229/45787421-95524000-bcb8-11e8-96fb-a9b826ca3029.png)

MikhailRyazanov:
Some thoughts about point 3.

The image basis functions look like this:
![basis](https://user-images.githubusercontent.com/12854133/46057071-489cb800-c143-11e8-86b7-1da68e56f832.png)
Colored lines show functions _ρ<sub>k</sub>_(_r_) centered at _r_ = _k_, and black is their sum (over _k_ < 10). Dashed are the actual functions, and the solid lines represent what is in the basis matrix (they are really just points at integer _r_ values).

Two simple observations: the _k_ = 0 function is broader than all other, and all other functions are zero at _r_ = 0.

Thus the sum at _r_ = 0 is lower than anywhere else (no contributions from _k_ > 0), and at small _r_ > 0 it is higher than anywhere else (the tail of _k_ = 0).

In general, this looks quite similar to the behavior seen in the example above. The problem is that in principle these basis functions are still linearly independent and cover the full _r_ range, so to work correctly they do not need to be normalized (or to be orthogonal, or whatever else), as long as their projections are correct.

The projections in the article apparently have a wrong factor: it is "2" in (15), but my calculations show that (at least for _k_ = 0) it should be π<sup>1/2</sup>. By the way, 2/π<sup>1/2</sup> ≈ 1.128, which is relatively close but not equal to `MAGIC_NUMBER = 1.112...`. However, that "wrong" factor is used for all projections, so it should produce wrong normalization, but not the observed artifact. Unless the projected-basis computation has errors?

MikhailRyazanov:
Here are the basis projections:
[...]

Issue #224, opened 2018-08-19
linbasex intensity scaling


stggh:
`linbasex` intensity scaling is incorrect, it does not agree with the other methods, and as @DanHickstein has found appears to be dependent on the number of Legendre polynomials orders.

DanHickstein:
> The required scaling factor is 1/2r

Okay, problem solved! Nice work!

> I have pushed the 1/2r factor, linbasex.py line 267, and set the default smoothing to zero, in the current PyAbel PR #232.

Great, some motivation for us to merge #232 :)

> linbasex transform alignment, relative to the other methods

That's annoying that there is this small shift. Can you subtract the 2D transform (say linbasex versus three-point) so that we can see if the shift is symmetric, or just in horizontal?

stggh:
I am looking into the `linbasex` code, which, I must admit I never fully understood during the transfer process of Thomas Geber's jupyter notebook code :-(. However, now it is starting to make sense. It looks possible to switch to `1/2`-image transforms, and associated basis calculations. In the process I hope to track down any subpixel offset issues. Although, my track record for this, with other methods (`hansenlaw`) is not good ;-)

DanHickstein:
Sounds exciting! Changing to half-image transforms sounds like a good step, and a better understanding (and comments!) of the code should help.

DanHickstein:
So, it looks like the intensity scaling was fixed in #232. @stggh, should we close this ancient issue?

stggh:
Hi @DanHickstein and @MikhailRyazanov, happy New Year. I am travelling at the moment, domestic, in South Australia until Jan 11. 

Yes, it is fine to close this issue.

DanHickstein:
Great! Thanks for letting me know. 

I hope that you are having great fun on your travels!

PR #232, opened 2018-10-03
Dasch cached memory basis

stggh:
This PR adds memory caching to the `dasch` methods (`onion_peeling`, `two_point`, `three_point`) and `linbasex`, in a similar manner to @MikhailRyazanov's `basex` work.  **Importantly**, `abel.tools.basis.get_bs_cached()` has been dropped, with `abel/tools/basis.py` removed, returning `get_bs_cached()` to each PyAbel method file `dasch.py` and `linbasex.py`. See #231. The differences in defining the basis files in a unique way meant that a common `get_bs_cached()` was useless.

The function `abel.{dasch/linbasex}.basis_cleanup()` from @MikhailRyazanov `basex` work has been added to delete the global basis variables.

`**kwargs` was dropped from `linbasex`, which exposed two old parameter name faults in the unit test and `linbasex` examples.

Testing. Perhaps, these could be unit tests, I am just worried about the execution time.
[code gist](https://gist.github.com/stggh/2e5ef31a8a211328d0091ea87a47e46c)
```
Dribinski image quadrant (251, 251) ========================================

(0) Generate basis ------------------------------

A suitable basis for 'three_point' was not found. A new basis will be generated.

three_point basis saved for later use to './three_point_basis_251.npy
                                                             ... in 12.8 ms

(1) Cached basis ------------------------------

Using memory cached basis
                                                             ... in 0.7 ms

(2) Read basis from file ------------------------------

Loading three_point basis ./three_point_basis_251.npy
                                                             ... in 1.4 ms



Dribinski image quadrant (2501, 2501) ========================================

(0) Generate basis ------------------------------

A suitable basis for 'three_point' was not found. A new basis will be generated.

three_point basis saved for later use to './three_point_basis_2501.npy
                                                             ... in 1363.3 ms

(1) Cached basis ------------------------------

Using memory cached basis
                                                             ... in 321.9 ms

(2) Read basis from file ------------------------------

Loading three_point basis ./three_point_basis_2501.npy
                                                             ... in 374.0 ms


 [remove small size basis file three_point_basis_251.npy]

Back to (251, 251) image size ----------------------------------------

(3) Cached from larger basis cache ------------------------------

Loading three_point basis ./three_point_basis_2501.npy
                                                             ... in 23.7 ms

(4) Read from larger basis file ------------------------------

Loading three_point basis ./three_point_basis_2501.npy
                                                             ... in 20.8 ms


 [clean up, remove basis files three_point_basis_2501.npy]

```
```
Dribinski image quadrant (51, 51) ========================================

(0) Generate basis ------------------------------

A suitable basis for linbasex was not found.
A new basis will be generated.
linbasex basis saved for later use to ./linbasex_basis_101_02_050_1_0.npy
                                                             ... in 11.8 ms

(1) Cached basis ------------------------------

Using memory cached basis
                                                             ... in 6.1 ms

(2) Read basis from file ------------------------------

loading ./linbasex_basis_101_02_050_1_0.npy ...
                                                             ... in 6.5 ms



Dribinski image quadrant (251, 251) ========================================

(0) Generate basis ------------------------------

A suitable basis for linbasex was not found.
A new basis will be generated.
linbasex basis saved for later use to ./linbasex_basis_501_02_050_1_0.npy
                                                             ... in 143.7 ms

(1) Cached basis ------------------------------

Using memory cached basis
                                                             ... in 128.9 ms

(2) Read basis from file ------------------------------

loading ./linbasex_basis_501_02_050_1_0.npy ...
                                                             ... in 133.7 ms
```

Happy to make any changes.


DanHickstein:
> Importantly, abel.tools.basis.get_bs_cached() has been dropped, with abel/tools/basis.py removed, returning get_bs_cached() to each PyAbel method file dasch.py and linbasex.py. See #231. The differences in defining the basis files in a unique way meant that a common get_bs_cached() was useless.

I think that this is the right move. It was a fine idea to try to unify the code, but I think that it turned out to be more complicated that way.

The code looks good to me!

stggh:
rebased with PR #228 (I hope)

DanHickstein:
I think that we can go ahead and merge this, right?

stggh:
**Summary**
1) Dasch methods (`onion_peeling`, `three_point`, `two_point`)  caching of  the "basis" (deconvolution array)

    Mirrors the work of @MikhailRyazanov for the `basex` method. A previously calculated deconvolution array is saved, and used for any subsequent `dasch` method calculation, if the image size has the same or smaller width (number of columns). Otherwise, as usual, the basis file is read, if available. A larger basis array is sliced for use with smaller images.

    Note, unlike `basex` caching does not result in a massive time saving, since reading the small basis from a `.npy` file is also pretty efficient.  Reading and slicing from a much larger basis file may actually be slower than a small column size calculation.

2) Basis calculation function `get_bs_cached()` returned to `dasch,py`

    Due to substantial differences between `dasch`, `basex`, and `linbasex` basis file labelling,  a common `abel/tools/basis/get_bs_cached()` did not work. This basis generation function has been returned to each respective method file.  For `basex` it never left. `abel/tools/basis.py` has been removed.

3) `linbasex` caching, intensity, alignment

   Now also has a cached basis and local `get_bs_cached()` function.

   A factor of `1/2r` has been included so that the transform intensity matches the other `PyAbel` transforms, for real photoelectron spectra. `linbasex` transforms only 2D-square arrays, and so could not be calibrated using the 1D transform pairs.

   There are some issues with subpixel alignment that are being addressed, see #224, and it seem possible that `linbasex` should be able to transform 1/2-images, as has been achieved for `basex` #230 (?), making it consistent with the other `PyAbel` methods, and saving substantial transform time.

MikhailRyazanov:
[`cache_cleanup()`](https://github.com/PyAbel/PyAbel/blob/f49e2212e3ced614dbe84a66c542f0aebf2ecf59/abel/dasch.py#L389-L393) should include `_source` in `global`. Or omit `_source = None`, if cleaning it is not needed.

stggh:
Good point. Please add `_source` to the `global` list for your upcoming `basex` PR

MikhailRyazanov:
OK!

stggh:
> it seem possible that `linbasex` should be able to transform 1/2-images

I was a too optimistic with this statement. It can be done, but there is little to gain, as the size of the basis calculation is unchanged, and the part input-image sizes must be padded to perform the projection.

The issue is that the `linbasex` method sums along each row, to form a projection <math>L<sub>iu</sub> (right side intensity profile of this figure):
image
This profile is least-squares fitted to the sum of Legendre polynomial products, providing coefficients <math>p<sub>lk</sub></math>
image

Hence, the size of the basis array is proportional to the number of rows, which is unchanged for a `1/2`-image size. The calculation of the projection sum for each projection angle using `scipy.ndimage.interpolation.rotate()` also requires padding, for part images, in order to centre the image for the rotation. 

Code for the above in my [linbasex-half-image](https://github.com/stggh/PyAbel/tree/linbasex-half-image) fork.

So, nothing to gain from the original approach of duplicating the part image to form a full image for `linbasex` processing.  Although, I see `linbasex_transform()` assumes a quadrant image, so a little tidying of the original code is required. I will do this when I update the docs.

PR #233, opened 2018-10-26
Use better timer for benchmarks on Windows

rth:
`time.time` has only a  ~16 ms precision on Windows and `time.clock` is more appropriate there.

This switches to using `timeit.default_timer` which takes the best available timer automatically.

### DIFF
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -43,7 +43,7 @@ class AbelTiming(object):
             maximum n run for the "slow" transform methods, so far including 
             only the "direct_python" implementation.
         """
-        from timeit import Timer
+        from timeit import Timer, default_timer
         import time
 
         self.n = n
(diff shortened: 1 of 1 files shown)


MikhailRyazanov:
The functions are always executed only once, which still produces a large run-to-run scatter for faster methods. Is there a low-overhead way to do something like "loop for at least 10 seconds"? This would be also very useful for transforms benchmarks, since they have quite different speeds (>100-fold), and using a single `transform_repeat` counter for all of them is problematic.

stggh:
I am fine with this change. However, the `benchmark` design requires a rethink. The basis calculations use `time.default()` measured once, while the PyAbel methods are timed with `timeit`, over 3 measurements (for small image sizes). The biggest issue I have is that by using a random data image input, the transforms cannot/are not verified. 

> like "loop for at least 10 seconds"?

  I am in favour of this concept.

MikhailRyazanov:
Actually, a [straightforward Python loop](https://gist.github.com/MikhailRyazanov/2687308a83c75438012b5e949ee53296) has negligible overhead for millisecond execution timescales.
(In the output, the i5 times increase towards the end because it is a laptop, and the CPU throttles down as it heats up.)

Perhaps that `time_loop_t(duration)` example can be modified to run at least _N_ times and optionally skip _M_ first iterations.

DanHickstein:
Okay, shall we merge this and open a new issue for potentially using a loop for the benchmarks in order to provide more control over the number of iterations and which are ignored?

MikhailRyazanov:
@DanHickstein, unless @rth wants to implement these loops now, merging this would definitely help.

DanHickstein:
This is good to go. Do you want to hit merge this time? :)

MikhailRyazanov:
Thanks, @rth!

Issue #226, opened 2018-09-05
Consider caching basis sets as variables to avoid reading from the hard drive


DanHickstein:
In https://github.com/PyAbel/PyAbel/issues/225#issuecomment-417806263, @MikhailRyazanov pointed out that we are (perhaps foolishly) loading the basis sets from disk for every trasform:

>Regarding caching, it seems that you load the matrices from disk every time. Have you considered keeping them in memory (as globals in abel.basex?) after the first load?

He is correct: we always load the basis set from disk, and could potentially keep them in RAM after the first load, which could make repeated transforms faster. In principle, it would be better to store the basis sets as variables in RAM rather than load them from the hard-drive (unless necessary). However, it's probably not worth the effort to change this until we can identify a situation where avoiding the hard-drive load would provide a considerable speed-up to the transform methods.

I checked for basex, and it seems that the loading of the basis sets takes less time than the transform (but the same order-of-magnitude). I get a similar result for the Dasch methods, with the loading of the basis set taking about 4 times less time than the actual transform. 

I realize that the major caveat here is that I'm using a laptop with an SSD, so probably have a high ratio of drive load speed to CPU power. @stggh, are your i7-6700 benchmarks done by saving basis sets to a solid-state drive? If not, then we may be hard-drive access/read-rate limited for some methods in those benchmarks.

stggh:
My Dell PC uses an SSD. 

All the basis methods (except `basex`) use `abel.tools.basis.get_bs_cached(method, cols, basis_dir=basis_dir)` to get the basis, and then call the transform routine.

e.g. `dash.py`
```python
def dasch_transform(IM, D):
    return np.tensordot(IM, D, axes=(1, 1))
```

In principle, I now believe that there could be significant gains.

MikhailRyazanov:
I think, the "disk" speed should not really matter, since the OS itself will keep the file cached in RAM. The only real overhead is to parse these pickled structures into actual objects, plus memory management, but all this is indeed cheaper than the transforms. In any case, if you strive for the highest throughput, this can give some speed-up.

But my suggestion was mostly related to the BASEX regularization issue (#225). Saving all these matrices for different regularization parameters would require a 2-level disk caching scheme, as @DanHickstein said, and will clutter the disk unnecessarily — the `vert_left`, `horz_right` calculations can be done relatively quickly (compared to the bases calculations) to be tolerable for single runs, but repeating them every time in batch processing makes no sense.

[...]

PR #235, opened 2018-11-02
BASEX improvements

MikhailRyazanov:
This addresses all issues in #230 and finalizes #226.

## Main changes

The basis-generation code (`_bs_basex()`) is rewritten almost completely.
- Now it works with arbitrary width parameters _σ_ and accepts `n, sigma` instead of `n, nbf`. `sigma` is actually used everywhere, since it is more fundamental than `nbf` (which is `n` / `sigma`).
- The computations became is >10 times faster:
   - All the needed Γ-function values are precomputed, which takes _O_(_n_<sup>2</sup>) extra space (~the same as for the basis matrices), but avoids recomputing them _O_(_n_<sup>3</sup>) times.
   - The summation cutoff is tightened very close to the limit for "numerically exact" results (floating-point precision).
- It can take previously computed smaller basis matrices for the same _σ_ and extend them by computing only the missing (outer) elements.

The `get_bs_cached()` code (renamed from `get_bs_basex_cached()` for consistency) is changed accordingly.
- The file-naming scheme is changed to `basex_basis_{n}_{sigma}.npy`.
- Attempts to get the basis (if not already in memory) are tried in the following order:
   1. If the file for given `n`, `sigma` exists, then it is used.
   2. If files for larger `n` and the same `sigma` exist, then the smallest of them is loaded and cropped.
   3. If files for smaller `n` and the same `sigma` exist, then the largest of them is loaded and extended by `_bs_basex()`.
   4. If nothing helps, then a completely new basis is generated.

The `MAGIC_NUMBER` intensity correction is replaced by `get_basex_correction()`, activated by passing `correction=True` to `basex_transform()` (or `get_bs_cached()`). See #230 for details.

A new module `abel.tools.polynomial` is added (primarily for `get_basex_correction()`), which implements the analytical Abel transform for arbitrary polynomials. `Polynomial` and `PiecewisePolynomial` are added to `abel.tools.analytical`.

Test are added/updated for `abel.basex` and `abel.tools.polynomial`.

## Additional changes

Declared `_source` as `global` in `dash.cache_cleanup()` (amending #232).

Minor corrections to documentation.
In particular, `abel.tools.analytical` docstrings are moved from `__init__()` to class levels, since they were not visible in the generated documentation.

## Questions

What is `mask_valid` in `BaseAnalytical` and how to set it meaningfully for polynomials?

What happened to the `GaussianAnalytical` docstring (the [`ratio_valid_sigma` description](https://github.com/PyAbel/PyAbel/blob/be2d1adb1b20a664ac6f757d1c40fb737222ba63/abel/tools/analytical.py#L187-L190))?

The are some problems with multiple parameters in docstrings, for example, see [`abel.tools.polar.cart2polar`](https://pyabel.readthedocs.io/en/latest/abel.html#abel.tools.polar.cart2polar) — `x, y : floats or arrays` is rendered as "**y** (_x,_)".
This is probably related to https://github.com/sphinx-doc/sphinx/issues/2227?

Perhaps, `linbasex.basis_cache_cleanup()` should be renamed to `linbasex.cache_cleanup()` for consistency with and `dasch.cache_cleanup()` `basex.cache_cleanup()`?

### DIFF
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -71,7 +71,7 @@ def basex_transform(data, nbf='auto', reg=0.0, bs_correction=False,
                     verbose=True, direction='inverse'):
     """
     This function performs the BASEX (BAsis Set EXpansion)
-    Abel Transform. It works on a "right side" image. I.e.,
+    Abel transform. It works on a "right side" image. I.e.,
     it works on just half of a cylindrically symmetric
     object, and ``data[0,0]`` should correspond to a central pixel.
     To perform a BASEX transorm on
@@ -384,6 +384,30 @@ def get_bs_basex_cached(n, nbf='auto', reg=0.0, bs_correction=False,
         _Ai = Ai
 
     return Ai
+
+
+def basex_cleanup():
+    """
+    Utility function.
+
+    Frees the memory caches created by ``get_bs_basex_cached()``.
+    This is usually pointless, but might be required after working
+    with very large images, if more RAM is needed for further tasks.
+
+    Parameters
+    ----------
+    None
+
+    Returns
+    -------
+    None
+    """
+    global _prm, _M, _reg, _Ai
+
+    _prm = None
+    _M = None
+    _reg = None
+    _Ai = None
 
 
 MAX_BASIS_SET_OFFSET = 4000
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -31,6 +31,8 @@ from ._version import __version__
 # https://doi.org/10.1063/1.1482156
 #
 #
+# 2018-10-27
+#   MR exposed BASIS_SET_CUTOFF for speed/accuracy control.
 # 2018-10-07
 #   MR added intensity correction.
 #   Also smaller basis sets are now reused when generating larger ones.
@@ -462,6 +464,24 @@ def get_basex_correction(Ai, sigma):
     return cor
 
 
+# The analytical expresion for the k-th basis-function projection
+# involves a sum of k^2 terms, most of which are very small.
+# Setting BASIS_SET_CUTOFF = c truncates this sum to ±ck terms
+# around the maximum.
+# The computation time is roughly proportional to this parameter,
+# while the accuracy (at least for n, k < 10000) is as follows:
+#   cutoff   relative error
+#     4        < 2e-4
+#     5        < 2e-6
+#     6        < 7e-9
+#     7        < 1e-11
+#     8        < 6e-15
+#     9        < 1e-15
+# The last one reaches the 64-bit floating-point precision,
+# so going beyond that is useless.
+# See https://github.com/PyAbel/PyAbel/issues/230
+BASIS_SET_CUTOFF = 6  # highest speed with reasonable accuracy
+
 def _bs_basex(n=251, sigma=1.0, oldM=None, verbose=True):
     """
     Generates horizontal basis sets for the BASEX method.
@@ -551,7 +571,7 @@ def _bs_basex(n=251, sigma=1.0, oldM=None, verbose=True):
             # index of the largest component
             lmax = min(int(u2), k2)
             # halfwidth of the important range
-            delta = 7 * k  # "7" reproduces the old method accuracy (~1e-8 abs)
+            delta = int(BASIS_SET_CUTOFF * k)
             # summation limits: +-delta from the maximum, but within [0, k^2]
             minl = max(0, lmax - delta)
             maxl = min(lmax + delta, k2)
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -16,8 +16,6 @@ import sys
 import numpy as np
 from scipy.special import gammaln
 from scipy.linalg import inv
-from scipy.ndimage import median_filter, gaussian_filter, center_of_mass
-import scipy
 
 from ._version import __version__
 
@@ -31,6 +29,8 @@ from ._version import __version__
 #
 # 2018-10-07
 #   MR added intensity correction.
+#   Also smaller basis sets are now reused when generating larger ones.
+#   Removed unnecessary scipy methods (scipy.dot is actually numpy.dot).
 # 2018-10-03
 #   MR completely rewrote basis generation (half-width, efficiency).
 #   Switched from (n, nbf) to (n, sigma) basis specification;
@@ -192,7 +192,7 @@ def basex_core_transform(rawdata, Ai, dr=1.0):
     # its overall effect is an identity transform.
 
     # Reconstructing image  - This is where the magic happens
-    IM = scipy.dot(rawdata, Ai) / dr
+    IM = rawdata.dot(Ai) / dr
     # P = dot(dot(Mc,Ci),M.T) # This calculates the projection, !! not
     # which should recreate the original image                  !! really
     return IM
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -296,23 +296,29 @@ def get_bs_basex_cached(n, sigma=1.0, reg=0.0, correction=False,
                 best_file = path_to_basis_file
             else:
                 # Find the best (smallest among sufficient)
+                # and the largest (to extend if not sufficient)
                 best_file = None
                 best_n = sys.maxint
+                largest_file = None
+                largest_n = 0
+                mask = re.compile(r'basex_basis_(\d+)_{}.npy$'.format(sigma))
                 for f in listdir(basis_dir):
                     # filter BASEX basis files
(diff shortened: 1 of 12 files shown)


doc/abel.rst
```diff
@@ -169,3 +177,13 @@ abel.benchmark module
     :members:
     :undoc-members:
     :show-inheritance:
+
+abel.tests module
```
MikhailRyazanov: I think, @rth has removed this in 5636c402c3791424859335f8d78ddb41ac47e5e9. I don't know how it got back. Need to remove again?

MikhailRyazanov:
_Who will test the testers?_

I am wondering why AppVeyor tests with Python 3.5 fail.

Most interestingly, the 32- and 64-bit builds fail at different checks in `test_basex_basis_sets_resize()`, all of which pass without problems in Travis CI and on both my computers (i5 GNU/Linux and Ryzen3 Win10). And they could not fail the way the do — producing different numbers in the arrays — since the only difference is which of these numbers are calculated and which are copied from the previous calculations with the same parameters.

It seems that their machines have some memory corruption or slightly broken floating-point units. Is it possible to rerun these tests? Or if somebody can reproduce the problem...

stggh:
The default, `correction=False`. Oh, you are right ~1.5%, I was looking at profile 2, but its maximum is at the origin. 1.5% correct then?

MikhailRyazanov:
It should be (2/√π) / `MAGIC_NUMBER` ≈ 1.0145, that is, ~1.45%.

Try `correction=True`. It should give proper normalization and remove the glitch near _r_ = 0.

Also using `sigma` ≠ 1 (with default `correction=False`) should produce oscillating but averagely correct intensities. Then `correction=True` should also remove these oscillations.

stggh:
Yes, that's it. Perhaps, `correction=True` should be the default?

MikhailRyazanov:
I also think so, but wanted to make sure that the fundamental stuff works properly and then that this new correction does not cause problems.

DanHickstein:
This is fantastic! Well done @MikhailRyazanov. 

> I have changed the assert np.array_equal() test to assert_allclose() with 10−15 tolerances to avoid this problem.

Yeah, `allclose()` is way nicer than checking for perfect equality. It's amazing how often Python thinks that two arrays that should be equal aren't actually equal :)

> What is mask_valid in BaseAnalytical and how to set it meaningfully for polynomials?

I think that @rth could give you the best answer to this question. But, I think that it's an array that is the same length as the returned array and serves as a True/False mask for where the transform is valid. I think that some functions only had a certain range where they were correct.

> What happened to the GaussianAnalytical docstring (the ratio_valid_sigma description)?

That docstring appears to be messed up. Maybe open an Issue?

> The are some problems with multiple parameters in docstrings, for example, see abel.tools.polar.cart2polar — x, y : floats or arrays is rendered as "y (x,)".
This is probably related to sphinx-doc/sphinx#2227?

Maybe we can the `napoleon_use_param = False` in `conf.py` as suggested. You can try to do that in this PR, or open a separate issue.

> Perhaps, linbasex.basis_cache_cleanup() should be renamed to linbasex.cache_cleanup() for consistency with and dasch.cache_cleanup() basex.cache_cleanup()?

That makes sense to me.

abel/basex.py
```diff
@@ -129,8 +153,8 @@ def basex_transform(data, nbf='auto', reg=0.0, basis_dir='./', dr=1.0,
     n = data.shape[1]
 
     # load the basis sets:
-    Ai = get_bs_basex_cached(n, nbf=nbf, reg=reg,
-                             basis_dir=basis_dir, verbose=verbose)
+    Ai = get_bs_cached(n, sigma=sigma, reg=reg, correction=correction,
```
stggh: function name change also within `benchmark.py` line 53: `'basex_bs': basex.get_bs_basex_cached

doc/abel.rst
rth: Yes, there is no point in including tests as part of the public API in the docs.

MikhailRyazanov: There is also `test_suite="abel.tests.run_cli"` remaining [in setup.py](https://github.com/PyAbel/PyAbel/blob/be2d1adb1b20a664ac6f757d1c40fb737222ba63/setup.py#L116). Perhaps, should be removed or updated?

abel/basex.py
MikhailRyazanov: Thanks!

MikhailRyazanov:
@DanHickstein, `napoleon_use_param = False` seems to work.

@stggh, thanks for noticing the remaining `get_bs_basex_cached()`!
Do you agree with renaming `linbasex.basis_cache_cleanup()` → `linbasex.cache_cleanup()`?

@rth, it seems that `GaussianAnalytical()` had that corrupted docstring [when you first committed it](https://github.com/PyAbel/PyAbel/blob/5da3cd9c77323be017d233a3f1fa82b474ae9f85/abel/analytical.py#L105-L107). Do you know what it was supposed to be?

stggh:
> renaming linbasex.basis_cache_cleanup() → linbasex.cache_cleanup()

Yes, feel free to change it in this PR.

MikhailRyazanov:
In principle, this could be merged as is, but I would like to hear from @rth (or anybody else) about the `mask_valid` (I can set it to `False` at each end of each interval, if this will help anything) and about
https://github.com/PyAbel/PyAbel/pull/235#discussion_r230621726 and the `GaussianAnalytical` docstring.

Also, what everybody thinks about using `correction=True` by default?

DanHickstein:
Yes, if there is any region of the analytic transform where the results are not correct, then set those positions of `mask_valid` to `False`. 

I would imaging that the majority of users would want `correction=True`, so I think that this is an appropriate default. By default we should offer whatever decisions we think produce the best results.

rth:
TBH, I don't have much recollection about anything related to `mask_valid`, and looking at the half written docstring I have no idea now what was missing :) If anything doesn't make sense, you might as well re-write it, sorry..

MikhailRyazanov:
@rth, something like this?
```
    ratio_valid_sigma : float
        in the benchmark take only the range
        0 < r < ration_valid_sigma * sigma
        (exclude possible artefacts on the axis and the possibly clipped tail)
```

And what about `test_suite="abel.tests.run_cli"` in setup.py?

rth:
Yes, that makes sense.

> And what about test_suite="abel.tests.run_cli" in setup.py?

Let's remove it, I'm pretty sure I saw somewhere that specifying `test_suite` in `setup.py` was a historical approach, not very recommended now. It's definitely not useful with pytest.

MikhailRyazanov:
@DanHickstein, OK, I'll then set the default `correction=True`.

The mask stuff is not very clear.

One the one hand, if the resulting function is continuous (is there any reason to use discontinuous piecewise polynomials in tests?), then it should be well-behaved everywhere.

One the other hand, if the function has singularities, then the extent of the problematic regions will depend on the transform method more than on the function itself.

Maybe adding a parameter like `valid_gap` to set ±`valid_gap` points around each domain boundary to `False` would be reasonable?

doc/abel.rst
MikhailRyazanov: OK, removed this and `test_suite="abel.tests.run_cli"`.

MikhailRyazanov:
Final changes
---

BASEX:
* Enabled `correction=True` by default.

Other:
* Renamed `linbasex.basis_cache_cleanup()` to `linbasex.cache_cleanup()` for consistency with other methods (see #232).
* Set `napoleon_use_param = False` in `doc/conf.py` for proper rendering of multiple input parameters with shared type/description. Please check that it did not break something else.
* Corrected the broken `GaussianAnalytical` docstring.
* Removed `test_suite="abel.tests.run_cli"` from `setup.py` (amending #228).

stggh:
> which functions had invalid regions

`direct`
   See #214 for issues with the grid/derivative that produced odd end values, now fixed.

`hansenlaw`
   End values are not defined, copied from adjacent columns.

MikhailRyazanov:
@stggh, so each method should also have its own `mask`.

Let's leave it as is for now. If some useful thoughts on where and how this can be used appear, then it would be better to make a separate decision, consistent for all functions and methods.

DanHickstein:
Is this ready to merge?

MikhailRyazanov:
Should be...

stggh:
@MikhailRyazanov as the code author, you have the honour of pushing the green button.

MikhailRyazanov:
Done.
Thanks everybody for the feedback and testing!

DanHickstein:
Awesome! PyAbel has seen some great improvements in the past few weeks :)

PR #240, opened 2018-12-05
BASEX forward transform

MikhailRyazanov:
This adds the forward Abel transform to the BASEX method.

The basic idea is that since the projected image and its reconstruction are described by the same coefficients, but in different basis sets (related to each other through the forward/inverse Abel transform), swapping these basis sets and applying the same method reverses the transform direction.

Then there are some technical details related to the _dr_ scaling and what to use for the intensity correction.

`cache_cleanup()` now accepts an optional parameter for selective cache purging. For example, it is possible to do a forward transform, discard its matrix, keeping the basis set cached, and do an inverse transform. By default, it still purges everything.

I've added new tests for the forward transform, but no benchmarks, since they should be the same. Do we need them for completeness?

### DIFF
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -243,10 +243,10 @@ def _nbf(n, sigma):
 _bs_prm = None   # [n, sigma]
 _bs = None       # [M, Mc]
 # forward transform
-_trf_prm = None  # [reg, correction]
+_trf_prm = None  # [reg, correction, dr]
 _trf = None      # Af
 # inverse transform
-_tri_prm = None  # [reg, correction]
+_tri_prm = None  # [reg, correction, dr]
 _tri = None      # Ai
 
 def get_bs_cached(n, sigma=1.0, reg=0.0, correction=True,
@@ -387,9 +387,9 @@ def get_bs_cached(n, sigma=1.0, reg=0.0, correction=True,
 
     # Check whether transform matrices for these parameters
     # are already created
-    if direction == 'forward' and _trf_prm == [reg, correction]:
+    if direction == 'forward' and _trf_prm == [reg, correction, dr]:
         A = _trf
-    elif direction == 'inverse' and _tri_prm == [reg, correction]:
+    elif direction == 'inverse' and _tri_prm == [reg, correction, dr]:
         A = _tri
     else:  # recalculate
         if verbose:
@@ -401,18 +401,17 @@ def get_bs_cached(n, sigma=1.0, reg=0.0, correction=True,
             cor = get_basex_correction(A, sigma, direction)
             A = np.multiply(A, cor)
         if direction == 'forward':
-            _trf_prm = [reg, correction]
+            _trf_prm = [reg, correction, dr]
             _trf = A
         else:  # 'inverse'
-            _tri_prm = [reg, correction]
+            _tri_prm = [reg, correction, dr]
             _tri = A
-
-    # Apply intensity scaling, if needed
-    if dr != 1.0:
-        if direction == 'forward':
-            A *= dr
-        else:  # 'inverse'
-            A /= dr
+        # apply intensity scaling, if needed
+        if dr != 1.0:
+            if direction == 'forward':
+                A *= dr
+            else:  # 'inverse'
+                A /= dr
 
     return A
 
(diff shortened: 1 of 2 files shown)


stggh:
This is beautiful work @MikhailRyazanov! 

There may be some issue with normalization. In this comparison `basex` is scaled:
transformpairs-forward


n  |  ratio
----|-------
101 |  10<sup>-4</sup>
1001 | 10<sup>-6</sup>

which appears to be `1/dr`<sup>2</sup>

```python
import numpy as np
import abel

n = 101

fig, axs = plt.subplots(4, 2, sharex=True, sharey=True)

for prof in range(1, 9):
    if prof != 4:
        f = abel.tools.analytical.TransformPair(n=n, profile=prof)

        hl = abel.hansenlaw.hansenlaw_transform(f.func, dr=f.dr, 
                                                direction='forward')
        dr = abel.direct.direct_transform(f.func, dr=f.dr, direction='forward')
        ba = abel.basex.basex_transform(f.func, dr=f.dr, direction='forward',
                                        correction=True, verbose=False)

    row = (prof - 1) // 2
    col = (prof - 1) % 2
    ax = axs[row, col]

    if prof != 4:
        ax.plot(f.r, f.abel)
        ax.plot(f.r, hl)
        ax.plot(f.r, dr)
        ratio = f.abel.max()/ba.max()
        print(ratio, f.dr)
        ax.plot(f.r, ba*ratio)
        ax.annotate(f.label, (0.75, 1.6), fontsize='smaller')
        msehl = np.square(hl-f.abel).mean()
        msedr = np.square(dr-f.abel).mean()
        mseba = np.square(ba*ratio-f.abel).mean()
        ax.annotate('mse', (0.8, 1.1), fontsize=9)
        ax.annotate('{:4.2e}'.format(msehl), (0.75, 0.85), fontsize=7,
                    color='C1')
        ax.annotate('{:4.2e}'.format(msedr), (0.75, 0.6), fontsize=7,
                    color='C2')
        ax.annotate('{:4.2e}'.format(mseba), (0.75, 0.35), fontsize=7,
                    color='C3')

        if col == 0:
            ax.set_ylabel(r'projection')
        if row == 3:
            ax.set_xlabel('radius')
    else:
        # dummy lines for a legend block
        ax.plot((10, 10.01), (5, 5.01), label='projection')
        ax.plot((10, 10.01), (5, 5.01), label=r'hansenlaw')
        ax.plot((10, 10.01), (5, 5.01), label=r'direct')
        ax.plot((10, 10.01), (5, 5.01), label=r'basex')
        ax.legend(fontsize='smaller', labelspacing=0.1, loc=10)
        ax.axis(xmin=0, xmax=1, ymin=0, ymax=2)

plt.subplots_adjust(wspace=0.1, hspace=0.1)
plt.suptitle(f'forward Abel transforms n={n}')
plt.savefig('TransformPairs-forward.png', dpi=75)
plt.show()
```

stggh:
Doc changes to `doc/transform_methods/basex.rst `, which has
```
Note that the forward BASEX transform is not yet implemented in PyAbel.
```

DanHickstein:
This looks great!

Once the scaling issue raised by @stggh is addressed, I'd say that this can be merged.

At some point, it would be nice to include the forward transform in the benchmarks for completeness. But, we can do this in a separate PR.

MikhailRyazanov:
@stggh, huge thanks for testing!

> which appears to be `1/dr`<sup>2</sup>

Aaaaa!
It actually is not. The first time the normalization was correct, but every call to `get_bs_cached()` divided/multiplied it by `dr`. This should have made the inverse transform becoming wrongly scaled as well.

I was testing with single transforms and thus did not notice that I've coded this incorrectly. We should probably add repetitive transforms in the unit tests.

The updated code should work correctly.
(I hope, in real-world applications, `dr` would not change from frame to frame, so caching it that way should not harm the performance.)

stggh:
All good. `basex` has the lowest mse. Feel free to merge.

MikhailRyazanov:
> At some point, it would be nice to include the forward transform in the benchmarks for completeness. But, we can do this in a separate PR.

Yeah, the benchmarks need to be updated anyway (for example, "basex(var)" is still not there), so it can be added then.

PR #241, opened 2018-12-06
Documentation updates (BASEX, polynomials, formatting issues)

MikhailRyazanov:
The documentation updated to reflect my previous code additions and changes (#227, #235, #240).

The two main changes are:
1. The updates/additions for the BASEX method (changes in its description `doc/transform_methods/basex.rst` and addition of computational details: `.../basex-comp.rst` with figures in `.../basex-basis.py` and `.../basex-vert.py`).
2. The addition of the polynomials mathematical documentation (`doc/tools/polynomial.rst` with a figure in `...\smoothstep.py`).

There are also many minor changes to address Sphinx warnings about formatting and to improve the structure in several cases. On my computer `make html` now runs without warnings. Please check whether you see any problems in the building process or in the results.


MikhailRyazanov:
I also suggest using SVG as the default format for plots in HTML. At high-resolution screens it looks much better than the default PNG:

PNG: basex-png
SVG: basex-svg

PNG: hl-png
SVG: hl-svg

Here I have also disabled the automatic addition of the source/formats links. In Examples, we already have the sources included, and in many other cases these links are merely distracting (if needed, they can be added on per-image basis, I believe).

All this can be done by adding
```
plot_formats = ['svg', 'png']
plot_html_show_source_link = False
plot_html_show_formats = False
```
to `doc/conf.py`.

Do we need to care about any other formats besides HTML? (I haven't tested any.)

MikhailRyazanov:
@stggh,
> generates this error

WTH is `xravel.mask`???

Do you see any other errors?

What happens if you run `doc/tools/smoothstep.py` manually (with uncommented `show()` or `savefig()`)?

stggh:
> WTH is xravel.mask???

Good point!  Something to do with matplotlib.

> What happens if you run doc/tools/smoothstep.py manually 

Same error:
```python
[etalon:.+R-doc/doc/tools]$ python smoothstep.py
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/units.py", line 168, in get_converter
    if not np.all(xravel.mask):
AttributeError: 'numpy.ndarray' object has no attribute 'mask'

  File "smoothstep.py", line 51, in <module>
    plt.plot(r, map(smoothstep, r), color='red')
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/pyplot.py", line 2813, in plot
    is not None else {}), **kwargs)
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/__init__.py", line 1805, in inner
    return func(ax, *args, **kwargs)
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/axes/_axes.py", line 1603, in plot
    for line in self._get_lines(*args, **kwargs):
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/axes/_base.py", line 393, in _grab_next_args
    yield from self._plot_args(this, kwargs)
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/axes/_base.py", line 370, in _plot_args
    x, y = self._xy_from_xy(x, y)
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/axes/_base.py", line 205, in _xy_from_xy
    by = self.axes.yaxis.update_units(y)
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/axis.py", line 1467, in update_units
    converter = munits.registry.get_converter(data)
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/units.py", line 181, in get_converter
    converter = self.get_converter(next_item)
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/units.py", line 187, in get_converter
    thisx = safe_first_element(x)
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/matplotlib/cbook/__init__.py", line 1635, in safe_first_element
    raise RuntimeError("matplotlib does not support generators "
RuntimeError: matplotlib does not support generators as input
```

Passing a list to matplotlib.pyplot.plot() "fixes"  the problem:
```python
ss = [smoothstep(x) for x in r]
plt.plot(r, ss, color='red')
```

smoothstep

MikhailRyazanov:
OK, I've just found the same solution (the issue apparently is that `map()` in Python 3 returns an iterator, whereas in Python 2 it was a list). I'll update that file.

Any other problems spotted?

stggh:
No, looks good. Excellent work fixing all the doc syntax errors.

MikhailRyazanov:
Yes, I think `linbasex` docs can wait for other `linbasex` changes (IIRC, you had some).

If people have time to read what I have added in this PR, I would appreciate notifying me of any problems with clarity and/or grammar (hopefully, there are no mathematical errors), such that we can release this in a well readable form.

doc/tools/polynomial.rst
```diff
@@ -0,0 +1,205 @@
+:orphan:
```
DanHickstein: Where does this get built as part of the documentation? I built the docs, and don't see it anywhere...

doc/transform_methods/basex-comp.rst
```diff
@@ -0,0 +1,298 @@
+:orphan:
```
DanHickstein: This file also appears not to be built in the docs anywhere...

MikhailRyazanov:
“BASEX: computational details” (`basex-comp.rst`) is linked from “BASEX” (`basex.rst`); “Polynomials” (`polynomial.rst`) is linked from `abel.tools.polynomial` (`polynomial.py`) and `abel.tools.analytical.Polynomial` (`analytical.py`). They are not included in any TOC, since I did not find a good place and method to do so — they describe technical/implementation details that are not really essential to the users and, logically, are subtopics of the corresponding method/tools,  so adding them in the common TOC at the same level does not look appropriate. Is it possible to add them to the TOC at a deeper level after the corresponding topic?

doc/transform_methods/basex-comp.rst
```diff
@@ -0,0 +1,298 @@
...
+
+.. _BASEXcomp:
+
+BASEX: computational details
+============================
+
+The BASEX article does not provide the derivation of the basis projections and
+is very terse regarding their computation, so here we provide the missing
```
DanHickstein: Can we re-word this to make it a little kinder to the BASEX manuscript. Maybe something like, "Here we provide the full derivation of the basis projections that are summarized in Dribinski et al."

DanHickstein:
Yeah, we can add them at a deeper level. At the end of `basex.rst`, maybe we can include something like this:

```
Computational Details
---------------------

The links below provide more information on the computaional details of the BASEX method:

.. toctree::
   basex-comp
```

This includes basex-comp in the TOC. 

For the polynomials article, I suppose it's okay for it to hide where it's currently hiding. Alternatively, since it's a really nice article, we could just stick in in the main TOC by changing index.rst to

```
.. toctree::
   :maxdepth: 2

   readme_link
   abel
   transform_methods
   anisotropy_parameter
   circularize_image
   tools/polynomial
   examples
   contributing_link
```

(Note, I also moved "contributing_link" to the bottom, since I think that this should come last in the TOC.

MikhailRyazanov:
Now I've read the `toctree` documentation and tried to add
```
.. toctree::
    :hidden:

    basex-comp
```
at the end of “PyAbel improvements”. Looks like this:
toc
This makes it included in the TOC at a deeper level, more or less in an appropriate place, without polluting the text. And sphinx does not complain about orphans any more.

The same thing with `abel.tools.polynomial` works even better (“Polynomials” appears as a subtopic of “abel.tools.polynomial” in the TOC).

doc/transform_methods/basex-comp.rst
MikhailRyazanov: They deserve it! :–)
What about “To complement the general description given in the BASEX article, here we
provide the full derivation of the basis projections and the details needed for
their efficient computation.”?

DanHickstein: > They deserve it! :–)

Ha!

> What about “To complement the general description given in the BASEX article, here we
provide the full derivation of the basis projections and the details needed for
their efficient computation.”?

Sounds good :)

DanHickstein:
Oh, nice work! I like your new approach.

MikhailRyazanov:
While we are still in “documentation updates”, what people think about using normal plots for the transform pairs?

pairs

If “yes”, then I need a technical advice.

I think that creating individual script files for each of the plots is silly, so I have a single [script](https://gist.github.com/MikhailRyazanov/6de6e8affab18b08f54bcb91c0d2d5c4) parameterized with the profile number. The problem is how to call it from the docstring.

In principle, the `plot` directive [allows](https://matplotlib.org/devel/plot_directive.html) calling specific functions:
```
.. plot:: path/to/plot.py plot_function1
```
so I could define `def plot1(): plot(1)` and so on. However, all these figures are _saved in the same file_, named like the script file, regardless of the function name, so producing many figures from one script does not work this way.

Using
```
.. plot::

    import imp
    imp.load_source('', 'tools/transform_pairs.py').plot(1)
```
works, but does not look like the cleanest way. And “The `imp` package is pending deprecation in favor of `importlib`” in Python 3, while in Python 2 `importlib` does not really exist. Although `imp` still works in both versions, so maybe it is fine for now and later we can do something about it (for example, if we switch to Python 3 only)?

Another questions about the pairs:
1. As far as I can tell, `profile8` is the same as `profile6`, just written slightly differently. Do we need them both? If yes, maybe this needs to be mentioned in the documentation?
2. Why do we have `profile4` broken? Wouldn't it make sense to correct it and mention that the published expression was incorrect?

stggh:
> Why do we have profile4 broken? Wouldn't it make sense to correct it and mention that the published expression was incorrect?

We were unable to deduce the correct coefficients.

> As far as I can tell, profile8 is the same as profile6, just written slightly differently.

Probably, my bad for not recognising this. Only one function is required.

> normal plots for the transform pairs?

Provided it does not mangle the docstring <tab> help.

MikhailRyazanov:
> Provided it does not mangle the docstring help.

The docstrings would look like
```
    """**profile1**:
    `Cremers and Birkebak App. Opt. 5, 1057-1064 (1966) Eq(13)
    <https://doi.org/10.1364/AO.5.001057>`_

    .. math::

        \epsilon(r) &= 0.75 + 12r^2 - 32r^3  & 0 \le r \le 0.25

        \epsilon(r) &= \\frac{16}{27}(1 + 6r - 15r^2 + 8r^3)
                    & 0.25 \lt r \le 1

        I(r) &= \\frac{1}{108}(128a_1 +a_{0.25}) + \\frac{2}{27}r^2
                  (283a_{0.25} - 112a_1) +

        & \,\,\,\, \\frac{8}{9}r^2\left[4(1+r^2)\ln\\frac{1+a_1}{r} -
          (4+31r^2)\ln\\frac{0.25+a_{0.25}}{r}\\right] &  0 \le r \le 0.25

        I(r) &= \\frac{32}{27}\left[a_1 - 7a_1 r + 3r^2(1+r^2)
                \ln\\frac{1+a_1}{r}\\right]  & 0.25 \lt r \le 1

    ..
              source                projection
        ┼+1.3                  ┼+1.3               
        │                      o   o               
        │     x                │     o             
        │   x   x              │       o           
        │ x                    │                   
        x         x            │         o         
        │           x          │                   
        │                      │           o       
        │             x        │                   
        │                      │             o     
        ┼+0─────────────x──┼   ┼+0─────────────o──┼
        0          r      +1   0          r      +1

    .. plot::

        import imp
        imp.load_source('', 'tools/transform_pairs.py').plot(1)
    """
```
It's basically turning the ASCII-plot “code block” into a comment and appending the `plot` directive.

>>   As far as I can tell, profile8 is the same as profile6, just written slightly differently.
>
> Probably, my bad for not recognising this. Only one function is required.

So should we remove `profile8` completely?

Regarding `profile4`, it is a piecewise polynomial, for which I can now get everything...

stggh:
Sounds good. Yes drop profile8. I look forward to seeing the real profile 4. 

(I  will be quite for the next 2 weeks, on vacation.)

MikhailRyazanov:
Well, seeing it is as simple as
``` python
import abel.tools.polynomial
n = 1001
poly = abel.tools.analytical.PiecewisePolynomial(n, 1, 
           [(0, 0.7, [0.1, 0, 5.51, -5.25]),
            (0.7, 1, [-40.74, 155.56, -188.89, 74.07])],
           symmetric=False)
plt.plot(poly.r, poly.abel)
plt.show()
```
(by the way, the function itself has a small discontinuity at _r_ = 0.7).

Writing the actual expression requires some manual work, which I'm going to do soon...

MikhailRyazanov:
It turned out that the errors in Table 1 were very simple: “193.30083” must be “196.30083” in both cases.

With this correction, the differences between `profile4.abel` and the polynomial `abel` above are ≲10<sup>−6</sup>, which is due truncated coefficients (they in principle have 1/3 terms). Numerical forward/inverse transforms also look fine.

MikhailRyazanov:
@DanHickstein, what do you think about [using SVG and removing automatic source/formats links](https://github.com/PyAbel/PyAbel/pull/241#issuecomment-444737025)?

DanHickstein:
Whoops, sorry for the delay in responding. I've been traveling too much the past few months and have been dropping the ball. 

If you'd like to switch some of the image to SVG, that's fine with me. My screen isn't big enough so that the PNG looks bad, and I think that there is less chance that something could get corrupted when the PNG is displayed (in theory, different browsers/programs could display SVG images slightly differently). But, I don't have strong feelings. If you want SVG, go for it!

Also, should we merge this soon?

MikhailRyazanov:
Matplotlib apparently creates relatively simple SVGs, so they should work well in all modern browsers:
https://caniuse.com/svg
https://caniuse.com/svg-img
If something goes wrong, we can always return to PNGs.

> Also, should we merge this soon?

I was basically waiting for your decision. :–)
I'm going to add the SVG and transform-pairs plots here. If nothing breaks (well, the code has not changed...), then we should definitely merge soon.

And perhaps we need to update the copyright dates in `doc/conf.py`?

MikhailRyazanov:
Now Travis CI has broken dependencies: `pip install coveralls pytest pytest-cov` reports:
```
Requirement already satisfied: pytest in /home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages (3.3.0)
```
then installs `pytest-cov` and after that complains:
```
pytest-cov 2.6.1 has requirement pytest>=3.6, but you'll have pytest 3.3.0 which is incompatible.
```
(what was it thinking about earlier?!) So the tests themselves cannot even start, since `cd tmp/; pytest -v --cov=abel --pyargs abel` fails with
```
pluggy.PluginValidationError: Plugin 'pytest_cov' could not be loaded: (pytest 3.3.0 (/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages), Requirement.parse('pytest>=3.6'))!
```

Previously it was installing `pytest-cov 2.6.0`, which was working fine with `pytest 3.3.0`. Interestingly, their Python 3.7 environment has `pytest 3.10.0` preinstalled, so it also works fine now, but Python 3.6 and 2.7 fail due to this unhandled `pytest-cov`/`pytest` version conflict.

Any ideas what to do?

MikhailRyazanov:
> Can we just explicitly tell pip to install pytest >= 3.6?

As a last resort, I think. At least, searching the Internet shows that some people solve this problem by doing just that (or pinning `pytest-cov` to an older version).

However, to me this looks like a hack and is likely to break again very soon.

I do not use pip on my system (relying on the OS package manager, for a good reason, evidently) and do not know how it works, but what if we run that `pip install coveralls pytest pytest-cov` again? Whould it then upgrade `pytest` to satisfy the new dependencies? Or maybe something with `--upgrade` would help?

MikhailRyazanov:
OK, I have experimented a little bit on a Windows machine. Running `pip install ...` again _does not_ update anything (still saying "already satisfied" even though the versions are clearly wrong). However, adding the `--upgrade` option does update the already installed packages before installing the rest, so it indeed helps. (BTW, `pip install --upgrade pip setuptools wheel` already had `--upgrade`.) Hopefully, this will be more robust...

DanHickstein:
Awesome! Looks like this can be merged now, yeah?

MikhailRyazanov:
OK, merging...

Issue #242, opened 2019-01-11
Migrating to travis-ci.com


DanHickstein:
As mentioned in https://github.com/PyAbel/PyAbel/issues/84#issuecomment-453611888, we currently had readthedocs and Travis CI installed as "Github Services", which are being phased-out and will stop working at the end of this month (Jan 2019). I switched readthedocs to a webhook, but Travis CI is supposed to be installed as a Github App now.

This opened a small can of worms, because it seems that Travis CI is migrating all open source projects from travis-ci.org to travis-ci.com, and that we can't switch to the new Github App until out project is migrated from .org to .com. I e-mailed travis to have them switch us over.

https://docs.travis-ci.com/user/open-source-on-travis-ci-com/#existing-open-source-repositories-on-travis-ciorg

The upshot is that, at the moment, I have broken Travis :). I'll get it set up again as soon as they get back to me.

DanHickstein:
As far as I can tell, this issue is now solved. :)

PR #243, opened 2019-02-10
Updates to convert from travis-ci.org to travis-ci.com

DanHickstein:
This is in response to https://github.com/PyAbel/PyAbel/issues/242. 

So far, I think that the only update is that we need to change the link to the badge in README.rst from .org to .com. Part of this PR is a test to see if we can successfully trigger builds on travis-ci.com.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,8 @@
 PyAbel README
 =============
 
-.. image:: https://travis-ci.org/PyAbel/PyAbel.svg?branch=master
-    :target: https://travis-ci.org/PyAbel/PyAbel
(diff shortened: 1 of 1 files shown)


DanHickstein:
Bonus: previously, PyAbel test running on travis-ci.org took 3-to-8 minutes. On travis-ci.com, tests ran in 49 seconds! So, a nice speedup that might let us identify issues a bit quicker.

rth:
It looks like Travis CI passed OK.

DanHickstein:
Yeah, it seems that the builds are triggering just fine on Travis-ci.com just fine. Merging.

PR #244, opened 2019-02-26
New benchmark code

MikhailRyazanov:
This is the reworked `abel.benchmark.AbelTiming`.

The main changes are related to the basis caching now used by all the basis-set methods, plus that `basex` also got the forward transform (and variable regularization, as previously addressed in the experimental [bench-var](https://github.com/MikhailRyazanov/PyAbel/tree/bench-var)). It also implements the “run for at least _n_ seconds” idea, which help to get more stable results for fast stuff without running the slow stuff too many times.

Since it turned out that different methods require somewhat different benchmarking procedures, I have created here an individual function for each method, which benchmarks everything related to that method (basis-set generation, forward and inverse transforms — as applicable). This makes the code more localized and easier to modify and to accommodate future methods. It should also consume less memory, since instead of first generating basis sets for all the methods and keeping them all in RAM before testing the transforms, the procedure is now
```
for each method:
  generate method's basis
  test method's transforms
  clean-up
```
so only at most one basis set is kept in RAM.

There is a technical decision to benchmark `get_bs_cached()` for each method instead of the specific low-level basis-generating functions. The reason is that this makes the code simpler and uniform, while these functions have a relatively small overhead (even for `basex`) compared to the basis generation itself.

From the interface point of view, the `transform_repeat` is renamed to `repeat` because it also applies to basis-generation benchmarks (I think that it should, since they take about the same time as some transform methods). Also the `duration` parameter is added.

Let's discuss whether all this makes sense and what needs to be changed (naming, comments/docs, default parameters)...

(I also propose to add a `verbose` parameter for printing some progress information (to `stderr`) instead of being completely silent for seconds–minutes–hours and only at the end showing the summary.)

### DIFF
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -16,6 +16,7 @@ from . import tools
 
 from timeit import default_timer as timer
 import itertools
+import sys
 
 
 class Timent(object):
@@ -123,22 +124,24 @@ class AbelTiming(object):
                     'linbasex', 'onion_bordas, 'onion_peeling', 'two_point',
                     'three_point']
 
-    n_max_bs : int
-        since the basis-sets generation takes a long time, do not run
-        benchmarks with *n* > **n_max_bs** for implementations that use basis
-        sets
-    n_max_slow : int
-        do not run benchmarks with *n* > **n_max_slow** for "slow"
-        implementations, so far including only "direct_Python"
     repeat : int
         repeat each benchmark at least this number of times to get the average
         values
-    duration : float
+    t_min : float
         repeat each benchmark for at least this number of seconds to get the
         average values
+    t_max : float
+        do not benchmark methods at array sizes when this is expected to take
+        longer than this number of seconds. Notice that benchmarks for the
+        smallest size from **n** are always run and that the estimations can be
+        off by a factor of 2 or so.
+    verbose : boolean
+        determines whether benchmark progress should be reported (to stderr)
 
     Attributes
     -------
+    n : list of int
+        array sizes from the parameter **n**, sorted in ascending order
     bs, fabel, iabel : dict of list of float
         benchmark results — dictionaries for
 
@@ -149,9 +152,9 @@ class AbelTiming(object):
             iabel
                 inverse Abel transform
 
-        with methods as keys and lists of timings in milliseconds as entries
-        (corresponding to array sizes from **n**, also available as
-        :py:attr:`AbelTiming.n`).
+        with methods as keys and lists of timings in milliseconds as entries.
+        Timings correspond to array sizes in :py:attr:`AbelTiming.n`; for
+        skipped benchmarks (see **t_max**) they are ``np.nan``.
 
     Notes
     -----
@@ -162,13 +165,15 @@ class AbelTiming(object):
     :math:`O(n^3)` time complexity, so going from *n* = 501 to *n* = 5001
     would require about 100 times more memory and take about 1000 times longer.
     """
-    def __init__(self, n=[301, 501], select=['all', ], n_max_bs=700,
-                 n_max_slow=700, repeat=1, duration=0.1):
-        self.n = n
-        self.n_max_bs = n_max_bs
-        self.n_max_slow = n_max_slow
+    def __init__(self, n=[301, 501], select=['all', ],
+                 repeat=1, t_min=0.1, t_max=np.inf,
+                 verbose=True):
+        self.n = sorted(n)
+        self.repeat = repeat
+        self.t_max = t_max
+        self.verbose = verbose
         # create the timing function
-        self._time = Timent(skip=-1, repeat=repeat, duration=duration).time
+        self._time = Timent(skip=-1, repeat=repeat, duration=t_min).time
 
         # which methods need half and whole images
         need_half = frozenset([
@@ -212,6 +217,8 @@ class AbelTiming(object):
         self.bs = self.res['bs']
         self.fabel = self.res['forward']
         self.iabel = self.res['inverse']
+        # inverse speed for time estimations
+        self._pace = {}
 
         # Loop over all image sizes
         for ni in self.n:
@@ -222,6 +229,7 @@ class AbelTiming(object):
             # assumption that we are transforming just the "right side" of
             # a square image.
             # see: https://github.com/PyAbel/PyAbel/issues/207
+            self._vprint('n =', self.ni)
 
             # create needed images (half and/or whole)
             if methods & need_half:
@@ -231,11 +239,21 @@ class AbelTiming(object):
 
             # call benchmark (see below) for each method at this image size
             for method in methods:
+                self._vprint(' ', method)
                 getattr(self, '_time_' + method)()
 
             # discard images
             self.half_image = None
             self.whole_image = None
+        self._vprint('')
+
+    def _vprint(self, *args, **kwargs):
+        """
+        Print to stderr, only if verbose=True.
+        """
+        if self.verbose:
+            print(*args, file=sys.stderr, **kwargs)
+            sys.stderr.flush()  # (Python 3 buffers stderr. Why?!)
 
     def _append(self, kind, method, result):
         """
@@ -254,6 +272,57 @@ class AbelTiming(object):
         self._append(kind, method,
                      self._time(func, *args, **kwargs) * 1000)  # [s] to [ms]
 
+    def _skip(*param):
+        """
+        Decorator for benchmarking functions.
+        Adds a check whether the estimated execution time would exceed t_max.
+        If so, fills the results with np.nan, otherwise executes the
+        benchmarking code.
+
+        Parameters are tuples "(kind, method)". Either item can be a list, then
+        all combinations of kind(s) and method(s) are implied. Altogether the
+        set of these kind–method pairs must be the same as in the "normal"
+        execution results.
+        """
+        # (ensuring list type by wrapping non-list argument into a list)
+        def ensure_list(x):
+            return x if isinstance(x, list) else [x]
+        # assemble all kind–method pairs
+        res_keys = []
+        for p in param:
+            res_keys += itertools.product(*map(ensure_list, p))
+        res_keys = [(k, m + '_bs' if k == 'bs' else m) for k, m in res_keys]
+
+        def decorator(f):
+            method = f.__name__[6:]  # (remove initial "_time_")
+
+            def decorated(self):
+                # (rounding to n significant figures)
+                def roundsf(x, n):
+                    return float('{:.{p}g}'.format(x, p=n))
+                # get the estimated time (use 0 if cannot) and report it
+                t_est = self._pace.get(method, 0) * self.ni**3
+                self._vprint('    estimated ' +
+                             ('{:g} s'.format(roundsf(t_est, 2)) if t_est
+                              else '???'), end='')
+                # skip the benchmark if it would take too long
+                if t_est > self.t_max:
+                    self._vprint(' -- skipped')
+                    # fill the results with nan
+                    for k, m in res_keys:
+                        self._append(k, m, np.nan)
+                    return
+                else:  # otherwise run the benchmark
+                    f(self)
+                    # calculate the actual total time and report it
+                    t = (sum(self.res[k][m][-1] for k, m in res_keys) *
+                         self.repeat) / 1000  # [ms] -> [s]
+                    self._vprint(', actually {:.3f} s'.format(t))
+                    # save the pace for future estimations
+                    self._pace[method] = t / self.ni**3
+            return decorated
+        return decorator
+
     # Benchmarking functions for each method.
     # Must be named "_time_method", where "method" is as in "select".
     # Do not take or return anything, but use instance variables:
@@ -267,15 +336,9 @@ class AbelTiming(object):
     #     kind = 'bs' (basis), 'forward', 'inverse' -- as applicable
     #     method -- as above, but can also include variants (like in basex)
 
+    @_skip(('bs', 'basex'),
+           (['inverse', 'forward'], ['basex', 'basex(var.reg.)']))
     def _time_basex(self):
-        # skip if too large
-        if self.ni > self.n_max_bs:
-            self._append('bs', 'basex_bs', np.nan)
-            for direction in ['inverse', 'forward']:
-                for method in ['basex', 'basex(var.reg.)']:
-                    self._append(direction, method, np.nan)
-            return
-
         # benchmark the basis generation (default parameters)
         def gen_basis():
             basex.cache_cleanup()
@@ -307,35 +370,30 @@ class AbelTiming(object):
         # discard all caches
         basex.cache_cleanup()
 
+    @_skip((['inverse', 'forward'], 'direct_C'))
     def _time_direct_C(self):
         for direction in ['inverse', 'forward']:
             self._benchmark(direction, 'direct_C',
                             direct.direct_transform,
                             self.half_image, direction=direction, backend='C')
 
+    @_skip((['inverse', 'forward'], 'direct_Python'))
     def _time_direct_Python(self):
         for direction in ['inverse', 'forward']:
-            if self.ni > self.n_max_slow:  # skip if too large
-                self._append(direction, 'direct_Python', np.nan)
-            else:
-                self._benchmark(direction, 'direct_Python',
-                                direct.direct_transform,
-                                self.half_image, direction=direction,
-                                backend='python')
+            self._benchmark(direction, 'direct_Python',
+                            direct.direct_transform,
+                            self.half_image, direction=direction,
+                            backend='python')
 
+    @_skip((['inverse', 'forward'], 'hansenlaw'))
     def _time_hansenlaw(self):
         for direction in ['inverse', 'forward']:
             self._benchmark(direction, 'hansenlaw',
                             hansenlaw.hansenlaw_transform,
                             self.half_image, direction=direction)
 
+    @_skip((['bs', 'inverse'], 'linbasex'))
     def _time_linbasex(self):
-        # skip if too large
-        if self.ni > self.n_max_bs:
-            self._append('bs', 'linbasex_bs', np.nan)
-            self._append('inverse', 'linbasex', np.nan)
-            return
-
         # benchmark the basis generation (default parameters)
         def gen_basis():
             linbasex.cache_cleanup()
@@ -353,6 +411,7 @@ class AbelTiming(object):
         # discard all caches
         linbasex.cache_cleanup()
 
+    @_skip(('inverse', 'onion_bordas'))
     def _time_onion_bordas(self):
         self._benchmark('inverse', 'onion_bordas',
                         onion_bordas.onion_bordas_transform,
@@ -360,12 +419,6 @@ class AbelTiming(object):
 
     # (Generic function for all Dasch methods; not called directly.)
     def _time_dasch(self, method):
-        # skip if too large
-        if self.ni > self.n_max_bs:
-            self._append('bs', method + '_bs', np.nan)
-            self._append('inverse', method, np.nan)
-            return
-
         # benchmark the basis generation (default parameters)
         def gen_basis():
             dasch.cache_cleanup()
@@ -383,12 +436,15 @@ class AbelTiming(object):
         # discard all caches
         dasch.cache_cleanup()
 
+    @_skip((['bs', 'inverse'], 'onion_peeling'))
     def _time_onion_peeling(self):
         self._time_dasch('onion_peeling')
 
+    @_skip((['bs', 'inverse'], 'two_point'))
     def _time_two_point(self):
         self._time_dasch('two_point')
 
+    @_skip((['bs', 'inverse'], 'three_point'))
     def _time_three_point(self):
         self._time_dasch('three_point')
 
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -19,6 +19,20 @@ import itertools
 import sys
 
 
+def _ensure_list(x):
+    """
+    Wrap the argument in a list, if it is not a list already.
+    """
+    return x if isinstance(x, list) else [x]
+
+
+def _roundsf(x, n):
+    """
+    Round to n significant digits
+    """
+    return float('{:.{p}g}'.format(x, p=n))
+
+
 class Timent(object):
     """
     Helper class for measuring execution times.
@@ -114,10 +128,10 @@ class AbelTiming(object):
 
     Parameters
     ----------
-    n : list of int
-        array sizes for the benchmark (assuming 2D square arrays (*n*, *n*))
-    select : list of str
-        methods to benchmark. Use ``['all']`` (default) for all available or
+    n : int or list of int
+        array size(s) for the benchmark (assuming 2D square arrays (*n*, *n*))
+    select : str or list of str
+        methods to benchmark. Use ``'all'`` (default) for all available or
         choose any combination of individual methods::
 
             select=['basex', 'direct_C', 'direct_Python', 'hansenlaw',
@@ -165,10 +179,11 @@ class AbelTiming(object):
     :math:`O(n^3)` time complexity, so going from *n* = 501 to *n* = 5001
     would require about 100 times more memory and take about 1000 times longer.
     """
-    def __init__(self, n=[301, 501], select=['all', ],
+    def __init__(self, n=[301, 501], select='all',
                  repeat=1, t_min=0.1, t_max=np.inf,
                  verbose=True):
-        self.n = sorted(n)
+        self.n = sorted(_ensure_list(n))
+        select = _ensure_list(select)
         self.repeat = repeat
         self.t_max = t_max
         self.verbose = verbose
@@ -302,25 +317,19 @@ class AbelTiming(object):
         set of these kind–method pairs must be the same as in the "normal"
         execution results.
         """
-        # (ensuring list type by wrapping non-list argument into a list)
-        def ensure_list(x):
-            return x if isinstance(x, list) else [x]
         # assemble all kind–method pairs
         res_keys = []
         for p in param:
-            res_keys += itertools.product(*map(ensure_list, p))
+            res_keys += itertools.product(*map(_ensure_list, p))
 
         def decorator(f):
             method = f.__name__[6:]  # (remove initial "_time_")
 
             def decorated(self):
-                # (rounding to n significant figures)
-                def roundsf(x, n):
-                    return float('{:.{p}g}'.format(x, p=n))
                 # get the estimated time (use 0 if cannot) and report it
                 t_est = self._pace.get(method, 0) * self.ni**3
                 self._vprint('    estimated ' +
-                             ('{:g} s'.format(roundsf(t_est, 2)) if t_est
+                             ('{:g} s'.format(_roundsf(t_est, 2)) if t_est
                               else '???'), end='')
                 # skip the benchmark if it would take too long
                 if t_est > self.t_max:
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -107,167 +107,289 @@ class Timent(object):
 
 
 class AbelTiming(object):
-    def __init__(self, n=[301, 501], select=["all", ], n_max_bs=700,
-                 n_max_slow=700, transform_repeat=1):
-        """
-        Benchmark performance of different iAbel/fAbel implementations.
+    """
+    Benchmark performance of different Abel implementations
+    (basis generation, forward and inverse transforms, as applicable).
 
-        Parameters
-        ----------
-        n: integer
-            a list of arrays sizes for the benchmark
-            (assuming 2D square arrays (n,n))
+    Parameters
+    ----------
+    n : list of int
+        array sizes for the benchmark (assuming 2D square arrays (*n*, *n*))
+    select : list of str
+        methods to benchmark. Use ``['all']`` (default) for all available or
+        choose any combination of individual methods::
 
-        select: list of str
-            list of transforms to benchmark select=['all',] (default) or 
-            choose transforms:
-            select=['basex', 'direct_Python', 'direct_C', 'hansenlaw', 
-                    'linbasex' 'onion_bordas, 'onion_peeling', 'two_point', 
+            select=['basex', 'direct_C', 'direct_Python', 'hansenlaw',
+                    'linbasex', 'onion_bordas, 'onion_peeling', 'two_point',
                     'three_point']
 
-        n_max_bs: integer
-            since the basis sets generation takes a long time,
-            do not run this benchmark for implementations that use basis sets
-            for n > n_max_bs
+    n_max_bs : int
+        since the basis-sets generation takes a long time, do not run
+        benchmarks with *n* > **n_max_bs** for implementations that use basis
+        sets
+    n_max_slow : int
+        do not run benchmarks with *n* > **n_max_slow** for "slow"
+        implementations, so far including only "direct_Python"
+    repeat : int
+        repeat each benchmark at least this number of times to get the average
+        values
+    duration : float
+        repeat each benchmark for at least this number of seconds to get the
+        average values
 
-        n_max_slow: integer
-            maximum n run for the "slow" transform methods, so far including 
-            only the "direct_python" implementation.
-        """
-        from timeit import Timer, default_timer
-        import time
+    Attributes
+    -------
+    bs, fabel, iabel : dict of list of float
+        benchmark results — dictionaries for
 
+            bs
+                basis-set generation
+            fabel
+                forward Abel transform
+            iabel
+                inverse Abel transform
+
+        with methods as keys and lists of timings in milliseconds as entries
+        (corresponding to array sizes from **n**, also available as
+        :py:attr:`AbelTiming.n`).
+
+    Notes
+    -----
+    The results can be output in a nice format by simply
+    ``print(AbelTiming(...))``.
+
+    Keep in mind that most methods have :math:`O(n^2)` memory and
+    :math:`O(n^3)` time complexity, so going from *n* = 501 to *n* = 5001
+    would require about 100 times more memory and take about 1000 times longer.
+    """
+    def __init__(self, n=[301, 501], select=['all', ], n_max_bs=700,
+                 n_max_slow=700, repeat=1, duration=0.1):
         self.n = n
+        self.n_max_bs = n_max_bs
+        self.n_max_slow = n_max_slow
+        # create the timing function
+        self._time = Timent(skip=-1, repeat=repeat, duration=duration).time
 
-        transform = {
-            'basex': basex.basex_core_transform,
-            'basex_bs': basex.get_bs_cached,
-            'direct_Python': direct.direct_transform,
-            'direct_C': direct.direct_transform,
-            'hansenlaw': hansenlaw.hansenlaw_transform,
-            'linbasex': linbasex._linbasex_transform_with_basis,
-            'linbasex_bs': linbasex._bs_linbasex,
-            'onion_bordas': onion_bordas.onion_bordas_transform,
-            'onion_peeling': dasch.dasch_transform,
-            'onion_peeling_bs': dasch._bs_onion_peeling,
-            'two_point': dasch.dasch_transform,
-            'two_point_bs': dasch._bs_two_point,
-            'three_point': dasch.dasch_transform,
-            'three_point_bs': dasch._bs_three_point,
-         }
+        # which methods need half and whole images
+        need_half = frozenset([
+            'basex',
+            'direct_C',
+            'direct_Python',
+            'hansenlaw',
+            'onion_bordas',
+            'onion_peeling',
+            'two_point',
+            'three_point',
+        ])
+        need_whole = frozenset([
+            'linbasex',
+        ])
+        # all available methods (= union of the above sets)
+        all_methods = need_half | need_whole
+        # remove direct_C, if not supported
+        if not direct.cython_ext:
+            all_methods = all_methods - frozenset(['direct_C'])
 
-        # result dicts
-        res = {}
-        res['bs'] = {'basex_bs': [], 'linbasex_bs': [], 'onion_peeling_bs': [], 
-                     'two_point_bs': [], 'three_point_bs': []}
-        res['forward'] = {'direct_Python': [], 'hansenlaw': []}
-        res['inverse'] = {'basex': [], 'direct_Python': [], 'hansenlaw': [],
-                          'linbasex': [],
-                          'onion_bordas': [], 'onion_peeling': [],
-                          'two_point': [], 'three_point': []}
-
-        if direct.cython_ext:
-            res['forward']['direct_C'] = []
-            res['inverse']['direct_C'] = []
-
-        # delete all keys not present in 'select' input parameter
-        if "all" not in select:
-            for trans in select:
-                if trans not in res['inverse'].keys():
-                    raise ValueError("'{}' is not a valid transform method".
-                                     format(trans), res['inverse'].keys())
-                     
-            for direction in ['forward', 'inverse']:
-                rm = []
-                for abel in res[direction]:
-                    if abel not in select:
-                        rm.append(abel)
-                for x in rm:
-                    del res[direction][x]
-            # repeat for 'bs' which has append '_bs'
-            rm = []
-            for abel in res['bs']:
-                if abel[:-3] not in select:
-                    rm.append(abel)
-            for x in rm:
-                del res['bs'][x] 
-
-        # ---- timing tests for various image sizes nxn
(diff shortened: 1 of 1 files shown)


stggh:
Beautiful work @MikhailRyazanov! Well done in decoding the complexities of each transform call.

`abel.benchmark.AbelTiming()` looks very good. 

Minor comments:
 (1) I do not like the use of (legacy) random number images, since the transform cannot be verified, and the timing may be sensitive to the actual array values. e.g. the random whole-image passed to `linebasex` is not identical to the random half-image passed to the other methods.

Either the Dribinski sample image could be used (and optionally, the transform verified), or better, the ability to pass a specific image to `benchmark`, to provide for relative method timings on a specific image character. 

 (2)  I find the legacy `n_max_bs` and `ns_max_slow` a pain. My default test `n=1001` gives:
```
In [36]: abel.benchmark.AbelTiming(n=[1001], select=['basex', 'three_point', ]) 
Out[36]: 

time in milliseconds

Implementation      n = 1001      
        basex_bs       nan         
  three_point_bs       nan         


Implementation      n = 1001      
           basex       nan         
 basex(var.reg.)       nan         


Implementation      n = 1001      
           basex       nan         
 basex(var.reg.)       nan         
     three_point       nan         
```
which is not very informative for a user. Can we eliminate `n_max_` and rely on `duration` to keep things sane?

MikhailRyazanov:
@stggh,

(1) I think, the “verified” part should be done in the unit tests. :–) But the randomness might (in principle) affect the results. The problem with the Dribinski sample image is that it has a fixed size. We can think about parametrizing it or invent some other synthetic test image. The idea about passing a specific image is also good, but we will need images for each resolution and separately for inverse/forward transforms.

(2) Yes, now the defaults `n_max_...=701` look very small. I thought that we can change them to `n_max_...=np.inf`, such that there are no limits by default, but they can be imposed manually when needed. (By the way, I think that `basex` should be also considered “slow”, since it requires basis-set generation, which is close to `direct_Python` in terms of speed.)

The `duration` does not limit the _maximal_ time, so it would not work as a substitution for `n_max_...`. However, in principle, if we know the approximate scaling, we can estimate the time requirements for larger sizes from the data for smaller sizes and then skip the benchmarks that would take longer than `t_max` (and rename `duration` to `t_min`).

stggh:
>  should be done in the unit tests

  good point. 

> Dribinski sample image is that it has a fixed size

   ```python
image = abel.tools.analytical.SampleImage(n).image
```
  but note that, this image should be first forward transformed before feeding back to the other inverse Abel transforms.

> passing a specific image 

  This is the users choice, with the aim for them to determine the fastest PyAbel method for their specific image type/size. 

> estimate the time requirements for larger sizes from the data for smaller sizes 

This is a very good idea!  (and the time estimate could be printed before the execution?)

**Comment by stggh** [comment]
Neat! and it works!

>  (and the time estimate could be printed before the execution?)

Perhaps, only if `verbose=True`

**Comment by MikhailRyazanov** [comment]
The estimates are printed only with `verbose=True`, of course. :–) The rest should work with `verbose=False` as well.

**Comment by MikhailRyazanov** [comment]
Here is a description of the last changes. The new approach tries to
> estimate the time requirements for larger sizes from the data for smaller sizes and then skip the benchmarks that would take longer than `t_max`

This is done by extrapolating the data from the largest previous size (thus the sizes are now sorted, just in case) using the hard-coded cubic scaling. This is not ideal, since, for example, `hansenlaw` is closer to quadratic, and massively parallel systems can have sub-cubic matrix multiplication (approach quadratic scaling for infinite parallelism). But it should be good enough to get the correct order of magnitude, and making better approximations that would be reliable in all cases is difficult (in principle, the local scaling power can be estimated from the last _two_ points, but this is much less robust against occasional glitches).

**Comment by stggh** [comment]
Opps! I did not realise `verbose=True` was the default ...

Great work. I like it!

**Comment by MikhailRyazanov** [comment]
@stggh, I have overlooked that `SampleImage(n)` actually scales the rings with the image size... So it can be used indeed. I would remark, however, that scaling the peak widths is not a good idea, since for larger sizes the image becomes “blurrier” (without sharp features), but even worse is that for small images the peaks become narrower than the pixel size. I think, the widths should correspond to `sigma` in pixels regardless of the image size.

> this image should be first forward transformed before feeding back to the other inverse Abel transforms

Then we can probably use one of the fast methods (`two_point`?).
The other possibility is to use something polynomial (in polar coordinates), since it has a quickly computable analytical Abel transform. And I even know how this can be done, but it will take more time to implement.

> This is the users choice, with the aim for them to determine the fastest PyAbel method for their specific image type/size.

I totally agree. The problem here is that the user probably also needs to impose some symmetry, so the benchmarks actually should be done with the high-level transform functions. This in principle should be not a problem, since now we have caching, but this needs some contemplation...

**Comment by MikhailRyazanov** [comment]
While thinking about using non-random/user-supplied images, I want to make a couple of minor changes.

1. Since I already have `ensure_list()` in the code, it can be used to make the interface more user-friendly. Namely, to allow doing something like `AbelTiming(n=1234, method='two_point')` instead of requiring `AbelTiming(n=[1234], method=['two_point'])`.

2. Modify the output. Change this:
```

Implementation      n = 513           n = 1025          n = 2049      
----------------------------------------------------------------------------
        basex_bs    1658.0               nan               nan         
     linbasex_bs      20.9              77.2               nan         
onion_peeling_bs       4.3              18.2              95.6         
  three_point_bs       9.8              41.8             195.8         
    two_point_bs       3.3              14.0              53.3         


Implementation      n = 513           n = 1025          n = 2049      
----------------------------------------------------------------------------
           basex       1.3               nan               nan         
 basex(var.reg.)      11.4               nan               nan         
   direct_Python     385.6               nan               nan         
       hansenlaw      27.9             111.7             414.4         


Implementation      n = 513           n = 1025          n = 2049      
----------------------------------------------------------------------------
           basex       1.3               nan               nan         
 basex(var.reg.)       9.9               nan               nan         
   direct_Python     401.9               nan               nan         
       hansenlaw      29.5             113.5             420.5         
        linbasex     221.7            1102.1               nan         
    onion_bordas     302.2            1819.6               nan         
   onion_peeling       1.4              10.3              75.2         
     three_point       1.3              10.0              73.4         
       two_point       1.3              10.3              72.9         
```
to something like this:
```
========== Basis generation ==========

Method            n = 513   n = 1025   n = 2049
-----------------------------------------------
basex              1658.0        nan        nan
linbasex             20.9       77.2        nan
onion_peeling         4.3       18.2       95.6
three_point           9.8       41.8      195.8
two_point             3.3       14.0       53.3


========== Forward Abel transform ==========

Method            n = 513   n = 1025   n = 2049
-----------------------------------------------
basex                 1.3        nan        nan
basex(var.reg.)      11.4        nan        nan
direct_Python       385.6        nan        nan
hansenlaw            27.9      111.7      414.4


========== Inverse Abel transform ==========

Method            n = 513   n = 1025   n = 2049
-----------------------------------------------
basex                 1.3        nan        nan
basex(var.reg.)       9.9        nan        nan
direct_Python       401.9        nan        nan
hansenlaw            29.5      113.5      420.5
linbasex            221.7     1102.1        nan
onion_bordas        302.2     1819.6        nan
onion_peeling         1.4       10.3       75.2
three_point           1.3       10.0       73.4
two_point             1.3       10.3       72.9
```
Namely:
* Use more relevant (I hope) section titles.
* Change the column header “Implementation” to “Method”.
* Remove the “_bs” suffixes (from the `AbelTiming.bs` dictionary keys also). I do not think that they are needed for anything, but without them the output will be cleaner and the code slightly simpler.
* Left-align method names.
* Reduce the spacing between columns (to something like 3 spaces between the longest entries).

Any thoughts?

**Comment by stggh** [comment]
> `ensure_list()`

Looks good.

> * Remove the “_bs” suffixes (from the `AbelTiming.bs` dictionary keys also)

How are the basis calculation times accessed?
(older version has `x.bs`):

screen shot 2019-02-27 at 18 45 27

**Comment by MikhailRyazanov** [comment]
Through the same `x.bs`, but as `x.bs['two_point']` instead of `x.bs['two_point_bs']`. I think, one “bs” is enough.

**Comment by stggh** [comment]
Ah, good point. Thanks.

**Comment by MikhailRyazanov** [comment]
OK, now it can accept single values for `n` and `select`, and the output looks like this:
```
PyAbel benchmark run on AMD64 Family 23 Model 17 Stepping 0, AuthenticAMD

time in milliseconds

=== Basis generation ======================================================

Method              n = 513    n = 1025    n = 2049    n = 4097   n = 10001
---------------------------------------------------------------------------
basex                1626.5      6673.9     45346.5         nan         nan
linbasex               28.9       139.4       476.9      1946.4         nan
onion_peeling           5.2        23.3       118.4       539.8      5232.7
three_point             9.2       108.1       439.3      1701.0     10479.6
two_point               2.6        20.3        95.6       369.3      2232.0


=== Forward Abel transform ================================================

Method              n = 513    n = 1025    n = 2049    n = 4097   n = 10001
---------------------------------------------------------------------------
basex                   2.9        10.1        68.2         nan         nan
basex(var.reg.)        20.1        37.8       223.8         nan         nan
direct_Python         549.3      9100.7     78167.3         nan         nan
hansenlaw              23.9        88.9       382.3      1966.3     18347.4


=== Inverse Abel transform ================================================

Method              n = 513    n = 1025    n = 2049    n = 4097   n = 10001
---------------------------------------------------------------------------
basex                   1.7        10.9        67.2         nan         nan
basex(var.reg.)        10.0        39.0       224.3         nan         nan
direct_Python         624.0      8986.4     78444.2         nan         nan
hansenlaw              25.8        88.8       399.7      2068.0     26515.3
linbasex              211.6       975.3      5574.7     34552.3         nan
onion_bordas          397.2      2280.3     15464.2    116495.5         nan
onion_peeling           1.5        10.3       113.0       452.8      5630.4
three_point             1.7        10.9        67.9       479.1      5538.1
two_point               1.7        10.5        67.7       479.4      5532.2
```
(the standard terminal width is now enough for up to 5 image sizes instead of only 3).

**Comment by stggh** [comment]
Looks good. I did not have the patience to wait. Here is the `<ctrl>c` output, after a while.
Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
```
abel.benchmark.AbelTiming(n=[513, 1025, 2049, 4097, 10001], verbose=False)

time in milliseconds

=== Basis generation ======================================================

Method              n = 513    n = 1025    n = 2049    n = 4097   n = 10001
---------------------------------------------------------------------------
basex                1290.4      7607.6     35134.0    172639.4         nan
linbasex               10.3        49.2       211.4      1038.8         nan
onion_peeling           2.3         9.0        43.3       263.4      2603.0
three_point             6.5        30.5       123.8       652.9      5040.1
two_point               2.1         7.8        32.7       165.4      1309.7


=== Forward Abel transform ================================================

Method              n = 513    n = 1025    n = 2049    n = 4097   n = 10001
---------------------------------------------------------------------------
basex                   0.5        20.5       140.4       870.7         nan
basex(var.reg.)         5.1        49.5       255.7      1536.5         nan
direct_C               21.6       161.0      1242.1     10311.3         nan
direct_Python         254.2      1995.5     20074.2    228319.6         nan
hansenlaw              21.7        84.7       458.5      1907.7         nan


=== Inverse Abel transform ================================================

Method              n = 513    n = 1025    n = 2049    n = 4097   n = 10001
---------------------------------------------------------------------------
basex                   0.5         4.6        28.9       814.0         nan
basex(var.reg.)         4.3        21.8       120.8       843.0         nan
direct_C               22.4       165.9      1282.0     11064.5         nan
direct_Python         259.8      1997.4     20488.3    227257.8         nan
hansenlaw              21.8        85.1       464.4      1957.7         nan
linbasex               94.3       440.4      3716.3     28248.0         nan
onion_bordas          401.9      2027.5     18628.2    143418.6         nan
onion_peeling           0.5         3.8        28.9       222.4      3036.4
three_point             0.5         3.4        26.2       204.2      2724.6
two_point               0.5         3.4        25.6       203.4      2754.6
```

```
In [8]: abel.benchmark.AbelTiming(n=1025, verbose=False)                        
Out[8]: 

time in milliseconds

=== Basis generation ======

Method             n = 1025
---------------------------
basex                7638.3
linbasex               43.6
onion_peeling           9.0
three_point            26.3
two_point               7.9


=== Forward Abel transform ===

Method             n = 1025
---------------------------
basex                  20.3
basex(var.reg.)        49.3
direct_C              161.3
direct_Python        1998.7
hansenlaw              86.7


=== Inverse Abel transform ===

Method             n = 1025
---------------------------
basex                   3.8
basex(var.reg.)        21.3
direct_C              166.9
direct_Python        1988.9
hansenlaw              87.2
linbasex              453.9
onion_bordas         2046.0
onion_peeling           3.9
three_point             3.5
two_point               3.4
```

**Comment by MikhailRyazanov** [comment]
Thanks, @stggh!

I see that you also have this strange behavior of `basex`: for sufficiently large images (_n_ ≳ 1000) the forward transform is much slower than the inverse. The most surprising for me is that when I run it on my laptop (Intel i5, GNU/Linux), I see the huge difference:
```
=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   1.3        50.1       299.5
basex(var.reg.)        11.5       106.3       625.3


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   1.2        10.3        85.9
basex(var.reg.)        10.0        51.7       356.9
```
but on an AMD Ryzen, Win10 desktop the forward is as fast as the inverse:
```
=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   1.5        10.0        66.8
basex(var.reg.)         6.9        36.9       221.8


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   1.5        10.0        66.7
basex(var.reg.)         7.3        36.7       222.7
```

@DanHickstein, @DhrubajyotiDas, @rth, could you please try
```
import abel
print(abel.benchmark.AbelTiming(n=[513, 1025, 2049], select='basex', repeat=3, t_min=10))
```
on whatever systems you have (it should take ~5 minutes) and report the results here?

**Comment by stggh** [comment]
Here is a repeat of your command on my i7:
```
In [1]: import abel                                                             

In [2]: abel.benchmark.AbelTiming(n=[513, 1025, 2049], select='basex', repeat=3,
   ...: t_min=10)                                                               
n = 513
  basex
    estimated ???, actually 4.537 s
n = 1025
  basex
    estimated 36 s, actually 30.273 s
n = 2049
  basex
    estimated 240 s, actually 130.427 s

Out[2]: 

time in milliseconds

=== Basis generation ==============================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                1388.5      9809.2     42702.5


=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   0.8        19.5       149.8
basex(var.reg.)        57.4       142.4       336.0


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   0.8         5.8        45.1
basex(var.reg.)        65.0       114.0       242.4
```
and Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz
```
In [1]: import abel

In [2]: abel.__path__
Out[2]: ['/users/vuv1/stg121/Software/Python/PyAbel/PyAbel-MR-new-benchmark/abel']

In [3]: abel.benchmark.AbelTiming(n=[513, 1025, 2049], select='basex', repeat=3,
   ...: t_min=10)
n = 513
  basex
    estimated ???, actually 6.529 s
n = 1025
  basex
    estimated 52 s, actually 59.591 s
n = 2049
  basex
    estimated 480 s, actually 258.764 s

Out[3]: 

time in milliseconds

=== Basis generation ==============================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                2156.6     19707.2     85738.3


=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   0.5         7.5        40.9
basex(var.reg.)         8.8        86.0       235.7


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   0.5         2.5        14.4
basex(var.reg.)        10.0        60.6       225.3

```

Intel definitely prefers the inverse transform.

**Comment by MikhailRyazanov** [comment]
@stggh, your Intel seems to be even worse than my: look at the ratios between `basex(var.reg.)` and `basex`...

While nobody else took care to run the benchmarks, I was trying to investigate what causes this strange behavior and found that it is most probably related to [subnormal numbers](https://en.wikipedia.org/wiki/Denormal_number) appearing in some of these calculations.

I've made a [benchmark](https://gist.github.com/MikhailRyazanov/63823e1bc7028085389d787b765a4f2c) for matrix multiplications and inversions to see how their speed depends on the matrix content. There are some results included in the gist, which make clear that indeed Intel dislikes subnormals very much, while AMD works with them as fast as with normal numbers. (This is especially shameful, since it was Intel's idea to include subnormals in the IEEE 754 standard. And now, like in the [_Meltdown_ case](https://en.wikipedia.org/wiki/Meltdown_(security_vulnerability)#Affected_hardware), they have “optimized” their CPUs to the point that their high-end models became slower that a 100-dollars AMD's CPU...)

The results in the gist are with the OpenBLAS libraries for both CPUs. I've also tried ATLAS and the reference BLAS on the i5 — they also show a noticeable performance degradation for subnormals (although smaller, probably since they both are already several times slower). It would be interesting to see how Intel MKL behaves (I don't have it).

The sad thing is that while we actually do not need that much accuracy and could live without submormals, and CPUs even have flags to disable them, there is no reasonable way to control this either from numpy (see https://github.com/numpy/numpy/issues/4581) or for OpenBLAS itself (see https://github.com/xianyi/OpenBLAS/issues/1237). The benchmark above has tests for a partial workaround by zeroing small elements before matrix operations — if nothing else helps, we can include it as an option in the PyAbel code. At least on my i5, it solves the problem for `basex` and somewhat helps for `basex(var.reg.)`. (I also wonder whether it can affect `onion_bordas`, since on my machines it is much faster than `direct_Python`, whereas on @stggh's they are very similar.)

(For those who can try to disable subnormals, there is a [test](https://gist.github.com/MikhailRyazanov/1444577a7a3811fcce491c6db8553283) to see whether they are really disabled or not. The numbers like `...e-320` are subnormal in the usual Python's `float`.)

File: abel/benchmark.py [hunk 1]
```diff
@@ -14,199 +14,504 @@
...
-            only the "direct_python" implementation.
+        Notes
+        -----
+        The measurements overhead can be estimated by executing ::
+
+            Timent(...).time(lambda: None)
+
+        with a sufficiently large number of iterations (to avoid rounding
+        errors due to the finite timer precision).
+        In 2018, this overhead was on the order of 100 ns per iteration.
```
DanHickstein: It's 2019 now :)

DanHickstein:
Great work Mikhail!! 

The benchmarks code has long needed an update, and what you have done is really awesome. 

Sorry for being absent from the conversation. The issue with the subnormal numbers is fascinating. 

Here are some benchmarks with my "Early 2015" MacBook with a dual-core intel
(Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz (https://browser.geekbench.com/processors/1672))

I observe the same behavior with the inverse transform being much faster than the forward:

```
PyAbel benchmark run on i386

time in milliseconds

=== Basis generation ==============================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                1009.1      7510.6     50709.5


=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   1.8        44.9       318.3
basex(var.reg.)        10.5       103.8       570.1


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   1.1         9.9        98.7
basex(var.reg.)        10.3        52.2       367.5
```

stggh:
@MikhailRyazanov neat! A very comprehensive investigation, on a topic that I had not even heard of.

Your [test](https://gist.github.com/MikhailRyazanov/1444577a7a3811fcce491c6db8553283) produces the same output on my i7. 

There is python code [daz](https://github.com/chainer/daz) to turn off the *dnormals*:
```
python test_subnormals.py 
i7 with subnormals
[1.e-308 1.e-309 1.e-310 1.e-311 1.e-312 1.e-313 1.e-314 1.e-315 1.e-316
 1.e-317 1.e-318 1.e-319 1.e-320 1.e-321 1.e-322 1.e-323 0.e+000]
[1.00000000e-308 9.99204336e-310 9.98409306e-311 9.97614908e-312
 9.96821142e-313 9.96028008e-314 9.95235505e-315 9.94443632e-316
 9.93652376e-317 9.92861970e-318 9.92073936e-319 9.91293312e-320
 9.90601620e-321 9.88131292e-322 9.88131292e-323 9.88131292e-324]
[[1.e-320 0.e+000 0.e+000 0.e+000 0.e+000]
 [2.e-320 1.e-320 0.e+000 0.e+000 0.e+000]
 [3.e-320 2.e-320 1.e-320 0.e+000 0.e+000]
 [4.e-320 3.e-320 2.e-320 1.e-320 0.e+000]
 [5.e-320 4.e-320 3.e-320 2.e-320 1.e-320]]
[[4.0e-320 0.0e+000 0.0e+000 0.0e+000 0.0e+000]
 [8.0e-320 4.0e-320 0.0e+000 0.0e+000 0.0e+000]
 [1.2e-319 8.0e-320 4.0e-320 0.0e+000 0.0e+000]
 [1.6e-319 1.2e-319 8.0e-320 4.0e-320 0.0e+000]
 [2.0e-319 1.6e-319 1.2e-319 8.0e-320 4.0e-320]]


i7 without subnormals
[0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
  File "test_subnormals.py", line 19, in <module>
    print(np.geomspace(1.0e-308, 1.0e-323, 323 - 308 + 1))
  File "/home/stg121/Software/Python/anaconda3/lib/python3.7/site-packages/numpy/core/function_base.py", line 326, in geomspace
    raise ValueError('Geometric sequence cannot include zero')
ValueError: Geometric sequence cannot include zero
```

When used with `PyAbel` it makes some improvement:
```
n [5]: abel.benchmark.AbelTiming(n=[513, 1025, 2049], select=['basex'], verbose=False)                  
Out[5]: 

time in milliseconds

=== Basis generation ==============================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                2158.5     11781.3     51260.6


=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   4.4        43.6       190.3
basex(var.reg.)       113.1       210.9       573.9


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   4.6        12.5        44.8
basex(var.reg.)       117.3       116.6       456.0

In [6]: import daz                                                                                       

In [7]: daz.set_ftz()                                                                                    

In [8]: daz.set_daz()                                                                                    

In [9]: abel.benchmark.AbelTiming(n=[513, 1025, 2049], select=['basex'], verbose=False)                  
Out[9]: 

time in milliseconds

=== Basis generation ==============================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                2128.0     12897.0     49352.7


=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   5.9        14.7        94.3
basex(var.reg.)        39.9        67.6       445.6


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   7.4        15.7        60.2
basex(var.reg.)       125.3       191.5       347.0
``

File: abel/benchmark.py (hunk 1 above)
Comment (MikhailRyazanov) [reply abel/benchmark.py]: It still _is_. I'm not a prophet to write “In 2019 ... was”.  ;–)
And, I think, none of us actually has a 2019 system (I have one from 2018 and one from 2015).

**Comment by MikhailRyazanov** [comment]
@stggh, thanks for trying daz!
> ```
> i7 without subnormals
> [0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
> ...
> ValueError: Geometric sequence cannot include zero
> ```
It actually works!

Could you please rerun these `AbelTiming`s with `t_min=5` for better accuracy (I'm surprised that disabling subnormals makes some calculations slower and hope that this is just an artifact).

@stggh and @DanHickstein, could you please try the [bm-mat.py](https://gist.github.com/MikhailRyazanov/63823e1bc7028085389d787b765a4f2c) tests? If possible, with [daz](https://github.com/chainer/daz) also. And tell me which BLAS/LAPACK implementation you have?

**Comment by MikhailRyazanov** [comment]
> ```
>         Z       N       J       X       Y      Xc      Yc   
> X       4.75    7.05    6.37    4.99    5.34    4.00    5.12
> Y       3.48    3.52    3.51   13.08   12.61    5.25    9.27
>
> with daz ---
> X       3.62    2.77    2.76    2.80    2.63    2.75    2.78
> Y       2.67    2.69    2.68    2.70    2.65    2.71    2.73
> ```
Seems to help.

The inversions, though, are terribly slow
> ```
>         Z       N       J       X       Y      Xc      Yc   
> inv           101.14           94.02   91.07   96.47   87.26
>
> with daz ---
> inv            80.07           40.08   66.76   92.74   72.55
> ```
compared to my (much cheaper!) systems:
```
AMD Ryzen 3 2200G, 3.5 GHz:
inv             8.63            8.32    8.08    8.32    7.99

Intel Core i5-4278U, 2.60 GHz:
inv            11.14           18.04   13.04   10.87   10.50
```

May I also see the `n = 1001, repeat = 100` results for completeness?

> Not sure whether this query provides the right information

I suspect, `mkl_rt` is what it is. Do you have it in your `/home/stg121/Software/Python/anaconda3/lib/`?

**Comment by stggh** [comment]
> mkl_rt is what it is. Do you have it in your

Yes.
```
[etalon:.+n/anaconda3/lib]$ ls libmkl*
libmkl_avx2.so                  libmkl_intel_thread.so
libmkl_avx512_mic.so            libmkl_mc3.so
libmkl_avx512.so                libmkl_mc.so
libmkl_avx.so                   libmkl_pgi_thread.so
libmkl_blacs_intelmpi_ilp64.so  libmkl_rt.so
libmkl_blacs_intelmpi_lp64.so   libmkl_scalapack_ilp64.so
libmkl_blacs_openmpi_ilp64.so   libmkl_scalapack_lp64.so
libmkl_blacs_openmpi_lp64.so    libmkl_sequential.so
libmkl_blacs_sgimpt_ilp64.so    libmkl_tbb_thread.so
libmkl_blacs_sgimpt_lp64.so     libmkl_vml_avx2.so
libmkl_cdft_core.so             libmkl_vml_avx512_mic.so
libmkl_core.so                  libmkl_vml_avx512.so
libmkl_def.so                   libmkl_vml_avx.so
libmkl_gf_ilp64.so              libmkl_vml_cmpt.so
libmkl_gf_lp64.so               libmkl_vml_def.so
libmkl_gnu_thread.so            libmkl_vml_mc2.so
libmkl_intel_ilp64.so           libmkl_vml_mc3.so
libmkl_intel_lp64.so            libmkl_vml_mc.so
```

I will run `1001` and repeat `100`.

**Comment by stggh** [comment]
```
[etalon:/home/stg121]$ python !$
python bm-mat.py
n = 1001, repeat = 100
        Z       N       J       X       Y      Xc      Yc   
Z       2.66    2.30    2.32    2.32    2.31    2.32    2.30
N       2.29    2.30    2.32    4.65   13.62    2.14    1.79
J       1.98    1.79    1.79    4.41   13.87    2.34    2.36
X       2.32    3.04    3.07    2.41    3.40    2.36    2.76
Y       2.32    8.02    7.65    4.44   17.49    2.81    7.89
Xc      2.14    2.15    2.14    2.19    2.55    2.31    2.37
Yc      2.35    2.36    2.35    3.75   10.89    2.34    2.33
inv            19.86           22.51   49.05   16.89   30.27
[etalon:/home/stg121]$ !vi
vi bm-mat.py
[etalon:/home/stg121]$ !py
python bm-mat.py
with daz ---
n = 1001, repeat = 100
        Z       N       J       X       Y      Xc      Yc   
Z       2.11    2.13    2.13    2.13    2.13    2.14    2.14
N       2.13    2.13    2.15    2.40    2.83    2.38    2.35
J       2.33    2.35    2.34    2.35    2.42    2.38    2.33
X       2.35    2.35    2.33    2.32    2.33    2.34    2.33
Y       2.33    2.33    2.35    2.33    2.35    2.35    2.35
Xc      2.31    2.33    2.32    2.34    2.30    2.30    2.34
Yc      2.32    2.31    2.18    2.20    2.15    2.16    2.15
inv            19.13           16.13   18.70   19.75   18.51
```

My system:
```
[etalon:/home/stg121]$ lscpu
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              8
On-line CPU(s) list: 0-7
Thread(s) per core:  2
Core(s) per socket:  4
Socket(s):           1
NUMA node(s):        1
Vendor ID:           GenuineIntel
CPU family:          6
Model:               94
Model name:          Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
Stepping:            3
CPU MHz:             3451.112
CPU max MHz:         4000.0000
CPU min MHz:         800.0000
BogoMIPS:            6816.00
Virtualisation:      VT-x
L1d cache:           32K
L1i cache:           32K
L2 cache:            256K
L3 cache:            8192K
NUMA node0 CPU(s):   0-7
```

**Comment by stggh** [comment]
Screen Shot 2019-03-12 at 21 36 27

```
[SteveMBP:~]$ python !$
python bm-mat.py
n = 1001, repeat = 100
        Z       N       J       X       Y      Xc      Yc   
Z       1.48    1.43    1.51    1.45    1.45    1.47    1.41
N       1.43    1.58    1.52    3.77   11.71    1.37    1.35
J       1.40    1.43    1.42    3.79   11.65    1.35    1.35
X       1.41    2.20    2.19    1.58    2.64    1.42    1.82
Y       1.41    7.00    6.81    3.82   16.95    2.23    7.30
Xc      1.40    1.45    1.44    1.51    2.03    1.43    1.45
Yc      1.42    1.45    1.43    2.92    9.92    1.36    1.35
inv             3.06            4.02   26.80    3.40   11.69
```

With daz:
```
[SteveMBP:~]$ !py
python bm-mat.py
with daz ---
n = 1001, repeat = 100
        Z       N       J       X       Y      Xc      Yc   
Z       1.69    1.70    1.79    1.82    1.46    1.59    1.53
N       1.63    1.65    1.58    1.74    1.64    1.70    1.62
J       1.52    1.48    1.59    1.43    1.41    1.42    1.41
X       1.42    1.38    1.37    1.35    1.36    1.50    1.68
Y       1.59    1.54    1.48    1.44    1.45    1.45    1.51
Xc      1.58    1.55    1.57    1.58    1.46    1.69    1.48
Yc      1.52    1.46    1.47    1.47    1.49    1.52    1.45
inv             3.17            2.96    2.49    2.92    2.50
```
`PyAbel`:
```
[SteveMBP:~]$ ipython
Python 3.7.1 (default, Oct 23 2018, 14:07:42) 
IPython 7.0.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import abel                                                                                                                              

In [2]: abel.__path__                                                                                                                            
Out[2]: ['/Users/stg121/Software/Python/VMI/PyAbel/PyAbel-MR-new_benchmark/abel']

In [3]: abel.benchmark.AbelTiming(n=[513, 1025, 2049], select=['basex'], verbose=False)                                                          
Out[3]: 
PyAbel benchmark run on i386

time in milliseconds

=== Basis generation ==============================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                 944.5      5883.9     32254.5


=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   0.8        23.0       156.3
basex(var.reg.)         5.8        56.6       301.8


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   0.8         5.8        33.9
basex(var.reg.)         5.2        25.8       153.1
```
with `daz`:
```

In [4]: import daz                                                                                                                               

In [5]: daz.set_ftz()                                                                                                                            

In [6]: daz.set_daz()                                                                                                                            

In [7]: abel.benchmark.AbelTiming(n=[513, 1025, 2049], select=['basex'], verbose=False)                                                          
Out[7]: 
PyAbel benchmark run on i386

time in milliseconds

=== Basis generation ==============================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                 928.2      5814.1     31496.5


=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   0.8         6.1        44.5
basex(var.reg.)         4.1        22.6       143.6


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   0.9         6.1        44.5
basex(var.reg.)         4.1        21.1       151.6
```

**Comment by MikhailRyazanov** [comment]
@stggh, the MBP results look much more reasonable — notice that "with daz", inversions are _less than twice_ slower than multiplications, as expected. And both `basex` and `basex(var.reg.)` behave well.
Does it also use MKL? (And Anaconda?)

**Comment by stggh** [comment]
> Does it also use MKL? (And Anaconda?)

Yes, identical installation.

> whether it can affect onion_bordas, since on my AMD it is much faster than direct_Python

No effect for `onion_bordas`:

```
In [1]: import abel                                                             

In [2]: abel.benchmark.AbelTiming(n=[513, 1025, 2049], select=['basex', 'onion_b
   ...: ordas', 'direct_Python'], verbose=False)                                
Out[2]: 

time in milliseconds

=== Basis generation ==============================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                2393.6     15323.4     76328.3


=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   9.7        41.5       247.7
basex(var.reg.)       328.9       354.9       633.5
direct_Python         392.1      7914.8     42111.8


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                  14.5        13.5        66.5
basex(var.reg.)       370.7       380.5       485.6
direct_Python         572.8      8005.3     40349.3
onion_bordas          829.4      3502.0     24776.2

In [3]: import daz                                                              

In [4]: daz.set_ftz()                                                           

In [5]: daz.set_daz()                                                           

In [6]: abel.benchmark.AbelTiming(n=[513, 1025, 2049], select=['basex', 'onion_b
   ...: ordas', 'direct_Python'], verbose=False)                                

Out[6]: 

time in milliseconds

=== Basis generation ==============================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                2094.9     15359.1     72516.8


=== Forward Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                   3.1        17.0        63.4
basex(var.reg.)       306.6       190.0       546.7
direct_Python         434.5      6850.7     40703.0


=== Inverse Abel transform ========================

Method              n = 513    n = 1025    n = 2049
---------------------------------------------------
basex                  13.0        20.9        56.8
basex(var.reg.)       190.3       150.0       421.2
direct_Python         543.2      7088.2     42756.9
onion_bordas          800.4      3662.2     25534.9
```

Quite an amazing discovery.

**Comment by DanHickstein** [comment]
Interesting stuff!

I think that we can merge this PR though, right? And open a new issue for this discussion of the subnormals?

In any case, here are some of the bm-mat.py benchmarks from my system. I have both the anaconda python (which has mkl) and the regular python that comes with my mac (/usr/bin/python, Steve, you probably have this too) which uses Apple's Accelerate Framework for math. 

Here is the output with anaconda/MKL:
```
n = 501, repeat = 1000
        Z       N       J       X       Y      Xc      Yc
Z       4.84    5.08    5.33    7.12    7.22    7.80    7.50
N       5.85    7.70    6.07   15.09    6.54    6.28    5.89
J       5.29    5.52    5.32   14.68    5.44    5.77    6.88
X       5.56   13.84   12.94    6.79   13.21    6.16   13.06
Y       5.51    5.22    5.08   14.91   14.69    7.78   13.11
Xc      6.08    7.06    6.19    7.33    7.40    4.96    5.44
Yc      6.13    5.58    4.99   14.31    8.16    5.51    5.96
inv            12.82           24.71   17.33   12.36   10.69
```

Here it is with anaconda/MKL and daz (inverses are ~2x faster):
```
n = 501, repeat = 1000
        Z       N       J       X       Y      Xc      Yc
Z       4.36    4.31    5.18    6.02    5.30    4.27    4.39
N       5.73    6.71    5.30    4.84    4.72    4.78    4.43
J       4.58    6.14    4.08    4.94    4.58    6.13    8.19
X       6.10    5.20    5.11    5.16    5.76    4.27    5.10
Y       4.60    4.51    4.30    4.19    4.24    4.44    4.31
Xc      4.05    4.02    3.90    4.29    4.03    3.95    3.93
Yc      4.06    4.34    4.12    4.07    4.11    4.09    4.21
inv            10.25            9.67    8.96    9.59    8.98
```

Apple's Accelerate Framework (/usr/bin/python), fairly competitive with MKL, and faster inverses than MKL without daz:
```
n = 501, repeat = 1000
        Z       N       J       X       Y      Xc      Yc
Z       5.48    5.12    5.34    5.19    5.31    5.23    6.19
N       5.65    5.94    6.66   14.75    7.15    5.99    5.79
J       5.26    5.40    5.42   14.06    5.95    6.90    5.96
X       5.22   15.74   14.50    6.74   13.57    8.07   13.21
Y       5.29    5.62    5.44   13.35   13.35    5.48    8.07
Xc      4.92    5.07    5.29    5.53    5.22    5.89    5.80
Yc      5.48    6.00    5.76   13.37    7.19    6.10    5.62
inv            10.46           18.55   12.95    8.98    9.06
```

Apple's Accelerate Framework with daz (daz doesn't seem to help):
```
n = 501, repeat = 1000
        Z       N       J       X       Y      Xc      Yc
Z       5.29    5.20    5.15    5.52    5.29    5.21    5.20
N       8.01    6.58    6.82   16.53    5.48    5.38    7.22
J       6.07    6.36    6.20   15.67    6.00    5.65    6.19
X       6.11   13.88   14.79    7.08   12.68    6.59   16.08
Y       6.52    6.45    7.19   14.46   14.08   10.84   10.26
Xc      7.64    5.62    5.26    7.36    6.31    7.49    8.11
Yc      6.64    5.70    5.29   12.98    9.95    7.44    6.60
inv            11.19           26.24   19.53   10.42   10.28
```

**Comment by stggh** [comment]
> /usr/bin/python, Steve, you probably have this too

Yes, but I use Anaconda python, and AFAICT the Mac version is python2.7

Subnormals appears to have an important effect on real images, even for the inverse Abel transform:
```python
In [1]: import numpy as np                                                   

In [2]: import abel                                                          

In [3]: im = np.loadtxt('O2-ANU4096.txt.bz2')                                

In [4]: imc = abel.tools.center.center_image(im, center='slice')             

In [5]: q = abel.tools.symmetry.get_image_quadrants(imc)                     

In [6]: %timeit abel.basex.basex_transform(q[0], verbose=False)              
442 ms ± 1.95 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [7]: import daz                                                           

In [8]: daz.set_ftz()                                                        

In [9]: daz.set_daz()                                                        

In [10]: %timeit abel.basex.basex_transform(q[0], verbose=False)             
114 ms ± 5.98 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [11]: daz.unset_daz()                                                     

In [12]: daz.unset_ftz()                                                     

In [13]: %timeit abel.basex.basex_transform(q[0], verbose=False)             
473 ms ± 19.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [14]: daz.set_ftz()                                                       

In [15]: daz.set_daz()                                                       

In [16]: %timeit abel.basex.basex_transform(q[0], verbose=False)             
110 ms ± 1.41 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

MikhailRyazanov:
OK, I am then going to merge this as is.
There is an outstanding task of using user-supplied images, but I think that such benchmarks would be quite different and thus should be done separately (and perhaps after reworking `linbasex` to accept half-images).

To continue the discussion, I'll open separate issues about subnormals and other stuff. Is there any way to transfer the relevant comments from here to there? (If not, I'll just add links.)

DanHickstein:
Awesome! Thanks for merging this. So much better!

And no, I don't think that there is a way to transfer comments, except by copy+paste.

PR #248, opened 2019-03-18
Fixed shift in onion-bordas

DanHickstein:
In the Onion-Bordas method, we shift the image by half a pixel to align our 0,1,2,3 grid with what the transform expects (0.5, 1.5, 2.5,...). When we shifted back, we didn't include `mode='nearest`, which resulted in the x=0 value getting set to zero. In fact, we would rather just set this value to the nearest value.

There is still a 1.5 pixel shift needed in the code (was previously 1 pixel, but 1.5 provides better agreement) that I don't know the origin of. This requires more attention. 

But, for now, we should merge this, since it at least fixes the first-pixel issue.

### DIFF
--- a/abel/onion_bordas.py
+++ b/abel/onion_bordas.py
@@ -109,13 +109,13 @@ def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=True,
     if direction != 'inverse':
         raise ValueError('Forward "onion_bordas" transform not implemented')
 
+    # make sure that the data is the right shape (1D must be converted to 2D):
+    IM = np.atleast_2d(IM.copy())
+    
     # onion-peeling uses grid rather than pixel values, 
     # odd shaped whole images require shift image (-1/2, -1/2)
     if shift_grid:
-        IM = shift(IM, -0.5, mode='nearest')
-
-    # make sure that the data is the right shape (1D must be converted to 2D):
-    IM = np.atleast_2d(IM.copy())
+        IM = shift(IM, (0,-0.5,) mode='nearest')
 
     # we would like to work from the outside to the inside of the image, 
     # so flip the image to put the "outside" at low index values.
@@ -161,6 +161,6 @@ def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=True,
 
     # shift back to pixel grid
     if shift_grid:
-        abel_arr = shift(abel_arr, 0.5, mode='nearest')
+        abel_arr = shift(abel_arr, (0, 0.5), mode='nearest')
         
     return abel_arr/(2*dr)  # x1/2 for 'correct' normalization   
--- a/abel/onion_bordas.py
+++ b/abel/onion_bordas.py
@@ -158,7 +158,7 @@ def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=True,
 
     # shift back to pixel grid
     if shift_grid:
-        abel_arr = shift(abel_arr, (0, 0.5), mode='nearest')
+        abel_arr = shift(abel_arr, (0, 1.0), mode='nearest')
     
     if abel_arr.shape[0] == 1:
         # flatten array
--- a/abel/onion_bordas.py
+++ b/abel/onion_bordas.py
@@ -109,7 +109,7 @@ def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=False,
     # onion-peeling uses grid rather than pixel values, 
     # odd shaped whole images require shift image (-1/2, -1/2)
     if shift_grid:
-        IM = shift(IM, -1/2)
+        IM = shift(IM, -0.5)
 
     # make sure that the data is the right shape (1D must be converted to 2D):
     IM = np.atleast_2d(IM.copy())
@@ -150,7 +150,7 @@ def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=False,
         abel_arr[:, col_index] = normfac * rest_col * vect.transpose()
 
     # set missing 1st column
-    abel_arr[:, 0] = abel_arr[:, 1]
+    # abel_arr[:, 0] = abel_arr[:, 1]
 
     # for some reason shift by 1 pixel aligns better? - FIX ME!
     # Just like hansenlaw
@@ -164,6 +164,6 @@ def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=False,
 
     # shift back to pixel grid
     if shift_grid:
-        abel_arr = shift(abel_arr, 1/2)
+        abel_arr = shift(abel_arr, 0.5, mode='nearest')
 
     return abel_arr/(2*dr)  # x1/2 for 'correct' normalization   
--- a/abel/onion_bordas.py
+++ b/abel/onion_bordas.py
@@ -57,7 +57,7 @@ def _init_abel(xc, yc):
     return val1, val2
 
 
-def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=False,
+def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=True,
                            **kwargs):
     r"""Onion peeling (or back projection) inverse Abel transform.
 
@@ -74,7 +74,10 @@ def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=False,
     This function operates on the "right side" of an image. i.e. it works on 
     just half of a cylindrically symmetric image.  Unlike the other transforms,
     the left edge should be the image center, not mid-first pixel. This 
-    corresponds to an even-width full image. If not, set `shift_grid=True`. 
+    corresponds to an even-width full image. 
+                           
+    However, shift_grid=True (default) provides the typical behavior,
+    where the first pixel corresponds to the center pixel of the image.
 
     To perform a onion-peeling transorm on a whole image, use ::
     
@@ -109,7 +112,7 @@ def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=False,
     # onion-peeling uses grid rather than pixel values, 
     # odd shaped whole images require shift image (-1/2, -1/2)
     if shift_grid:
-        IM = shift(IM, -0.5)
+        IM = shift(IM, -0.5, mode='nearest')
 
     # make sure that the data is the right shape (1D must be converted to 2D):
     IM = np.atleast_2d(IM.copy())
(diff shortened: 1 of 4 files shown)


stggh:
Transform pair profiles 1-6 (excluding profile 4), %difference curves, with MSE (mean-square-error):
old:
TransformPairs-inverse-diff-orig

@DanHickstein "fix" (as above):
TransformPairs-inverse-diff

There is an improvement in `mse`.

The `1.5` pixel shift provides the main `r > 0` improvement. There is still a smaller issue near `r=0`.
old:
Screenshot-20190319153054-563x419

New:
Screenshot-20190319144414-586x430

The transform pair results suggest the normalization for `onion_bordas` is correct. 

For a real VMI (O2-ANU####.txt) the relative amplitude of the `onion_bordas` image improves with image size.
`1024x1024`:
Screenshot-20190319152641-589x420
cf `3072x3072`:
Screenshot-20190319152356-589x421

`onion_bordas` appears more sensitive to the image grid size (`dr`).

I think the normalization for `onion_bordas` is correct (no doubt we checked this in the past). This PR improves the `mse` for the transform-pair functions. As good as it gets for `onion_bordas` until there is closer scrutiny of the algorithm code.

abel/onion_bordas.py
```diff
@@ -109,7 +112,7 @@ def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=False,
     # onion-peeling uses grid rather than pixel values, 
     # odd shaped whole images require shift image (-1/2, -1/2)
     if shift_grid:
-        IM = shift(IM, -1/2)
+        IM = shift(IM, -0.5, mode='nearest')
```
stggh: `__future__.division` ensures `0.5`, roll on EOL python2.7. Personal preference for decimals, no worries. ;-)

abel/onion_bordas.py
```diff
@@ -57,7 +57,7 @@ def _init_abel(xc, yc):
     return val1, val2
 
 
-def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=False,
+def onion_bordas_transform(IM, dr=1, direction="inverse", shift_grid=True,
```
stggh: Note, this default increases the "transform" time. It might be better for the user to add a dummy column. Here is a comparison of the even pixel size (default O2-ANU1024.txt) and odd pixel, recentred analysis:
image

MikhailRyazanov:
@stggh,
> the relative amplitude of the `onion_bordas` image improves with image size.

Is it an effect of the interpolation? Visually, the integral intensity seems to be OK, but the peaks are slightly broader. And shifted by something like 0.5–1 px. Is this shift radially-dependent?

stggh:
That is a tricky question, because the plot is an angular integration (code [gist](https://gist.github.com/stggh/a032128f547f2d423c55f457aee02124))

image

*Looks* pretty uniform

MikhailRyazanov:
> _Looks_ pretty uniform

Interesting. Can you test it with a uniform comb of gaussians to have a fair comparison at different radii?

stggh:
Sorry, I had other things to do today.

The uniform comb of gaussians will need to be forward Abel transformed first?

MikhailRyazanov:
Something like [this](https://gist.github.com/MikhailRyazanov/a8c7d3264aacab8860848e5e0aa41ab5) (Gaussians would be better due to their smoothness, but this is analytical):
f1
f2
`shift_grid=True` seems to work correctly, but the central pixel is apparently lost in any case.

abel/onion_bordas.py
DanHickstein: Yeah, living in the Py2 universe for the past 10 years has given me a deeply-rooted fear of integer division :)

abel/onion_bordas.py
DanHickstein: For performance, it probably is better for the user to add a column. But, since onion-bordas isn't one of the faster methods anyway, perhaps consistency with other transform methods is more important than performance?

DanHickstein:
Oh, so there is a one-pixel shift when `shift_grid=False`?

There was a line that I deleted that shifted the results by one pixel. Probably that line should be returned.

stggh:
Good point, I wondered why the `onion_bordas` shift was not so obvious in Fig. 4(c) of the PyAbelPaper.

stggh:
> Something like this 

Neat coding @MikhailRyazanov!

> shift_grid=True seems to work correctly

Good enough, but not perfect:
Screenshot-20190321090640-640x535

I am not sure how to change `dr`, altered a few other parameters:
```python
hw = 2  # peak half-width
step = 20  # center-to-center distance between peaks
n = 5  # number of peaks

rmax = int(3*n * step)

   :
subr = np.logical_and(comb.r >= 35, comb.r <= 45)
print(f'area Bordas = {simps(recT[subr], comb.r[subr]):5.3f},'
      f' area actual = {simps(comb.func[subr], comb.r[subr]):5.3f}')
print(f'Bordas CoM = {center_of_mass(recT[subr])[0]:5.3f},'
      f' actual CoM = {center_of_mass(comb.func[subr])[0]:5.3f}')
```
```
area Bordas = 1.988, area actual = 2.000
Bordas CoM = 5.087, actual CoM = 5.000
```

DanHickstein:
I modified the code to reinstate the 1 pixel shift for both True and False values of shift_grid. I think that the results are now more reasonable:
image

MikhailRyazanov:
After f60f22b5e305d2f7024cab2cd3bea98663d94fad seems to work well for the experimental image also (very small shift, < 1 pixel):
expt

But the central pixel (_r_ = 0) is still lost somewhere:
Figure_1

DanHickstein:
Should we go ahead and merge this then?

MikhailRyazanov:
Probably yes (but please do not forget to rebase).
The central pixel can be investigated later (maybe we need to open an issue as a reminder).

DanHickstein:
> but please do not forget to rebase

Uh, so, I basically don't understand the motivation for rebasing, nor how to do it easily. But, I know that whenever I submit a PR for another repo, people ask me to do it. What is the deal? It basically just makes the revision history cleaner?

MikhailRyazanov:
I actually do not know. If it will not rewrite the previous changes (“This branch is 4 commits ahead, **13 commits behind** PyAbel:master.”), then it might be fine to merge as is. You can try, and if this does not work, then revert and rebase/merge again.

DanHickstein:
Ah, okay, I see what you are saying. I can easily just make a commit to merge the latest changes into this branch. This is easier than actually rebasing, which you can use to change numerous commit messages into a single commit message. True rebasing has always been a huge hassle for me. 

This branch is now "5 commits ahead" of Pyabel:master.

stggh:
The image shift to create an odd-image should only be applied to the columns:
```python
    # onion-peeling uses grid rather than pixel values, 
    # odd shaped whole images require shift image (0, -1/2)
    if shift_grid:
         IM = shift(IM, (0, -0.5),  mode='nearest')
```
 
See [onion_bordas.py](https://gist.github.com/stggh/a5eefcf93e23155b43f3c3a7fc7db533) for the full version. It is necessary to call `IM = np.atleast_2d(IM.copy())` before the single axis shift. Similarly, for the return to an even image.

DanHickstein:
Woah! Yeah, thanks for catching that. We should certainly only be shifting in the horizontal and not the  vertical. I'm kinda surprised that `scipy.ndimage.shift` even allows a single value to be entered.

DanHickstein:
Okay, should I go ahead and merge this?

stggh:
> only be shifting in the horizontal

This is not as obvious as it might first appear, because in principle the image centre is moved from centre-of-the-centre-pixel to the pixel-grid. However, empirically, for the O<sub>2</sub> velocity-map image, only the column shift yields the correct photoelectron spectrum, consistent with the other `PyAbel` methods.

Yes, good to merge.

stggh:
I think the `shift(image, -0.5)` should remain. The pseudo resolution change improvement for a column-only shift, appears to have been due to `shift=False`, i.e. the grid shift not being applied. The effect of a grid shift of `(-0.5, -0.5)` vs `(0, -0.5)` on the O<sub>2</sub> peak resolution is miniscule. I believe that `(-0.5, -0.5)` is correct.

@DanHickstein please revert the `(0, -0.5)` shift change. Sorry, about this, to much going on at the moment.

DanHickstein:
> The effect of a grid shift of (-0.5, -0.5) vs (0, -0.5) on the O2 peak resolution is miniscule. I believe that (-0.5, -0.5) is correct.

Why do you expect that shifting in the vertical is correct? Intuitively, there is not reason that the transform should be having any issue with the rows. Shifting rows should cause clear issues with the transform on a pixel level.

Here is a comparison. I've made a synthetic image using our "comb" of several sharp peaks. I've then made it into a 2D image by putting rows of zeros on the top and bottom. ([Gist link](https://gist.github.com/DanHickstein/3eb197236afa3b15a4c4e28278321b27))

The left side of this figure shows the projection, and the right side is the inverse Abel transform using the new ((0, -0.5)) shifted onion_bordas. Everything looks fine:

image

If I revert back to the ((-0.5, -0.5)) behavior, clear artifacts are seen, as a result of the shift:

image

So, I'm fairly sure that the vertical shift should be zero.

DanHickstein:
Hmmm, so another issue is that a mysterious half-pixel shift is required. Shifting the image back by (0, 0.5) produces fairly good results, but shifting by (0, 1.0) produces even better agreement with the gaussian test (0.81% error rather than 1.45%) and the transform of the "comb" also matches better.

In the absence of really understanding what's going on, I'm tempted to make the heuristic correction that improves agreement with our test datasets.

stggh:
> Why do you expect that shifting in the vertical is correct?

Because of the definition of even vs odd pixel size image centre https://github.com/PyAbel/PyAbel/issues/39#issue-119736937

1c7e06f4-9840-11e5-9791-76d18db76ca5

stggh:
I think the initial shift, for an odd-size image, should be `(-1/2, -1/2)` to align the centre to that for an even-size image, as required for `onion_bordas`. Then a post inverse Abel shift `(+1/2, +1/2)` to return to an odd-size image. Plus `(0, #)` to correct any grid misalignment of the `onion_bordas` implementation.  I am still testing this.

stggh:
The `onion_bordas` transform is unaffected by any row shift, so it might as well be zero `(0, dy)`. My own O<sup>-</sup> test spectrum [gist](https://gist.github.com/stggh/bceb01f1d1ba3a492728046b641dce62) gives a return shift of `(0, 0.9)` i.e. `+0.4` rather than the `+0.5` that you obtained, but `+0.5` may be close enough.
onion-y

DanHickstein:
I found that the 0.9 shift provided better agreement with a sharp peak (the "comb" example) but worse agreement with the gaussian.

DanHickstein:
So, it's clear that we still don't perfectly understand how the onion_bordas method works. At some point we need to re-write the basic code so that it operates on the 0,1,2,3 grid instead of the 0.5, 1.5, 2.5 grid so that it can stand on a more even footing with the other PyAbel methods (and not need a shift). 

Still, this PR makes the method way better than it was before. I suggest that we merge this and tackle a better re-write of onion_bordas later. Sound good? If @stggh or @MikhailRyazanov agree, please hit the green button :)

MikhailRyazanov:
As I understand, this version was used for the article figures? Then we should definitely merge it, even if it still has some unresolved questions.

DanHickstein:
Yes, this version is needed to fix the bug at r=0, so it was used for the article figures.

stggh:
Yes, merge.  Thanks, @DanHickstein

DanHickstein:
Merging... :)

PR #249, opened 2019-04-18
hansenlaw with gradient calculated as a first-order finite difference

stggh:
For `hansenlaw` evaluate the image row gradient as a first-order finite difference, as per @MikhailRyazanov suggestion:
> Well, in one dimension, “gradient” is the derivative. But I see that PyAbel actually calculates it using numpy.gradient, which assumes “that f ... has at least 3 continuous derivatives” — I think, this is way too much to expect from any experimental data.

> Basically, the problem is that NumPy uses the two-sided formula to estimate the derivative at the grid points, whereas our integral (2) actually needs the derivatives within the integration intervals. I suspect that using the standard first-order finite differences and choosing the integration points half-way between the image sample points will work better. For the “direct” method at least, I don't know how the Hansel–Law method works.

Bonus, the `sub_pixel_shift=-0.35` grid-alignment (for `hold_order=0`) is no longer required, and has been dropped.

### DIFF
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -22,6 +22,8 @@ from scipy.ndimage import interpolation
 #      J. Opt. Soc. Am A2, 510-520 (1985)
 #      doi: 10.1364/JOSAA.2.000510
 #
+# 2019-04   : replace gradient with first-order finite difference as
+#             per @MikhailRyazanov suggestion
 # 2018-04   : New code rewrite, implementing the 1st-order hold approx. of
 #             Ref. [1], with the assistance of Eric Hansen. See PR #211.
 #
@@ -53,7 +55,7 @@ from scipy.ndimage import interpolation
 
 
 def hansenlaw_transform(image, dr=1, direction='inverse', hold_order=0,
-                        sub_pixel_shift=-0.35, **kwargs):
+                        **kwargs):
     r"""Forward/Inverse Abel transformation using the algorithm of:
 
     `E. W. Hansen "Fast Hankel Transform" IEEE Trans. Acoust. Speech Signal
@@ -122,14 +124,6 @@ def hansenlaw_transform(image, dr=1, direction='inverse', hold_order=0,
         yield a more accurate transform for some functions (see PR 211).
         Default: `0`.
 
-    sub_pixel_shift : float 
-        For the zero-order hold approximation `hold_order=0`, a sub-pixel
-        left-shift of the driving function (image-forward or gradient-inverse)
-        improves the transform alignment with the other PyAbel methods,
-        and Abel transform-pair functions.  See the discussion in final summary
-        of `PR #211 <https://github.com/PyAbel/PyAbel/pull/211>`_ 
-        Default: -0.35.
-
     Returns
     -------
     aim : 1D or 2D numpy array
@@ -168,7 +162,8 @@ def hansenlaw_transform(image, dr=1, direction='inverse', hold_order=0,
         drive = -2*dr*np.pi*np.copy(image)
         a = 1  # integration increases lambda + 1
     else:  # inverse Abel transform
-        drive = np.gradient(image, dr, axis=-1)
+        drive = np.zeros_like(image)
+        drive[:, :-1] = (image[:, 1:] - image[:, :-1])/dr
         a = 0  # due to 1/piR factor
 
     n = np.arange(cols-1, 1, -1)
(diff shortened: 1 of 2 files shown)


DanHickstein:
This looks cool! I'm still catching up on your other discussion with @MikhailRyazanov, but can you post a before/after comparison figure here and/or elaborate a bit on how this helps?

stggh:
The `hansenlaw` inverse Abel transform operates on the gradient image, and this was implemented using `drive = numpy.gradient(image, dr, axis=-1)`. My interpretation of @MikhailRyazanov `numpy.gradient` comment is that the derivative calculation requires more than just adjacent points, and so fails if there is any sharp structure. I agree with his comment:

> using the standard first-order finite differences ... will work better

This has been implemented by replacing `np.gradient()` with a sliced difference: `drive[:, :-1] = (image[:, 1:] - image[:, :-1])/dr`

Test case: We have been investigating the inverse Abel transform of a comb of 1-pixel-width peaks (with noise). Here are the before and after images, comparing the `hansenlaw` inverse Abel transform with the **noiseless** source function (blue). Note the improved agreement with the first-order difference method (*after*):
hl-before
hl-after

This change also resolves a long standing issue concerning image alignment of the `hansenlaw` inverse Abel image, that required a post transform image shift of `-0.35` pixels, which is now, no longer required. It seems the `numpy.gradient()` does some funny business related to  

> two-sided formula to estimate the derivative at the grid points

vs

> needs the derivatives within the integration intervals

MikhailRyazanov:
> `numpy.gradient` ... the derivative calculation requires more than just adjacent points, and so fails if there is any sharp structure.

Yes, as I understand from its documentation, it does
```
grad[i] = (f[i+1] - f[i-1]) / (2 * dr)
```
instead of
```
grad[i] = (f[i+1] - f[i]) / dr
```
or
```
grad[i] = (f[i] - f[i-1]) / dr
```
So basically returns the derivative averaged over the two adjacent intervals and needs 3 points (actually does not use the central one, but needs this 3-point span).

You can also try whether `numpy.ediff1d` or `numpy.diff` will work faster than Python slices.

stggh:
> interesting to look at the Gaussian results “before” and “after”

Here are the transform pairs [code gist](https://gist.github.com/stggh/08fee13b7f2852a7b6bb24596da1e499), before (*gradient*) and after (*1st-order-difference*):

1st-order difference produces a slight improvement, **except** with `hold_order=1`, which prefers `np.gradient()`. 

1) *gradient*
 ```
gradient
profile  hansenlaw   basex    3pt
1        3.92e-06  5.21e-08  3.08e-08
2        3.01e-06  6.50e-08  5.82e-09
3        3.22e-06  7.00e-08  5.68e-09
5        5.02e-04  1.00e-02  4.93e-03
6        5.25e-06  6.03e-08  6.76e-08
7        5.44e-06  7.79e-08  4.35e-08
```
TP-gradient

2) *1st-order-difference*
```
1st-order-diff
profile  hansenlaw   basex    3pt
1        1.31e-06  5.21e-08  3.08e-08
2        1.16e-06  6.50e-08  5.82e-09
3        1.14e-06  7.00e-08  5.68e-09
5        1.32e-03  1.00e-02  4.93e-03
6        3.06e-06  6.03e-08  6.76e-08
7        3.42e-06  7.79e-08  4.35e-08
```
TP-1st-order-diff

@MikhailRyazanov note there is something odd with `basex` for `profile5`

For the Gaussian function, the agreement slightly degrades, from `0.23%` to `0.33%`:
gaussian-grad
gaussian-1st


Given the large improvement for the comb structure, and the excution time improvement(*) from dropping `interpolation.shift`,  I am in favour of merging this PR (after the `onion_bordas.py` PR)

(*) *gradient*
In [6]: %timeit abel.Transform(imc, method='hansenlaw')                         
275 ms ± 10.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

*1st-order-difference*
In [6]: %timeit abel.Transform(imc, method='hansenlaw')                         
157 ms ± 841 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

MikhailRyazanov:
So I would say that `hansenlaw` now produces results almost identical to `two_point`, except _r_ = 0 (I have noticed this in your [comb figure](https://user-images.githubusercontent.com/10932229/56333459-9f4d0a80-61d7-11e9-8c64-dfbacf845ad0.png)). Which is interesting, since if this holds for large images as well, then `hansenlaw` will be the fastest (and among the highest-resolution) method for them. Especially after becoming more computationally efficient. I'm not sure that it can actually beat `two_point` and reach the “HD video” goal, but it should get very close.

(I'll take a look at what happens with `basex`.)

DanHickstein:
This seems to work well. Should we go ahead and merge?

PR #252, opened 2019-04-26
New benchmark code (update)

MikhailRyazanov:
1. For convenience, multiple image sizes (`n=...`) can now be passed as `numpy.ndarray` (or any other object convertible to Python's `list`). The previous implementation required multiple sizes to be passed as a list and was producing not very helpful error messages otherwise.
2. Some people try to use `abel.benchmark.AbelTiming` for very large image sizes, and if it runs out of memory, the program crashes, losing all the previous results. To avoid that, the `MemoryError` exception is now handled in the same way as `SIGINT` (Ctrl+C), skipping the rest of the benchmarks, but preserving the existing results.

I have only tested that this memory handling works in Linux when the requested size is too large.
If anybody is willing to test it in Windows and macOS, and also how it works (in all systems) in cases when the requested memory is close to the limit (for example, less than the RAM size but more than the free RAM, and more than the physical RAM but less than the RAM + swap), please report here.

### DIFF
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -21,9 +21,9 @@ import sys
 
 def _ensure_list(x):
     """
-    Wrap the argument in a list, if it is not a list already.
+    Convert the argument to a list (a scalar becomes a single-element list).
     """
-    return x if isinstance(x, list) else [x]
+    return [x] if np.ndim(x) == 0 else list(x)
 
 
 def _roundsf(x, n):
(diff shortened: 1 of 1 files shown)


MikhailRyazanov:
I nobody has objections, I'm going to merge this...

DanHickstein:
Sounds good! Thanks!

stggh:
Yes, good work. Thanks.

PR #253, opened 2019-05-01
abel.tools.vmi.angular_integration() Jacobian should be R^2 not R

stggh:
Trivial bug, the Jacobian, line 66, in `abel.tools.vmi.angular_integration()` should be <math>R<sup>2</sup>sin&theta;</math> not <math>Rsin&theta;</math>.

Tracked down from some recent velocity-map imaging data that has intensity near the image centre. The `xR` resultant photoelectron spectrum is way too strong.

<math>R</math>   vs <math>R<sup>2</sup></math>:

imageimage

### DIFF
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -62,8 +62,8 @@ def angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None):
 
     dt = T[0, 1] - T[0, 0]
 
-    if Jacobian:  # x r sinθ
-        polarIM = polarIM * R * np.abs(np.sin(T))
+    if Jacobian:  # x r^2 sinθ
+        polarIM *= R**2 * np.abs(np.sin(T))
 
     speeds = np.trapz(polarIM, axis=1, dx=dt)
 


stggh:
Hmm, this fails `test_linbasex.py` which compares the integrated radial profile with `linbasex`'s `Beta[0]`. Something amiss ...

DanHickstein:
Oh no! Not the Jacobian again! :)

Dan from three years ago would make the argument that: "The `angular_integration` function itself applies only the `r * sin(theta)` part, while the `reproject_image_into_polar` applies the extra factor of `r`." (See https://github.com/PyAbel/PyAbel/pull/172#issuecomment-211984743)

What is ANUVMI? Is that another program that you wrote? Is it possible that it is including an extra factor of `r`?

It would be nice if we could analytically work out the angular integration of a simple function so that we had complete confidence in that result. And then we can match PyAbel to that. This should probably be a unit test for the `angular_integration` function.

stggh:
I am 99% certain ;-) that the extra `R` scaling is required. Just in the process of collecting evidence to present my case.

DanHickstein:
Oh my, that is a high level of certainty :)

Between `angular_integration` and `reproject_image_into_polar`, there is, for sure, a factor of r^2 already, right? Maybe there is an effective divide by `r` somewhere...

stggh:
Closing this until I have time to prove my case (or not). Proving the Jacobian one way or the other is not as easy/straightforward as I thought.

PR #256, opened 2019-06-10
Bump version to v0.8.2

DanHickstein:
Incrementing the version number so that we can release 0.8.2. 

Also, included a link to our arxiv paper in the readme.


MikhailRyazanov:
We should have the RSI article (with better figures than in arXiv) out next week. Do you want to include it later, or this can wait a little bit to have both references?

DanHickstein:
Thanks guys! Let's update the readme later to include both reference. I'm merging now so that I can release the new version.

Issue #254, opened 2019-05-22
Anisotropy distribution


MikhailRyazanov:
I did not fully read the previous “anisotropy” issues (#60 and #176), but from a more recent [discussion](https://github.com/PyAbel/PyAbel/issues/225#issuecomment-422654901) with @stggh, it seems that the means to calculate the anisotropy distribution (radially dependent anisotropy parameters) are still demanded and not really available.

So I have a [proof of concept](https://gist.github.com/MikhailRyazanov/8092647cbf3a4d5efc24e5120d364fd6) implementing several approaches that seem reasonable for fast calculations. They all seem to work, with somewhat different properties:

For the simple image with separate peaks (1 quadrant):
![img](https://user-images.githubusercontent.com/12854133/58145227-956f5a80-7c0e-11e9-8ad4-5f3a0f2d31ba.png)

![aniso](https://user-images.githubusercontent.com/12854133/58145032-ee8abe80-7c0d-11e9-8765-2d52b12751cc.png)

For the Dribinski sample image:
![drib](https://user-images.githubusercontent.com/12854133/58145135-45909380-7c0e-11e9-96e1-c383f9399369.png)

The implemented methods are:
1. `nearest` bins image pixels to the nearest radial bin.
2. `linear` bins image pixels, distributing them over the 2 closest radial bins with linear weights.
3. `remap` converts the image to a uniform polar grid (like `reproject_image_into_polar`).

Except these differences, they all do linear least-squares fitting to cos<sup>_l_</sup> θ terms, then convert them to Legendre polynomial and finally to the intensity and anisotropy parameters.

Some of this stuff does not depend on the image data and can be cached, some needs to be done with the image.
* `nearest` has the fastest non-cacheable part (~50 ms on my system for 1 quadrant of a megapixel image), seems to be the least biased, but the noisiest.
* `linear` is about twice slower, less noisier, but produces somehow biased β in very-low-intensity parts.
* `remap` has the best total time, but a slightly worse throughput (non-cacheable) than `nearest`, has the lowest noise on the peaks, but behaves crazily in very-low-intensity parts.

So it all seems to work _in principle_, but needs some improvements before adding to PyAbel:
1. Now it works with 1 quadrant only. This should be extended to full and half (?) images. Maybe by automatic folding (fast, but limited), maybe by actually going through all the supplied data. What do other `polar`/`vmi` tools expect in this regard?
[...]

PR #257, opened 2019-06-24
Various radial distributions

MikhailRyazanov:
This is the implementation of various radial distributions (intensity, anisotropy), discussed in #254.
(Plus some minor changes to code and docstrings formatting and the updated copyright years.)

Works for any image parts (any quadrant, any half, full image with the origin anywhere) and any even angular orders. Also is fast:
```
order = 2, time in milliseconds

=== Half image, single ==========================================

Method                n = 257     n = 513    n = 1025    n = 2049
-----------------------------------------------------------------
nearest
   MIN, none              2.0         5.4        18.1        52.5
   MIN, sin               2.0         6.2        20.2        55.7
   MIN, sin+array         2.1         6.5        21.5        62.3
linear
   MIN, none              2.3         6.9        24.9        82.0
   MIN, sin               2.4         8.2        30.6        92.6
   MIN, sin+array         2.5         8.9        32.6        98.9
remap
   MIN, none              8.5        29.6       118.3       461.7
   MIN, sin               8.5        31.5       121.9       473.4
   MIN, sin+array         8.7        31.8       123.6       487.4


=== Half image, cached ==========================================

Method                n = 257     n = 513    n = 1025    n = 2049
-----------------------------------------------------------------
nearest
   MIN, none              0.7         1.3         3.6        12.1
   MIN, sin               0.7         1.4         4.1        14.1
   MIN, sin+array         0.8         1.7         5.1        18.3
linear
   MIN, none              0.9         2.0         6.6        24.1
   MIN, sin               0.9         2.1         7.2        26.6
   MIN, sin+array         1.0         2.3         8.1        30.5
remap
   MIN, none              4.0        14.2        59.4       222.8
   MIN, sin               4.1        14.3        58.2       225.8
   MIN, sin+array         4.2        14.6        58.0       231.5
```
One quadrant is somewhat faster, full image is somewhat slower. Timings for the radial intensity distribution only (`order = 0`) are also somewhat smaller. For comparison, here are the timings for `abel.tools.vmi.angular_integration`:
```
Shape                 n = 257     n = 513    n = 1025    n = 2049
-----------------------------------------------------------------
one quadrant              3.1        14.7        57.4       208.1
half                     15.1        57.5       213.9       839.6
full                     13.8        54.8       203.4       800.0
```

Besides asking everybody to test this, I have two design questions:
1. It seems that different tools use different conventions for the image center:
   * (row, col), as for arrays,
   * (x, y) with y from the bottom, as usual coordinates (but contrary to “PyAbel adopts the “television” convention”).

   Here I use (row, col), but maybe it should be changed?
2. The results are arranged as arrays with the first index corresponding to the radius, and the second to the angular terms. This means that, for example, instead of
   `I, beta = results.Ibeta()`
   one needs to write
   `I, beta = results.Ibeta().T`
   which is less neat. On the other hand, this allows to write
   `plt.plot(results.harmonics())`
   to plot all radial distributions together. And might be considered more convenient is some situations, since `Ibeta[r]` gives all the information (_I_(_r_), _β_<sub>2</sub>(_r_), _β_<sub>4</sub>(_r_), ...) about radius _r_.
   Or it should be the opposite?

### DIFF
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -535,7 +535,10 @@ class DistributionsTiming(object):
         ``'full'``:
             full image (*n*, *n*), centered
     rmax : str or sequence of str
-        symbolic **rmax** for :class:`abel.tools.vmi.Distributions`
+        ``'MIN'`` (default) and/or ``'all'``, see **rmax** in
+        :class:`abel.tools.vmi.Distributions`
+    order : int
+        highest order in the angular distributions. Even number ≥ 0.
     weight : str or sequence of str
         weighting to test. Use ``'all'`` (default) for all available or
         choose any combination of individual types::
@@ -573,8 +576,8 @@ class DistributionsTiming(object):
     The results can be output in a nice format by simply
     ``print(DistributionsTiming(...))``.
     """
-    def __init__(self, n=[301, 501], shape='half', rmax='MIN', weight='all',
-                 method='all', repeat=1, t_min=0.1):
+    def __init__(self, n=[301, 501], shape='half', rmax='MIN', order=2,
+                 weight='all', method='all', repeat=1, t_min=0.1):
         self.n = _ensure_list(n)
 
         if shape == 'Q':
@@ -590,6 +593,8 @@ class DistributionsTiming(object):
             raise ValueError('Incorrect shape "{}"'.format(shape))
 
         self.rmaxs = rmaxs = _ensure_list(rmax)
+
+        self.order = order
 
         weights = _ensure_list(weight)
         if 'all' in weights:
@@ -634,9 +639,10 @@ class DistributionsTiming(object):
                             w = weight
                         # single-image
                         t1 = time(Ibeta,
-                                  IM, origin, rmax, weight=w, method=method)
+                                  IM, origin, rmax, order, weight=w,
+                                  method=method)
                         # cached
-                        distr = Distributions(origin, rmax, weight=w,
+                        distr = Distributions(origin, rmax, order, weight=w,
                                               method=method)
                         distr(IM)  # trigger precalculations
 
@@ -652,7 +658,7 @@ class DistributionsTiming(object):
         import platform
 
         out = ['PyAbel benchmark run on {}\n'.format(platform.processor()),
-               'time in milliseconds']
+               'order = {}, time in milliseconds'.format(self.order)]
 
         # field widths are chosen to accommodate up to:
         #   rmax + weight = 3 leading spaces + 10 characters
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -516,6 +516,181 @@ class AbelTiming(object):
         return '\n'.join(out)
 
 
+class DistributionsTiming(object):
+    """
+    Benchmark performance of different VMI distributions implementations.
+
+    Parameters
+    ----------
+    n : int or sequence of int
+        array size(s) for the benchmark (assuming full images to be 2D square
+        arrays (*n*, *n*))
+    shape : str
+        image shape:
+
+        ``'Q'``:
+            one quadrant ((*n* + 1)/2, (*n* + 1)/2)
+        ``'half'`` (default):
+            half image (*n*, (*n* + 1)/2), vertically centered
+        ``'full'``:
+            full image (*n*, *n*), centered
+    rmax : str or sequence of str
+        symbolic **rmax** for :class:`abel.tools.vmi.Distributions`
+    weight : str or sequence of str
+        weighting to test. Use ``'all'`` (default) for all available or
+        choose any combination of individual types::
+
+            weight=[None, 'sin', 'array']
+
+    method : str or sequence of str
+        methods to benchmark. Use ``'all'`` (default) for all available or
+        choose any combination of individual methods::
+
+            method=['nearest', 'linear', 'remap']
+
+    repeat : int
+        repeat each benchmark at least this number of times to get the average
+        values
+    t_min : float
+        repeat each benchmark for at least this number of seconds to get the
+        average values
+
+    Attributes
+    -------
+    n : list of int
+        array sizes from the parameter **n**
+    results : dict of dict of dict of list of tuple of float
+        benchmark results — multi-level dictionary, in which
+        ``results[method][rmax][weight]`` is the list of timings in
+        milliseconds corresponding to array sizes in
+        :py:attr:`DistributionsTiming.n`. Each timing is a tuple (*t*:sub:`1`,
+        *t*:sub:`∞`) with *t*:sub:`1` corresponding to single-image
+        (non-cached) performance, and *t*:sub:`∞` corresponding to batch
+        (cached) performance.
+
+    Notes
+    -----
+    The results can be output in a nice format by simply
+    ``print(DistributionsTiming(...))``.
+    """
+    def __init__(self, n=[301, 501], shape='half', rmax='MIN', weight='all',
+                 method='all', repeat=1, t_min=0.1):
+        self.n = _ensure_list(n)
+
+        if shape == 'Q':
+            origin = 'll'
+            self.shape = 'One quadrant'
+        elif shape == 'half':
+            origin = 'cl'
+            self.shape = 'Half image'
+        elif shape == 'full':
+            origin = 'cc'
+            self.shape = 'Full image'
+        else:
+            raise ValueError('Incorrect shape "{}"'.format(shape))
+
+        self.rmaxs = rmaxs = _ensure_list(rmax)
+
+        weights = _ensure_list(weight)
+        if 'all' in weights:
+            weights = [None, 'sin', 'array']
+        self.weights = weights
+
+        methods = _ensure_list(method)
+        if 'all' in methods:
+            methods = ['nearest', 'linear', 'remap']
+        self.methods = methods
+
+        # create the timing function
+        time = Timent(skip=-1, repeat=repeat, duration=t_min).time
+
+        # dictionary for the results
+        self.results = {m: {r: {w: [] for w in weights}
+                                      for r in rmaxs}
+                                      for m in methods}
+
+        from abel.tools.vmi import Ibeta, Distributions
+        # make sure that everything is loaded
+        # (otherwise the 1st timing call is very slow)
+        Ibeta(np.array([[0]]))
+
+        # Loop over all image sizes
+        for ni in self.n:
+            ni = int(ni)
+
+            # create image and weight array
+            rows = (ni + 1) // 2 if shape == 'Q' else ni
+            cols = (ni + 1) // 2 if shape in ['Q', 'half'] else ni
+            IM = np.random.randn(rows, cols)
+            warray = np.random.randn(rows, cols)
+
+            # benchmark each combination of parameters
+            for method in methods:
+                for rmax in rmaxs:
+                    for weight in weights:
+                        if weight == 'array':
+                            w = warray
+                        else:
+                            w = weight
+                        # single-image
+                        t1 = time(Ibeta,
+                                  IM, origin, rmax, weight=w, method=method)
+                        # cached
+                        distr = Distributions(origin, rmax, weight=w,
+                                              method=method)
+                        distr(IM)  # trigger precalculations
+
+                        def distrIMIbeta(IM):
+                            return distr(IM).Ibeta()
+                        tn = time(distrIMIbeta,
+                                  IM)
+                        # save results
+                        self.results[method][rmax][weight].append((t1 * 1000,
+                                                                   tn * 1000))
+
+    def __repr__(self):
+        import platform
+
+        out = ['PyAbel benchmark run on {}\n'.format(platform.processor()),
+               'time in milliseconds']
+
+        # field widths are chosen to accommodate up to:
+        #   rmax + weight = 3 leading spaces + 10 characters
+        #   ni = 99999
+        #   time = 9999999.9 ms (almost 3 hours)
+        # data columns are 9 characters wide and separated by 3 spaces
+        TITLE_FORMAT = '=== ' + self.shape + ', {} ==='
+        HEADER_ROW = 'Method       ' + \
+                     ''.join(['   {:>9}'.
+                              format('n = {}'.format(ni)) for ni in self.n])
+        SEP_ROW = '-' * len(HEADER_ROW)
+        ROW1_FORMAT = '   {}, {:5}' + '   {:9.1f}' * len(self.n)
+        ROWn_FORMAT = '             ' + '   {:9.1f}' * len(self.n)
+
+        def print_benchmark(mode):
+            title = '{:=<{w}}'.format(TITLE_FORMAT.format(mode),
+                                      w=len(SEP_ROW))
+            out = ['\n' + title + '\n']
+            out += [HEADER_ROW]
+            out += [SEP_ROW]
+            for method in self.methods:
+                out += [method]
+                resm = self.results[method]
+                for rmax in self.rmaxs:
+                    resr = resm[rmax]
+                    for weight in self.weights:
+                        res = resr[weight]
+                        t = zip(*res)[0 if mode == 'single' else 1]
+                        out += [ROW1_FORMAT.format(rmax, weight, *t)]
+            return out
+
+        out += print_benchmark('single')
+        out += ['']
+        out += print_benchmark('cached')
+
+        return '\n'.join(out)
+
+
 def is_symmetric(arr, i_sym=True, j_sym=True):
     """
     Takes in an array of shape (n, m) and check if it is symmetric
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -540,10 +540,10 @@ class DistributionsTiming(object):
     order : int
         highest order in the angular distributions. Even number ≥ 0.
     weight : str or sequence of str
-        weighting to test. Use ``'all'`` (default) for all available or
-        choose any combination of individual types::
+        weighting to test. Use ``'all'`` for all available or choose any
+        combination of individual types::
 
-            weight=[None, 'sin', 'array']
+            weight=['none', 'sin', 'array', 'sin+array']
 
     method : str or sequence of str
         methods to benchmark. Use ``'all'`` (default) for all available or
@@ -577,7 +577,8 @@ class DistributionsTiming(object):
     ``print(DistributionsTiming(...))``.
     """
     def __init__(self, n=[301, 501], shape='half', rmax='MIN', order=2,
-                 weight='all', method='all', repeat=1, t_min=0.1):
+                 weight=['none', 'sin', 'sin+array'], method='all',
+                 repeat=1, t_min=0.1):
         self.n = _ensure_list(n)
 
         if shape == 'Q':
@@ -598,7 +599,7 @@ class DistributionsTiming(object):
 
         weights = _ensure_list(weight)
         if 'all' in weights:
-            weights = [None, 'sin', 'array']
+            weights = ['none', 'sin', 'array', 'sin+array']
         self.weights = weights
 
         methods = _ensure_list(method)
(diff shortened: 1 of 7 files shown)


MikhailRyazanov:
It will also need some examples. So far I have a [demo](https://gist.github.com/MikhailRyazanov/db5b18c740d64a4d0edabba6baa5e40e) of using user-supplied weights to get usable distributions from a damaged image.

stggh:
Thanks @MikhailRyazanov I will take a look at your nice work!

DanHickstein:
This looks like a huge PR, but all really good stuff!

Regarding the design decisions:

1) Which tools adopt (row, column)? With the "television" convention, I think that we are basically trying to follow (columns, row) notation, or what np.meshgrid does. For example, if we do
`X, Y = np.meshgrid([0,1,2],[0,1,2])`
We get `Y=`
```
       [[0, 0, 0],
       [1, 1, 1],
       [2, 2, 2]])
```
So, the "bottom" of the image only corresponds to y=0 if we plot using `plt.imshow(Y, origin='lower')`

I think that we should try to be consistent, and I think that the `(x, y)` format will make more sense to the novice user. But, I could be convinced either way. And as long as we document it well, it shouldn't be too much of a problem. Do you have an opinion?

2) I suppose that my first instinct would have been to make `I, beta = results.Ibeta()` work, but I don't have a strong opinion on this one. Again, as long as it's documented well, it should work for users.

MikhailRyazanov:
> This looks like a huge PR, but all really good stuff!

It is probably over-engineered, :–) but I was trying to make it more convenient than concise...

> 1. Which tools adopt (row, column)?

Actually, seems to be most:
* [abel.transform.Transform](https://pyabel.readthedocs.io/en/latest/abel.html#abel.transform.Transform.__init__) says “center (tuple or str) – If a tuple (float, float) is provided, this specifies the image center in (y,x) (row, column) format.”
* [abel.tools.center.find_center](https://pyabel.readthedocs.io/en/latest/abel.html#abel.tools.center.find_center) returns “out – coordinate of the center of the image in the (y,x) format (row, column)”.
* and some other places.

But:
* [abel.tools.polar.reproject_image_into_polar](https://pyabel.readthedocs.io/en/latest/abel.html#abel.tools.polar.reproject_image_into_polar) says that “ “origin” is a tuple of (x0, y0) relative to the bottom-left image corner”
* [abel.tools.vmi.angular_integration](https://pyabel.readthedocs.io/en/latest/abel.html#abel.tools.vmi.angular_integration) has “origin (tuple) – Image center coordinate relative to _bottom-left_ corner defaults to rows//2, cols//2”. Not clear whether it is (y, x) or (x, y).

> I think that we should try to be consistent, and I think that the (x, y) format will make more sense to the novice user. But, I could be convinced either way. And as long as we document it well, it shouldn't be too much of a problem. Do you have an opinion?

Well, I was also confused about (y, x) or (x, y) when I tried to use something. :–)

The (x, y) convention (with the television-like origin in the top-left corner) is probably more usual for the people working with images in general.

But the (row, col) convention is the standard for matrices, and if the users are going to mix our function calls with direct matrix operations, it will be more consistent. At least, it allows to write
```
origin = (row, col)
IM[origin] = ...
```
which would require more work with the (x, y) convention.

I don't know about `meshgrid` (in principle, it does have `indexing='ij'` for those who want matrix-like order).
But in any case the “bottom-left corner” is odd.

> 2. I suppose that my first instinct would have been to make `I, beta = results.Ibeta()` work.

I also think so... The question is whether this is the most common use case, or people would mostly need all terms at specific radii?
(As I remember, this [r][ang_term] ordering was partially related to NumPy's matrix-multiplication rules. But maybe the calculations can be transposed easily. Or, in any case, returning the transposed results is not a problem.)

---
By the way, the “Xenon” sample image is yours? Do you know what is the maximal process order (the number of photons) there? Or it is non-perturbative? I would like to have some multiphoton example to test/demonstrate higher-order angular distributions.

MikhailRyazanov:
Interesting... I've changes one docstring, and now AppVeyor [fails at `conda info -a`](https://ci.appveyor.com/project/DanHickstein/pyabel/builds/25542796/job/9u2csw5kl8b670an#L70) during its environment installation for Python 3. @DanHickstein, @stggh, @rth, any ideas what does it want?

stggh:
1) Can't help with the Windows build faults. I suggest try again.

2) This PR is neat and does a great job:
    image

A few thoughts:
3) The `Distribution` class instance requires two calls 
  ```python
    dist = abel.tools.vmi.Distributions()
    imdist = dist(BIM.transform) 
    rM, I, beta = imdist.rIbeta(window=5).T
  ``` 
    Could the image be passed to the constructor?

4) The 2-letter code origin position might be better using the `matplotlib` `loc` location code/string, which may be more familiar to users?
   ```
        ===============   =============
        Location String   Location Code
        ===============   =============
        'upper right'     1
        'upper left'      2
        'lower left'      3
        'lower right'     4
        'right'           5
        'center left'     6
        'center right'    7
        'lower center'    8
        'upper center'    9
        'center'          10
        ===============   =============
```

**Comment by MikhailRyazanov** [comment]
> Can't help with the Windows build faults. I suggest try again.

Well, I'll see what happens when I add more commits, but my guess is that the new version of Conda (4.7.5-py37_0, was 4.6.14-py37_0 before) fails, so this problem is likely to persist. No idea why it's only “-py37_0”, whereas “-py27_0” works fine.

I tried to search by the error message and found https://github.com/conda/conda-build/issues/3220, which is about `conda-build`, but might be related. In that case, it seems that it does not like when “active environment” is None, and the solution is to activate something beforehand. Probably makes sense for the `build`, but I would not expect `info` to crash instead of reporting `active environment : None`, as before.

The script creates a new “abel-env” environment immediately after that, but maybe calling `activate base` before `info` would help and not harm anything else? You and @DanHickstein are using Conda, as I understand. Could you please test that? (However, I do not know whether this is a Windows-only problem.)

> Could the image be passed to the constructor?

There is `abel.tools.vmi.rIbeta` for single-image analysis.

The purpose of creating a separate `Distributions` object is to keep the cached data somewhere when multiple images need to be analyzed. And then the calls can be chained:
```python
rM, I, beta = dist(BIM.transform).rIbeta(window=5).T
```

By the way, as the most active _user_, do you have an opinion on the output transposition?
`r, I, beta = rIbeta(...).T` or `r, I, beta = rIbeta(...)`?

> The 2-letter code origin position might be better using the `matplotlib` `loc` location code/string, which may be more familiar to users? 

I can add these location codes, but I never remember what the numbers mean... They are also not very systematic:
```
2 —  9 —  1
|    |    |
6 — 10 — 5,7
|    |    |
3 —  8 —  4
```
The full-word strings are more meaningful (except the bare `'right'`), I'll try to add them.

MikhailRyazanov:
@DanHickstein, @rth, could you please take a look at https://github.com/conda/conda/issues/8836#issuecomment-506460650 and below and modify our [appveyor.yml](https://github.com/PyAbel/PyAbel/blob/master/appveyor.yml) accordingly?

stggh:
> By the way, as the most active user, do you have an opinion on the output transposition?
r, I, beta = rIbeta(...).T or r, I, beta = rIbeta(...)?

Not sure about that. I "retire" today, well, stop being paid by the ANU. I will still be active, my lab has research funding until the end of 2019, and there are plenty of species to analyse in the future. This PR will be great for finding anisotropy anomalies, that are the signatures of vibronic coupling.

For the function return, I prefer the transpose to be applied, and the data returned as a tuple of 1D numpy arrays. i.e. the 2nd choice.

> full-word strings are more meaningful

I agree. It's your call on string 'vh' vs matplotlib style loc.

MikhailRyazanov:
OK, I'll then switch the return format to [ang_term][r], such that `I, beta = Ibeta()` will work directly, and change the documentation to suggest that those who need the opposite can use `.T`.

>> full-word strings are more meaningful
>
> It's your call on string 'vh' vs matplotlib style loc.

We can allow both, I already have the code (~9 additional lines). I'll wait until the AppVeyor problem is corrected and then commit it.

DanHickstein:
Great work and great discussion @MikhailRyazanov and @stggh!

>  I "retire" today, well, stop being paid by the ANU. 

Congratulations on your retirement Steve! Now that they aren't paying you anymore, you can work on whatever you want :)

> @DanHickstein, @rth, could you please take a look at conda/conda#8836 (comment) and below and modify our appveyor.yml accordingly?

I'm not sure if I understand @msarahan's suggestion completely, but I think that we can swap our current line 34 in appveyor.yml:
```- "set PATH=%MINICONDA%;%MINICONDA%\\Scripts;%PATH%"```
with:
```- CALL "%PYTHON_ROOT%\\Scripts\\activate.bat"```

> But the (row, col) convention is the standard for matrices, and if the users are going to mix our function calls with direct matrix operations, it will be more consistent.

Well, I think that you make some strong arguments that we should be using (row, column) everywhere. Since we are already using this as the standard for Abel.Transform (I forgot that this was the case), then it seems that we are already embracing this convention. We should probably open a new issue to discuss this and identify the places in the code where things need to change to adopt the same convention.

**Comment by MikhailRyazanov** [comment]
> I'm not sure if I understand @msarahan's suggestion completely, but I think that we can swap our current line 34 in appveyor.yml:
> `- "set PATH=%MINICONDA%;%MINICONDA%\\Scripts;%PATH%"`
> with:
> `- CALL "%PYTHON_ROOT%\\Scripts\\activate.bat"`

I understand even less, since I do not use Conda and don't know what that `activate.bat` is (plus, I don't know how AppVeyor works, what and where it has, and do not have access to its setup). But my guess is also that this is probably the what needs to be done. If you can make a PR with this change and see whether it helps, it would be great.

**Comment by DanHickstein** [comment]
I got: 
```
CALL "%PYTHON_ROOT%\\Scripts\\activate.bat"
The network path was not found.
```

**Comment by DanHickstein** [comment]
I also barely understand how AppVeyor works, except that it tests our software on Windows :)

I think that, as a member of the PyAbel team, you should be able to log into appveyor with your github account and see the build status and tinker with stuff. 

I tried the above idea in #258, and it didn't seem to work. Will try to work more on this later.

**Comment by MikhailRyazanov** [comment]
Great, now it works again.

**Comment by MikhailRyazanov** [comment]
OK, now it also accepts strings like `'upper left'` for the origin, and the result are returned transposed, such that they can be used directly as `I, beta = Ibeta(...)`.

The origin still uses the (row, column) convention, as it probably should (https://github.com/PyAbel/PyAbel/pull/257#issuecomment-506603278). And the (in)consistency of other tools indeed should be addressed in another issue and then a separate PR.

The remaining task is to update the “Anisotropy Parameter” section and examples. Does anybody have a 2-photon (or more) image with obvious higher-order anisotropy?

**Comment by stggh** [comment]
A nice piece of work @MikhailRyazanov 

There is a scaling issue for the radial intensity between the `angular_integration` and `rIbeta` methods, for the whole image transform (code below), and the single quadrant transform:

intvsbasex

factor of ~4.  Given @MikhailRyazanov's skills, this may be a bug in `angular_integration()`?

```python
import numpy as np
import abel
import matplotlib.pylab as plt

imagefile = 'O2-ANU1024.txt.bz2'
IM = np.loadtxt(imagefile)
IMc = abel.tools.center.center_image(IM, center='slice', square=True)


# Abel transform - basex ------------------------------------------
BIM = abel.Transform(IMc, method='basex',
                     transform_options=dict(reg=100, basis_dir='./', dr=1),
                     symmetry_axis=None, angular_integration=True)

r, I, beta = abel.tools.vmi.rIbeta(BIM.transform, window=50)

# plots ----------------------------
fig, ax = plt.subplots()

ax.plot(*BIM.angular_integration, lw=1, label='angular integration')
ax.plot(r, I/4, lw=1, label=r'rIbeta$/4$')
# ax.plot(r, beta*I, lw=1) 

ax.legend()
ax.set_xlabel('radius (pixels)')
ax.set_ylabel('intensity (arb. u.)')
ax.set_title(imagefile)
ax.axis(xmin=325, xmax=425, ymin=-1e4, ymax=1.5e6)

plt.subplots_adjust(left=0.15)
plt.savefig('intvsbasex.png', dpi=75, bb_inches='tight')
plt.show()
```

MikhailRyazanov:
My skills are not infallible. :–)
I was testing that a uniform ball _I_(_r_ < _R_, _θ_, _φ_) = 1 produces the radial intensity _I_(_r_ < _R_) = 4 _π_ r<sup>2</sup>. But probably have not tested it for all possible cases. Do the results differ depending on which image part (quadrant, half, whole) is used?

stggh:
Yes, oddly, a single quadrant appears to be x1/8. I will look more closely tomorrow.
b


```python
q = abel.tools.symmetry.get_image_quadrants(IMc)


# Abel transform - basex ------------------------------------------
bq0 = abel.basex.basex_transform(q[0], reg=100, basis_dir='./', dr=0.5)

speed = abel.tools.vmi.angular_integration(bq0, origin=(0, 0), dr=0.5)

r, I, beta = abel.tools.vmi.rIbeta(bq0, origin='lower left', window=50)
```

MikhailRyazanov:
My `Ibeta()` is at least internally consistent for different image parts:
```python
import numpy as np

import abel

IM = np.loadtxt('data/O2-ANU1024.txt.bz2')
#plt.imshow(IM); plt.show()

full = abel.Transform(IM, method='basex',
                      transform_options=dict(reg=100)).transform
#plt.imshow(full, clim=(0, 2)); plt.show()

height, width = full.shape
half = full[:, width // 2:]
Q = half[:height // 2 + 1]

Ifull, = abel.tools.vmi.Ibeta(full, 'cc', order=0)
Ihalf, = abel.tools.vmi.Ibeta(half, 'cl', order=0)
IQ,    = abel.tools.vmi.Ibeta(Q,    'll', order=0)

plt.plot(Ifull, '-',  label='full')
plt.plot(Ihalf, '--', label='half')
plt.plot(IQ,    ':',  label='Q')

plt.autoscale(tight=True)
plt.xlim(140, 430)
plt.legend()
plt.tight_layout()
plt.show()
```
produces practically identical distributions:
parts
And, as I said before, for a uniform “slice” (actually `np.ones()`) it gives 4 _π_ _r_<sup>2</sup>, as expected. So the problems are either somewhere else or appear only in some specific situations.

stggh:
The intensity should scale in proportion to the image size. The one quadrant profile should have 1/4 the full image profile?

MikhailRyazanov:
Why? The purpose of `Ibeta` is to estimate the original speed and anisotropy distributions, which do not depend on which image part we analyze. At least, your [expression](https://pyabel.readthedocs.io/en/latest/anisotropy_parameter.html) _I_ = _σ_<sub>total</sub>... does not have the image size. :–)

The purpose of `angular_integration` might be different, so it might take the size and shape of the image part into account, if this makes sense for some applications. I don't know.

stggh:
In the sense that the radial speed distribution <math>I(r)</math> is the <math>&theta;</math> angle integrated intensity <math>I(r, &theta;)</math>. 

 i.e. I think that the `abel.tools.vmi.angular_integration()` speed should agree with `abel.tools.vmi.rIbeta()`?

MikhailRyazanov:
1. “I(r, θ)” is ambiguous, since it can be the 3D distribution with implied no dependence on φ and corresponding to dΩ (equivalently, a 2D _slice_ of the 3D distribution), or actually I(r, θ) dr dθ integrated over φ (with the r sin θ weight).
2. “The θ angle integrated intensity” can also mean different things, depending on whether it is integrated over the full possible range (the “speed distribution” usually means integration over the full sphere, unless explicitly stated otherwise) or over the available range, or maybe over some user-specified range.

As I understand, currently `abel.tools.vmi.angular_integration()` is integrating over the angular range present in the image and allows choosing between the “I(r, θ)” interpretations by the `Jacobian` parameter (not sure about the _r_ factor for `Jacobian=False`). So, in general, it does not necessarily agree with `abel.tools.vmi.Ibeta()`. Maybe this integration over a limited angular range is also useful for some purposes (and then `angular_integration()` should probably accept `angular_ranges`, similarly to `radial_ranges` in `radial_integration()`). Otherwise, why do we need two different functions? :–)

stggh:
Good points. 

Does the magnitude of `I` from `Ibeta()` relate to the image intensity?

In the end both speed distribution methods may be (absolute) scaled as needed. 

IMO it would be nice if the total integral (over <math>r, &theta; &varphi;</math>)= # of particles. However, I am not sure of this is even the case for `angular_integration()`. 

This is aspect should not prevent merging your very useful code.

MikhailRyazanov:
I think, all our Abel-transform methods treat the intensity properly, producing the reconstructed image that corresponds to a slice of  1-pixel (voxel?) thickness. Then my radial intensity _I_(_r_) should equal the number of particles within a spherical shell of radius _r_ and 1-pixel thickness, and the sum of `I[r]` over `r` (or its integral, if I had _dr_ ≠ 1) should equal the total number of particles.

This can be easily tested by randomly sampling some 3D distribution, projecting it (= ignoring one coordinate), applying the inverse Abel transform and then feeding the result to `Ibeta`. It will not give the exact number of particles (due to the propagation of randomness), but should confirm whether the scaling is correct.

Actually, adding
```python
print('sum image =', np.sum(IM))
print('sum Ifull =', np.sum(Ifull))
print('sum Ihalf =', np.sum(Ihalf))
print('sum IQ    =', np.sum(IQ))
```
to my [example above](https://github.com/PyAbel/PyAbel/pull/257#issuecomment-507882878), I get
```
sum image = 58104223.0
sum Ifull = 58130837.06012754
sum Ihalf = 58368319.70721016
sum IQ    = 57760999.92453349
```
which looks close enough — 3 correct significant digits for the full image, and 2 digits for the parts (but the image itself is not perfectly symmetric).

stggh:
Looks good, although, I claim the sums should be in proportion to the whole image fraction, if they represent the number of particles: `Ihalf ~ 58M/2`, `IQ ~ 58M/4`.

For `PyAbel`:
```
sum image = 58104218
sum abel basex = 76474
sum speed basex = 18482469
sum abel 3pt = 76628
sum speed 3pt = 18494539
sum I = 58088472
```

The speed sum to whole image sum ratio is <math> 58104217/18482469 ~ &pi;</math> 
The Abel inverse sum to whole image sum ratio is <math>58104218/76474 = 760 &asymp; </math> some <math>R_max</math> value?

For a single quadrant:
```
sum image = 14440639
sum abel basex = 19050
sum speed basex = 4823003
sum abel 3pt = 18833
sum speed 3pt = 4824050
sum I = 57554465
```
all magnitudes reduced by <math>~1/4</math> (except for `Ibeta`).

So, I think @MikhailRyazanov's `Ibeta()` values look good, apart from the necessity for an image fraction (which may be in my own head). 

For the `PyAbel` inverse Abel methods the particle count is reduced by <math> ~ 1/R_max</math>, the speed intensity count is reduced by <math>~ 1/&pi;</math>. **However**, the repeat transform `image -> forward Abel -> inverse Abel -> image` yields the correct count = `image.sum()` for the double transformed image.  

It appears that there are Jacobian factors at play that make the speed intensity from `angular_integration()` ~ <math>1/&pi;</math> smaller than `Ibeta()`. The ~760 (<math> 1/R_max</math>) inverse Able image intensity reduction is unclear (to me).

MikhailRyazanov:
@stggh,
Currently `Distributions` solves the problem “estimate the original distributions from the available data”. What exactly is the problem you want it to solve and what practical applications the solution would have?

Keep in mind that:

1. For the number of particles within various image parts, for example, sum(left) + sum(right) ≠ sum(full). Because the axial pixels are included in both halves. Do you want ∫ _I_<sub>left</sub>(_r_) _dr_ = sum(left) and so on? Or do you want _I_<sub>left</sub>(_r_) + _I_<sub>right</sub>(_r_) = _I_<sub>full</sub>(_r_)?

2. A partially clipped image
clipped
has different angular ranges for different radii. It this case, _r_ < 300 has the full span, and for _r_ > 300 the range of present angles decreases progressively. Should _I_(_r_) drop accordingly for _r_ > 300? Or _I_(_r_) should at all _r_ be proportional to the original distribution, but globally downscaled to match the total number of detected particles?

> The whole image sum to Abel inverse sum ratio is 58104218/76474 = 760 ≈ some R_max value?

Basically, the inverse Abel transform produces a one-pixel central slice, which is ~2R_max times thinner than the whole 3D distribution, so its integrated intensity is proportionally smaller. The actual ratio depends on the radial and angular distributions (and is not very meaningful).

stggh:
> What exactly is the problem you want it to solve

Ideally, I would like the speed distribution from `abel.tools.vmi.Ibeta()` to match the speed distribution from `abel.tools.vmi.angular_integration()`.  They currently differ in amplitude. 

intvsbasex

MikhailRyazanov:
@DanHickstein, @stggh,
I have finally updated the “Anisotropy Parameter” section to mention these new tools (and did some other edits there). Please check that it looks fine, or maybe needs some corrections, and let's then merge this.

(Steve apparently gave up on persuading me to scale the intensity. :–) But we can continue if there are clear ideas how exactly this should be done and how it might be useful...)

stggh:
> Steve apparently gave up on persuading me to scale the intensity.

Yes, I did not want to be the road block for merging your nice work

> exactly this should be done

Good point

> how it might be useful

The two methods should be equivalent. One or the other requires a scale factor.  For another PR.

stggh:
Docs look beautiful. Great work, as usual.

MikhailRyazanov:
Thanks! Waiting for Dan's comments...

Regarding the scaling, `abel.tools.vmi.angular_integration()` definitely needs some adjustments, since it never reproduces the original image intensity, [as you have shown](https://github.com/PyAbel/PyAbel/pull/257#issuecomment-508327639). I'm not sure that `Ibeta()` and `angular_integration()` must be completely _equivalent_, and would expect that `angular_integration()` really does angular integration of the data, :–) but they should be _consistent_. Which is not a problem — if you know that you pass 1/2 or 1/4 of an image, you can easily divide or multiply the results by 2 or 4. Similarly, if needed, the total intensity can be easily renormalized to `image.sum()`. The problem is that the original (projected) image is not known inside `Ibeta()`, and doing this sort of scaling analytically in the general case, for arbitrary origins and weights, is not trivial. So I suggest to open an issue to discuss _what exactly_ is needed from `angular_integration()` and `Ibeta()` and then to modify them accordingly and to document clearly what they do and how they are related to each other.

PR #258, opened 2019-06-28
Appveyor fix

DanHickstein:
Trying to see if we can get Appveyor working again by calling "activate.bat" rather than setting the path manually, as discusesd in https://github.com/PyAbel/PyAbel/pull/257#issuecomment-506608835

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ init:
   - "ECHO %PYTHON% %PYTHON_VERSION% %PYTHON_ARCH% %MINICONDA%"
 
 install:
-  - CALL "%PYTHON_ROOT%\\Scripts\\activate.bat"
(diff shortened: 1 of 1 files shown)


DanHickstein:
I got 
```
CALL "%PYTHON_ROOT%\\Scripts\\activate.bat"
The network path was not found.
```

MikhailRyazanov:
I guess, in our case it should be `%MINICONDA%` instead of `%PYTHON_ROOT%`.

DanHickstein:
You are now the appveyor expert :)

DanHickstein:
I think that this can be merged now. Feel free to hit the green button.

MikhailRyazanov:
Thanks! I'll then merge this.

Issue #259, opened 2019-07-25
BASEX cached basis set issue


fourniermartin:
I'm having an issue re-using cached BASEX basis set. 

`imgout=abel.Transform(img, direction='inverse', method='basex').transform`

does work properly but fails to re-use the basis set, even when on the exact same image file.

`Loading basis sets...`
`Cached basis file incompatible.`
`A suitable basis set was not found.`
`A new basis set will be generated.`

I do not believe this is due to an error in PyAbel, but some dependency that was updated. I am not sure how to find the cause of the issue however. I'm thinking that the cached file is not read or saved properly? 

I am up to date on all dependencies, using python 3.7.4 on macOS.

MikhailRyazanov:
Try deleting all `basex` cache files and run again.

I have noticed previously that NumPy for Python 2 can read files written by NumPy for Python 3, but, surprisingly, this does not work in the opposite direction (Python 3 cannot read files written by Python 2). And, if I remember correctly, at some point, different versions of NumPy had format incompatibilities, so after updating NumPy, the basis-set files had to be regenerated.

Are you sure that PyAbel is always using the same Python and NumPy versions? Do the cache files/directories have correct permissions?

fourniermartin:
I have tried to wipe all files, or work from a different root location, the behavior is the same. Regarding python versions, I am exclusively using 3.7.4, installed with Homebrew. NumPy is version 1.16.4. Permissions seem correct.

The behavior is that is reads the file, finds it incompatible, generates a new one. The new one has a creation timestamp that fits the moment I launch the script.

MikhailRyazanov:
Mysterious...

Try to add `allow_pickle=True` as the last parameter or `np.load()` here:
https://github.com/PyAbel/PyAbel/blob/ddf5df2f1c044107bd9ab088bd117f87c2d4dd72/abel/basex.py#L347
and here:
https://github.com/PyAbel/PyAbel/blob/ddf5df2f1c044107bd9ab088bd117f87c2d4dd72/abel/basex.py#L368
(basically, to all occurrences of `np.load()` in `abel/basex.py`).

This is my guess based on that [`numpy.load()` docs](https://docs.scipy.org/doc/numpy/reference/generated/numpy.load.html) say that it throws `ValueError` if “the file contains an object array, but allow_pickle=False given”. Even though the same docs say that `allow_pickle=True` is the default. But maybe your installation has it disabled somehow?

[...]

PR #260, opened 2019-07-28
Drop PyAbel version from BASEX cache files

MikhailRyazanov:
As noticed in #259, recent versions of NumPy switched the default parameters of `numpy.load()` to `allow_pickle=False`, which makes any attempts to load `basex` cache files always fail and thus triggers basis-set generation every time. The problem is that the PyAbel version (`__version__`) is saved in the cache files, and NumPy need to “pickle”/“unpickle” this _string_, so `allow_pickle=True` is required in the `load()` functions. This was actually the default behavior, but now is disabled by default (allegedly, due to security reasons).

However, this version string, although saved and loaded, is in fact not used for any purposes. And other Abel methods do not even write it to their cache files. So, I think, the best solution is just to drop it from `basex` cache files, which is done in this PR. This also solves the problem that PyAbel executed under Python 3 could not read cache files generated by PyAbel executed under Python 2 (probably due to different assumed string encodings).

### DIFF
--- a/abel/basex.py
+++ b/abel/basex.py
@@ -344,7 +344,7 @@ def get_bs_cached(n, sigma=1.0, reg=0.0, correction=True,
                     print('Loading basis sets...')
                     # saved as a .npy file
                 try:
-                    M, Mc, M_version = np.load(full_path(best_file))
+                    M, Mc = np.load(full_path(best_file))
                     # crop if loaded larger
                     if M.shape != (n, nbf):
                         M = M[:n, :nbf]
@@ -365,7 +365,7 @@ def get_bs_cached(n, sigma=1.0, reg=0.0, correction=True,
 
             # Try to extend the largest available
             try:
-                oldM, oldMc, M_version = np.load(full_path(largest_file))
+                oldM, oldMc = np.load(full_path(largest_file))
                 if verbose:
                     print('(extending {})'.format(largest_file))
             except:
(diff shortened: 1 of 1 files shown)


DanHickstein:
Thanks for figuring this out Mikhail! 

This seems like a good solution for now. For the record, at one point we did discuss checking the version of PyAbel that wrote the basis sets, so that we wouldn't be loading incorrect basis sets if the user updates their version of PyAbel and we have made some change to the way that basis sets are generated. In general, this still seems like a good idea to me. But, since we are currently doing no checking of the version, then there is probably no reason to read/write the version to the files.

MikhailRyazanov:
@DanHickstein, I understand that being able to detect “incorrect” (old-version) basis-set caches is important, but I think that using the PyAbel version for this is not a good idea:
1. Not every PyAbel version changes cache formats for every method, so if we just directly compare the PyAbel versions, this will lead to mostly unnecessary recalculations after every update.
2. It seems that PyAbel versions are changed in separate pull requests than the methods are, thus between PyAbel releases the behavior of affected methods would be inconsistent.

So maybe keeping internal “method versions” would be better for this purpose.

However, I do not expect `basex` changes in the near future, :–) and when we have some incompatible changes in any of the caching methods, we should implement some reasonable versioning system for all of them consistently. This PR also brings some consistency — neither method would store the version — and after implementing the versioning, it will be easy to detect these “old” files.

MikhailRyazanov:
Thanks to @fourniermartin for reporting #259 and testing this, and to @stggh and @DanHickstein for reviewing!

I'm going to merge this...

DanHickstein:
Merge away!

I agree with your identified flaws in saving the pyabel version with the basis sets. The version doesn't necessarily correlate with changes to our basis set generation methods.

Really, we would want something similar to how .pyc files are generated: https://stackoverflow.com/questions/15839555/when-are-pyc-files-refreshed
Basically, we could check that the basis set files have a date-stamp that is newer than basex.py (or whatever method is used). If the basis sets are out-of-date, we could either re-generate them or use them anyway. I would suggest that, by default, we just re-generate them. Perhaps we allow an optional keyword argument to allow out-of-date basis sets to be used. This would allow users with a huge basis set stockpile to keep using their basis sets if necessary, but means that the 99.9% of users that don't care much about the time it takes to generate basis sets will always have up-to-date basis sets.

PR #262, opened 2019-07-30
Increment version to 0.8.3 and included RSI reference to readme

DanHickstein:
This bumps the version to 0.8.3, mainly to allow users to access new features in #257 and the bug fix in #260. Big thanks to @MikhailRyazanov for all of the great work! 

Also, I added the RSI reference to the readme file.

@PyAbel/developers, anything else we should include before we release v0.8.3?


MikhailRyazanov:
While we are here, maybe also include some minor corrections to README.rst?
```
inverse abel transform
        ^ uc

numerical Abel Transform
               ^ lc

Numpy and Scipy
   ^ uc      ^ uc

and Matplotlib is required
    ^^^^^^^^^^ link

no warrenty
       ^ "a"

implementation of the Abel Transform
                           ^ lc
```

DanHickstein:
@stggh, the link to the zenodo is the "latest version" link, so it should always go to the most recent version.

@MikhailRyazanov, thanks for catching those issues with the readme! I made the changes.

MikhailRyazanov:
I've also noticed that https://pyabel.readthedocs.io/ is still called “PyAbel **0.7**”! This seems to be defined here:
https://github.com/PyAbel/PyAbel/blob/e0a9961f86d48afaae5dccb6bbaf1a0525560e1d/doc/conf.py#L90-L93
which also needs to be updated. And it apparently can be done automatically like
```python
import abel
# The short X.Y version.
version = abel._version.__version__
# The full version, including alpha/beta/rc tags.
release = version
```
or so (at least, Sphinx itself [does it like this](https://github.com/sphinx-doc/sphinx/blob/47cd262b3e50ed650a82f272ba128a1f872cda4d/doc/conf.py#L5-L19)).

DanHickstein:
Good catch @MikhailRyazanov! 

I made the change that you suggested. 

Should I go ahead and merge now?

MikhailRyazanov:
> Should I go ahead and merge now?

Yes! :–)

DanHickstein:
Done! :)

MikhailRyazanov:
Thanks!

Will a new [release](https://github.com/PyAbel/PyAbel/releases) appear automatically, or you need to do something to create it?

DanHickstein:
I needed to make a new release. I just did it. It's in the "releases" tab.

Issue #265, opened 2019-10-17
help for including odd orders during basex inversion


haribo3127:
Hi,

I have a left-right symmetric images, but with positive values in the north hemisphere and negative values in the south hemisphere. I would like to force the use of odd orders for the inversion.

With the basex method, I can reconstruct such images but it is not possible to extract odd parameters. Is there a way to do so with this method? For example, linbasex allows to set legendre_order[1, 3].

Thanks!

MikhailRyazanov:
I actually want to add odd orders to `abel.tools.vmi.Distributions`, but I need to know what is the most reasonable convention for them.

I saw some people using just a finite number of odd Legendre polynomials and some other people using even Legendre polynomials multiplied by cos θ (which is essentially equivalent to using odd polynomials, but with different conventions). And my impression is that in most physically meaningful cases, asymmetric distributions cannot be described _exactly_ by a finite number of odd orders (in contrast to many cases where even polynomials up to some predetermined order give the exact description), although I did not see any good literature about this.

Can you please explain what exactly you need and why (like what is the mathematical model of your angular distributions)? References to what _other_ people might need will be also very useful.

haribo3127:
Sorry for my late reply. Indeed there seems to be a very few cases where the use of odd orders is meaningful. One of them is regarding photoelectron distributions of optically active (chiral) system. In this case one typically look at differential images where negative values can arise for one of the hemisphere.

I would refer to the following two references for the formalism, the first one gives a general overview for applying higher odd orders to the inversion procedure.
 
1. https://aip.scitation.org/doi/full/10.1063/1.4844295 (Section II)
2. https://journals.aps.org/pra/abstract/10.1103/PhysRevA.13.1411 (Original work)

In those cases, one already has inital information from the physical system to define the orders to be used for the reconstruction. Am I clear enough?

haribo3127:
as an additional comment:
from the 'classResults(r, cn)' one see that beta0 = 1 by definition (does this mean that each beta parameters are normalized by beta0?). As some information can also be contained in beta0 does it makes sense to relax this constraint in the abel.tools.vmi.Distributions?

MikhailRyazanov:
Thanks for the references! It seems to make sense that non-oriented chiral systems under circularly polarized light might produce a finite number of odd terms, although I did not read into details yet to see why higher-order terms vanish (and Janssen's group work does not seem to explain whether they have actually derived that multiphoton formula or just extended it by analogy).

[...]

PR #266, opened 2019-10-31
Odd orders for abel.tools.vmi.Distributions

MikhailRyazanov:
This is (finally!) an implementation of odd angular orders for the anisotropy distributions, to finish with #254 and address #265.

The behavior and performance for even orders (still the default mode) should not be affected, but odd terms can now be included by specifying `odd=True` after `order` or supplying an odd `order`.

An example with pure Legendre polynomials:
```python
import numpy as np
from abel.tools.vmi import Distributions

fig = plt.figure(figsize=(16, 8))

R = 150
n = 2 * R + 1  # width
m = 2 * R + 1  # height
x0 = R; y0 = R  # center
# coordinates:
x = np.arange(float(n)) - x0
y = y0 - np.arange(float(m))[:, None]
r = np.sqrt(x**2 + y**2)
# cos, sin:
r[y0, x0] = np.inf
s = x / r
c = y / r
r[y0, x0] = 0

sigma = 5.0  # peak SD
step = 5 * sigma  # distance between centers

def peak(i):
  return np.exp(-(r - (i + 1) * step)**2 / (2 * sigma**2))
  
IM = peak(0) + \
     peak(1) * c + \
     peak(2) * (3 * c**2 - 1) / 2 + \
     peak(3) * (5 * c**3 - 3 * c) / 2 + \
     peak(4) * (35 * c**4 - 30 * c**2 + 3) / 8

fig.add_subplot(1, 2, 1)
plt.imshow(IM, cmap='bwr')

distr = Distributions((y0, x0), order=4, odd=True)
res = distr(IM)

fig.add_subplot(2, 2, 2)
for n, c in enumerate(res.cos()):
  plt.plot(c, label='$\\cos^{}\\theta$'.format(n))
plt.legend()

fig.add_subplot(2, 2, 4)
for n, h in enumerate(res.harmonics()):
  plt.plot(h, label='$P_{}(\\cos\\theta)$'.format(n))
plt.legend()
plt.xlabel('$r$')

plt.tight_layout()
plt.show()
```
harmonics

And the same, but with radial intensity and anisotropy distributions for a non-negative image:
```python
...
IM = peak(0) + \
     peak(1) * (1 + c) + \
     peak(2) * (1 + (3 * c**2 - 1) / 2) + \
     peak(3) * (1 + (5 * c**3 - 3 * c) / 2) + \
     peak(4) * (1 + (35 * c**4 - 30 * c**2 + 3) / 8)
...
plt.imshow(IM, cmap='binary')
...
(I,), beta = np.vsplit(res.Ibeta(), (1,))
...
plt.plot(I, label='$I$')
...
for n, b in enumerate(beta):
  plt.plot(b, label='$\\beta_{}$'.format(1 + n))
...
```
Ibeta

If @haribo3127 (the initiator of #265) or anybody else can provide an experimental image for testing, it would be great. Other testing, feedback and suggestions are also very welcome.

### DIFF
--- a/abel/tests/test_tools_distributions.py
+++ b/abel/tests/test_tools_distributions.py
@@ -8,6 +8,9 @@ from abel.tools.vmi import Distributions, harmonics
 
 
 def test_origin():
+    """
+    Test symbolic origin wrt corresponding numeric form.
+    """
     # image size
     n = 11  # width
     m = 20  # height
@@ -28,8 +31,13 @@ def test_origin():
 
 
 def run_order(method, tol):
-    sigma = 5.0
-    step = 6 * sigma
+    """
+    Test cossin distributions for even orders using Gaussian peaks.
+
+    tol = list of tolerances for order=0, order=2, order=4, ...
+    """
+    sigma = 5.0  # peak SD
+    step = 6 * sigma  # distance between peak centers
 
     for n in range(len(tol)):  # order = 2n
         size = int((n + 2) * step)
@@ -43,10 +51,12 @@ def run_order(method, tol):
         # radius
         r = np.sqrt(r2)
 
+        # Gaussian peak with one cossin angular term.
         def peak(i):
             return c2 ** (n - i) * s2 ** i * \
                    np.exp(-(r - (i + 1) * step) ** 2 / (2 * sigma**2))
 
+        # quadrant with all peaks
         Q = peak(0)
         for i in range(1, n + 1):
             Q += peak(i)
@@ -56,6 +66,7 @@ def run_order(method, tol):
                     format(rmax, 2 * n, method)
             res = Distributions('ul', rmax, 2 * n, method=method).image(Q)
             cossin = res.cossin()
+            # extract values at peak centers
             cs = np.array([cossin[:, int((i + 1) * step)]
                            for i in range(n + 1)])
             assert_allclose(cs, np.identity(n + 1), atol=tol[n],
@@ -77,8 +88,13 @@ def test_order_remap():
 
 
 def run_order_odd(method, tol):
-    sigma = 5.0
-    step = 6 * sigma
+    """
+    Test cossin distributions including odd orders using Gaussian peaks.
+
+    tol = list of tolerances for order=0, order=1, order=2, ...
+    """
+    sigma = 5.0  # peak SD
+    step = 6 * sigma  # distance between peak centers
 
     for n in range(0, len(tol)):  # order = n
         size = int((n + 2) * step)
@@ -93,12 +109,14 @@ def run_order_odd(method, tol):
         s = x / r
         r[size, 0] = 0
 
+        # Gaussian peak with one cossin angular term
         def peak(i):
-            m = n - i
-            k = (n - m) & ~1  # round down to even
+            m = n - i  # cos power
+            k = (n - m) & ~1  # sin power (round down to even)
             return c ** m * s ** k * \
                    np.exp(-(r - (i + 1) * step) ** 2 / (2 * sigma**2))
 
+        # quadrant with all peaks
         Q = peak(0)
         for i in range(1, n + 1):
             Q += peak(i)
@@ -109,6 +127,7 @@ def run_order_odd(method, tol):
             res = Distributions((int(size), 0), rmax, n, odd=True,
                                 method=method).image(Q)
             cossin = res.cossin()
+            # extract values at peak centers
             cs = np.array([cossin[:, int((i + 1) * step)]
                            for i in range(n + 1)])
             assert_allclose(cs, np.identity(n + 1), atol=tol[n],
@@ -131,6 +150,9 @@ def test_order_odd_remap():
 
 def run_method(method, rmax, tolP0, tolP2, tolI, tolbeta, weq=True):
     """
+    Test harmonics and Ibeta for various combinations of origins and weights
+    for default order=2, odd=False.
+
     method = method name
     rmax = 'MIN' or 'all'
     tol... = (atol, rmstol) for ...
@@ -146,9 +168,11 @@ def run_method(method, rmax, tolP0, tolP2, tolI, tolbeta, weq=True):
     # peak SD
     sigma = 2.0
 
+    # Gaussian peak.
     def peak(i, r):
         return np.exp(-(r - i * step)**2 / (2 * sigma**2))
 
+    # Test image.
     def image():
         # squared coordinates:
         x2 = (np.arange(float(n)) - x0)**2
@@ -165,6 +189,7 @@ def run_method(method, rmax, tolP0, tolP2, tolI, tolbeta, weq=True):
              c2 * peak(3, r)
         return IM, np.sqrt(s2)  # image, sin theta
 
+    # Reference distribution for test image.
     def ref_distr():
         r = np.arange(R + 1)
         P0 = 2/3 * peak(1, r) + \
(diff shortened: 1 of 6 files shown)


MikhailRyazanov:
I'm going to merge this, if nobody objects...

DanHickstein:
Sorry for the delay, I'll have a look at the code now...

DanHickstein:
The parsing of the anisotropy parameters is a little in-obvious:

``` python
(I,), beta = np.vsplit(res.Ibeta(), (1,))
```

Can we include an example of how to do this in the docstring for the `Ibeta` function?

abel/tools/vmi.py
```diff
@@ -475,7 +475,11 @@ class Distributions(object):
             angular dependences must be inferred from very small available
             angular ranges)
     order : int
-        highest order in the angular distributions. Even number ≥ 0.
+        highest order in the angular distributions, ≥ 0
+    odd : bool
+        include odd angular orders (is enabled automatically if **order** is
```
DanHickstein: ...it is enabled automatically if order is odd, *otherwise is False*?

stggh:
Looks good to me. `pycodestyle vmi.py` is not happy with inline `if` statements, but they were not introduced in this PR.

(PS: making great use of your PR #254)

MikhailRyazanov:
@DanHickstein, thank for reviewing!

> The parsing of the anisotropy parameters is a little in-obvious:
> ```(I,), beta = np.vsplit(res.Ibeta(), (1,))```
> Can we include an example of how to do this in the docstring for the Ibeta function?

I don't think that this Python 2 tricky stuff will be needed in real life. Basically, people will use either
```I, beta1, beta2 = res.Ibeta()```
when they need particular orders. Or, if they need to separate `I` and collect all `beta`s, the readable Python 3 example
```I, *beta = res.Ibeta()```
is given in the documentation. As I remember, I initially had that Python 2 example in the docstring, but then removed it, since it looks ugly, and Python 2 anyway officially retires in April 2020... Do you think that we need it?

Regarding the changelog, I was actually going to ask you about reviving it for my next PR (addressing #263) with incompatible changes, but starting here is probably even better. It seems that it has not been updated adequately for a long time and that the current format is not the best. I would recommend to list changes by release versions (not just dates) and present them as itemized lists under Added/Changed/Removed subsections. Something [like this](https://github.com/olivierlacan/keep-a-changelog/blob/master/CHANGELOG.md) (and perhaps formatted with 80-character line limit, such that it remains readable as plain text). Let's discuss!

File: abel/tools/vmi.py (hunk 1 above)
Comment (MikhailRyazanov) [reply abel/tools/vmi.py]: Well, the default is indeed `False`, as can be seen in the signature. Do we need to repeat it explicitly in the description? Then this probably should be done for all other parameters as well. Or just the phrase is unclear?

File: abel/tests/test_tools_distributions.py [hunk 2]
```diff
@@ -76,6 +76,59 @@ def test_order_remap():
...
+
+        for rmax in ['MIN', 'all']:
+            param = '-> rmax = {}, order={}, method = {}'.\
+                    format(rmax, n, method)
+            res = Distributions((int(size), 0), rmax, n, odd=True,
+                                method=method).image(Q)
+            cossin = res.cossin()
+            cs = np.array([cossin[:, int((i + 1) * step)]
+                           for i in range(n + 1)])
+            assert_allclose(cs, np.identity(n + 1), atol=tol[n],
```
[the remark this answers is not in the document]
MikhailRyazanov: OK. I saw that other tests are not (very) documented, so did not pay much attention here... I'll add some comments about what's going on.

abel/tools/vmi.py
DanHickstein: Yeah, I just wanted the description to say what the default was. Probably a good idea to have this for the other parameters as well, but since you were just changing this one in this PR, you don't need to change the others as well.

DanHickstein:
Okay, if we have the Python 3 example in the documentation, that seems fine. I think that Python 2 actually dies in only a few days! (Jan 1, 2020). 

I like the idea of organizing the changelog by version. I kinda like having dates for each change, and I'm not sure if we really need to added/changed/removed subsections, since we usually don't have that many changes per version. But, if you like the subsections, then that's fine with me. As long as we are using the changelog versus not using it, that seems like an improvement to me :).

abel/tests/test_tools_distributions.py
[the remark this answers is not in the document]
MikhailRyazanov: I've added some documentation and more comment to distribution test. Please check 
9af4978!

abel/tools/vmi.py
MikhailRyazanov: OK, I've changed in to “include odd angular orders. By default is ``False``, but is enabled automatically if **order** is odd.” and mentioned defaults for **order** and **use_sin** too (other parameters already had or do not need them) in 9af4978. Looks good now?

MikhailRyazanov:
@DanHickstein, @stggh, I've done some documentation changes suggested by Dan (see the last commit) and also prepared a draft of the changelog updated with the changes made during my presence here (after release v0.8.0) — see [here](https://github.com/MikhailRyazanov/PyAbel/blob/changelog/CHANGELOG.rst) (and how it looks [in plain text](https://raw.githubusercontent.com/MikhailRyazanov/PyAbel/changelog/CHANGELOG.rst)). I've only listed changes in code behavior, which are important for end users, but did not include documentation updates and technical changes (CI, docs generation...), as these less interesting fine details can be found in the Git log or “Pull requests”. What do you think about the format, wording and so on?

stggh:
Happy New Year from a fire ravaged, smoke polluted, Australia.

@MikhailRyazanov  your changelog looks great. Thanks, for all of your fantastic work on PyAbel.

MikhailRyazanov:
@DanHickstein, what do you think about the new changelog (see my comment above)?
I guess, we should also add it the documentation TOC, either at the end (after “Contributing”) or after “README”.

DanHickstein:
This looks great!

And I'd say that we should go ahead with you suggested formatting of the changelog. Do you want to incorporate those past changes as part of this PR?

MikhailRyazanov:
I've reformatted the existing old entries (although without versions), but somebody will need to go over the past releases and fill the gap. We can do this as a separate PR later...

The changelog is now also included in the documentation as the last TOC item (it looks better there than after README).

So if everything is fine, I'm going to merge this and move to the center/origin story...

MikhailRyazanov:
@DanHickstein, something strange happened with Travis CI: as you can see (?), it's checks are duplicated, and one is shown as “Build Passed”, whereas the other one is hanging at “Queued — Build Created” since yesterday. What is even more strange, if you go to that “queued” one, it actually shows _the same jobs_ as the “passed” one, and if you click them, inside they all are shown as “passed”...

This might be related to https://github.com/travis-ci/travis-ci/issues/10204 or might be something else/new (at least I haven't changed anything build-related, so this effect should be from some changes in their tools or GitHub configuration). Can you (the owner) please check whether the setting have something suspicious? If there is nothing obvious, we should probably just merge this PR and see what happens with the next...

DanHickstein:
There was a Travis Webhook and a Travis GitHubApp. So, I removed the webhook, and nothing changed. Then, I uninstalled and re-installed the GitHubApp. Now there are three travis builds :). I'm hoping that two of them are old and future PRs will only have one.

So, I think that we should go ahead and merge this and see what happens with the next PR. 

Thanks again for your fantastic work with this PR!

MikhailRyazanov:
OK, thanks!

Issue #263, opened 2019-08-17
Center/origin consistency


MikhailRyazanov:
Let's discuss the issue raised in https://github.com/PyAbel/PyAbel/pull/257#issuecomment-505319466 about the conventions related to image coordinates. There are actually two problems:
1. Some places use “origin” and some use “center” to refer to the same thing, and some use “center” to refer to different things.
2. Some places use the (row, column) convention, as for matrix indices, counting from the upper left corner, while some use (_x_, _y_) — in the opposite order and counting from the _bottom_ left corner.

### `center`

[README.rst](https://github.com/PyAbel/PyAbel/blob/master/README.rst)
uses “center” and “adopts the "television" convention, where IM[0,0] refers to the upper left corner of the image”.

[abel/transform.py](https://github.com/PyAbel/PyAbel/blob/master/abel/transform.py)
`Transform` uses `center` more as a verb (how/where _to center_ the image) and the (row, column) convention:
https://github.com/PyAbel/PyAbel/blob/a5d4f11615957d46145f2e3a42725f16783cb56b/abel/transform.py#L130-L136

[abel/tools/center.py](https://github.com/PyAbel/PyAbel/blob/master/abel/tools/center.py)
`find_center()` uses `center` for several different things: the parameter actually selects the _method_, and in
https://github.com/PyAbel/PyAbel/blob/a5d4f11615957d46145f2e3a42725f16783cb56b/abel/tools/center.py#L30-L31
one instance means the geometric center of the matrix, while the other means the symmetry axis/plane location.
The result is in the (row, column) format.
(And, by the way, the `square` parameter is not used and probably should be deleted.)

`center_image()` — same as in `Transform` above, with the same problems as in `find_center()`.

`set_center()` uses the `center` parameter for coordinates in the (row, column) format.

`find_center_by_...()` are called from `find_center()` and use the same conventions. Except that
https://github.com/PyAbel/PyAbel/blob/a5d4f11615957d46145f2e3a42725f16783cb56b/abel/tools/center.py#L318-L322
returns (col, row) instead of (row, col) — apparently a **bug**.

[abel/tools/circularize.py](https://github.com/PyAbel/PyAbel/blob/master/abel/tools/circularize.py)
`circularize_image()` uses the `center` parameter to select the _method_ or to pass the coordinates as (row, column).

### `origin`
[abel/tools/circularize.py](https://github.com/PyAbel/PyAbel/blob/master/abel/tools/circularize.py)
[...]

PR #267, opened 2020-01-12
Center/origin consistency

MikhailRyazanov:
This one is difficult but is better to be done sooner than later...
It addresses the inconsistent conventions for the image origin, see #263, and also tries to rename the related functions and parameters more appropriately, as also discussed in #263.

## Summary
* The image origin (center of symmetry) is now called “origin” everywhere.
* The word “center” now refers to the geometric center (midpoint) of the frame (matrix) and as a verb refers to shifting/cropping the image to move its _origin_ to the _center_.
* The parameter selecting the method of this centering is now called `method` when possible (some functions already had a parameter `method` for selecting the Abel transform).
* Most importantly, all functions that accept or return the origin are now using the (row, column) format, which is consistent with matrix indexing in NumPy and `scipy.ndimage`.
* To facilitate coding, negative values are now allowed for row and column, with the same meaning as in Python and NumPy, that is, counting form the end. For example, the quadrant Q0 (upper right) can have its origin (lower left) defined as `[-1, 0]` (_last_ row, 0th column).

## Compatibility issues
The renaming is accompanied by compatibility “shims” that print a deprecation warning and pass control to the corresponding new methods (or copy the parameters). This should not break old programs and should make their conversion to the new API easier (the warning tells where the problem is and what to use instead).

The “(row, column)” change might be problematic, because there is no automatic way to tell whether the code is using the old format (which sometime was (x, y) from the _bottom_ left, sometimes (column, row), and sometimes (row, column) as now) or the new. In some cases, when the function/parameter name is changed, the deprecation warning described above should help to locate the problematic code, but in general it seems that manual testing is the only way. I have mentioned this in the changelog, hopefully people will pay attention to it...

So, everybody, please test how your existing code works with the new API. Maybe we can develop better recommendations to simplify the transition (and/or do some other changes to the API).

## Additional changes
During this revision I've noticed that some links (internal and external) were broken or had errors, so I have repaired them and reformatted all literature references consistently, with full titles and author lists. And linked the examples that were referenced (and actually existed) but not included in the documentation.

### DIFF
--- a/abel/tools/polar.py
+++ b/abel/tools/polar.py
@@ -56,6 +56,13 @@ def reproject_image_into_polar(data, origin=None, Jacobian=False,
     ny, nx = data.shape[:2]
     if origin is None:
         origin = (ny // 2, nx // 2)
+    else:
+        origin = list(origin)
+        # wrap negative coordinates
+        if origin[0] < 0:
+            origin[0] += ny
+        if origin[1] < 0:
+            origin[1] += nx
 
     # Determine that the min and max r and theta coords will be...
     x, y = index_coords(data, origin=origin)  # (x,y) coordinates of each pixel
@@ -112,6 +119,11 @@ def index_coords(data, origin=None):
         origin_x, origin_y = nx // 2, ny // 2
     else:
         origin_y, origin_x = origin
+        # wrap negative coordinates
+        if origin_y < 0:
+            origin_y += ny
+        if origin_x < 0:
+            origin_x += nx
 
     x, y = np.meshgrid(np.arange(float(nx)) - origin_x,
                        origin_y - np.arange(float(ny)))
--- a/abel/tools/polar.py
+++ b/abel/tools/polar.py
@@ -48,7 +48,7 @@ def reproject_image_into_polar(data, origin=None, Jacobian=False,
     Notes
     -----
     Adapted from: 
-    http://stackoverflow.com/questions/3798333/image-information-along-a-polar-coordinate-system
+    https://stackoverflow.com/questions/3798333/image-information-along-a-polar-coordinate-system
     
     """
     # bottom-left coordinate system requires numpy image to be np.flipud
--- a/abel/tools/polar.py
+++ b/abel/tools/polar.py
@@ -14,36 +14,38 @@ from scipy.optimize import curve_fit, minimize
 def reproject_image_into_polar(data, origin=None, Jacobian=False,
                                dr=1, dt=None):
     """
-    Reprojects a 2D numpy array (``data``) into a polar coordinate system.
-    "origin" is a tuple of (x0, y0) relative to the bottom-left image corner,
-    and defaults to the center of the image.
+    Reprojects a 2D numpy array (**data**) into a polar coordinate system,
+    with the pole placed at **origin** and the angle measured clockwise from
+    the upward direction. The resulting array has rows corresponding to the
+    radial grid, and columns corresponding to the angular grid.
 
     Parameters
     ----------
     data : 2D np.array
-    origin : tuple
-        The coordinate of the image origin, relative to bottom-left
-    Jacobian : boolean
+        the image array
+    origin : tuple or None
+        (row, column) coordinates of the image origin. If ``None``, the
+        geometric center of the image is used.
+    Jacobian : bool
         Include `r` intensity scaling in the coordinate transform.
         This should be included to account for the changing pixel size that
         occurs during the transform.
     dr : float
-        Radial coordinate spacing for the grid interpolation
-        tests show that there is not much point in going below 0.5
-    dt : float
-        Angular coordinate spacing (in radians)
-        if ``dt=None``, **dt** will be set such that the number of theta values
-        is equal to the maximum value between the height or the width of
-        the image.
+        radial coordinate spacing for the grid interpolation.
+        Tests show that there is not much point in going below 0.5.
+    dt : float or None
+        angular coordinate spacing (in radians).
+        If ``None``, the number of angular grid points will be set to the
+        largest dimension (the height or the width) of the image.
 
     Returns
     -------
     output : 2D np.array
-        The polar image (r, theta)
+        the polar image (r, theta)
     r_grid : 2D np.array
         meshgrid of radial coordinates
     theta_grid : 2D np.array
-        meshgrid of theta coordinates
+        meshgrid of angular coordinates
 
     Notes
     -----
@@ -51,12 +53,9 @@ def reproject_image_into_polar(data, origin=None, Jacobian=False,
     https://stackoverflow.com/questions/3798333/image-information-along-a-polar-coordinate-system
 
     """
-    # bottom-left coordinate system requires numpy image to be np.flipud
-    data = np.flipud(data)
-
     ny, nx = data.shape[:2]
     if origin is None:
-        origin = (nx // 2, ny // 2)
+        origin = (ny // 2, nx // 2)
 
     # Determine that the min and max r and theta coords will be...
     x, y = index_coords(data, origin=origin)  # (x,y) coordinates of each pixel
@@ -75,49 +74,47 @@ def reproject_image_into_polar(data, origin=None, Jacobian=False,
     theta_i = np.linspace(theta.min(), theta.max(), nt, endpoint=False)
     theta_grid, r_grid = np.meshgrid(theta_i, r_i)
 
-    # Project the r and theta grid back into pixel coordinates
+    # Convert the r and theta grids to Cartesian coordinates
     X, Y = polar2cart(r_grid, theta_grid)
-
-    X += origin[0]  # We need to shift the origin
-    Y += origin[1]  # back to the bottom-left corner...
-    xi, yi = X.flatten(), Y.flatten()
-    coords = np.vstack((yi, xi))  # (map_coordinates requires a 2xn array)
+    # then to a 2×n array of row and column indices for np.map_coordinates()
+    rowi = (origin[0] - Y).flatten()
+    coli = (X + origin[1]).flatten()
+    coords = np.vstack((rowi, coli))
 
     zi = map_coordinates(data, coords)
     output = zi.reshape((nr, nt))
 
     if Jacobian:
-        output = output * r_i[:, np.newaxis]
+        output *= r_i[:, np.newaxis]
 
     return output, r_grid, theta_grid
 
 
 def index_coords(data, origin=None):
     """
-    Creates `x` and `y` coordinates for the indices in a numpy array.
+    Creates `x` and `y` coordinates for the indices in a numpy array, relative
+    to the **origin**, with the `x` axis going to the right, and the `y` axis
+    going `up`.
 
     Parameters
     ----------
     data : numpy array
-        2D data
-    origin : tuple
-        (x, y). Defaults to the center of the image. Specify ``origin=(0,0)``
-        to set the origin to the *bottom-left* corner of the image.
+        2D data. Only the array shape is used.
+    origin : tuple or None
+        (row, column). Defaults to the geometric center of the image.
 
     Returns
     -------
-        x, y : arrays
+        x, y : 2D numpy arrays
     """
     ny, nx = data.shape[:2]
     if origin is None:
         origin_x, origin_y = nx // 2, ny // 2
     else:
-        origin_x, origin_y = origin
+        origin_y, origin_x = origin
 
-    x, y = np.meshgrid(np.arange(float(nx)), np.arange(float(ny)))
-
-    x -= origin_x
-    y -= origin_y
+    x, y = np.meshgrid(np.arange(float(nx)) - origin_x,
+                       origin_y - np.arange(float(ny)))
     return x, y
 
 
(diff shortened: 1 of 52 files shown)


MikhailRyazanov:
Additional thoughts and noticed problems (not caused by my changes but somewhat related to the topic):

----

`abel.tools.vmi.radial_integration(IM, radial_ranges=None)`
should probably have the `origin` parameter, as
`abel.tools.vmi.angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None)`
does.

----

`abel.tools.center.find_origin_by_convolution(IM, **kwargs)`
has a poor signature, and its docsting says:
“**projections** (_any_) – if this parameter is present, the autoconvoluted projections along both axes will be returned after the origin.”
I guess, it would make more sense to declare it as
`abel.tools.center.find_origin_by_convolution(IM, projections=False)`
and describe “**projections** (_bool_) – ...”.

----

`abel.tools.polar.reproject_image_into_polar()` produces inconsistent angular ranges for different quadrants and half-planes. Probably needs some attention.

----

The ` abel.transform.Transform` documentation mentioned that `origin=(None, 250)` (actually `center=(None, 250)` before these changes) can be used for centering along one axis, but this apparently did not work, so I have removed this claim. Was it ever working? How did it break? Does anybody need this capability (I guess, @DhrubajyotiDas might)?

I haven't tested, but from the code it seems that `crop='maintain_data'` also should not work if at least one of the shifts is zero.

----

The `'slice'` method in `abel.tools.center.find_origin()` has some implementation problems. There are no comments, so I do not know how it is supposed to work, but the following code illustrates wrong things that can happen:
```
from __future__ import division, print_function
import numpy as np
from abel.tools.center import func_method, find_origin

# image size:
rows, cols = 100, 110
# origin:
row, col = 60, 40  # 'slice' returns ~52, ~40.5
#row, col = 55, 50  # 'slice' fails (verbose undefined <= fit failure)

r = np.arange(rows)[:, None]
c = np.arange(cols)
IM = np.exp(-((r - row)**2 + (c - col)**2) / 20**2)
plt.imshow(IM); plt.tight_layout(); plt.show()

print('size', IM.shape)
print('Origin:\nactual', (row, col))
for method in sorted(func_method):
    print(method, find_origin(IM, method))
```
That is, for a reasonable distribution it can return wrong parameters or fail completely.

abel/tools/polar.py
```diff
@@ -14,127 +14,136 @@
...
+        origin = (ny // 2, nx // 2)
+    else:
+        origin = list(origin)
+        # wrap negative coordinates
+        if origin[0] < 0:
+            origin[0] += ny
+        if origin[1] < 0:
+            origin[1] += nx
 
     # Determine that the min and max r and theta coords will be...
```
DanHickstein: There is a typo on this line (I know that you didn't add this, but I just noticed it now). It should say "Determine *what* the min and max..."

DanHickstein:
Okay, so this is an awesome heroic PR. The clarification of "origin" versus "center" is great, and the (row, columns) consistency is really important. And, you fixed a bunch of style mistakes and broken article links along the way. So much good stuff!

I think that you deprecation warning is a good idea. And we can put some information about the change to (row, column) in the release notes. Otherwise, I'm not sure how to make sure that users are aware of the change. Hopefully not too many people will be confused for too long.

> abel.tools.vmi.radial_integration(IM, radial_ranges=None)
should probably have the origin parameter, as

Yeah, that seems right. 

> I guess, it would make more sense to declare it as
abel.tools.center.find_origin_by_convolution(IM, projections=False)
and describe “projections (bool) – ...”.

I agree, that seems to make more sense.

> so I have removed this claim. Was it ever working? How did it break? Does anybody need this capability (I guess, @DhrubajyotiDas might)?

I could see this being appealing for those working with flames or other similar images that lack an up/down point of symmetry. If one of the origin coordinates is None, then can we just set that coordinate to the center? Probably no need to implement right now.

> abel.tools.polar.reproject_image_into_polar() produces inconsistent angular ranges for different quadrants and half-planes. Probably needs some attention.

I'm not sure exactly what you mean. I guess open a new issue to remind us to fix this?

> The 'slice' method in abel.tools.center.find_origin() has some implementation problems.

Probably open a new issue for this one too and we can investigate. I think that @stggh is the mastermind behind the "slice" method, but I could be mistaken.

Anyway, this is such a solid PR and a great improvement for PyAbel. Let's give @stggh and @DhrubajyotiDas a few more days to look over things if they have time and then let's merge this.

MikhailRyazanov:
OK, thanks @DanHickstein. I'll make the obvious edits and open new issues for the rest.

> And we can put some information about the change to (row, column) in the release notes. Otherwise, I'm not sure how to make sure that users are aware of the change.

Yes, this is where the changelog will be useful. And if these changes are mentioned in the release notes, hopefully al interested people will notice them. :–)

Do you or your colleagues use PyAbel now? If we can test the migration process, this will help to catch potentially overlooked problems and to make recommendations for other users.

stggh:
Very nice work @MikhailRyazanov, much better than I could have done!  My cursory tests work fine. I will look at more test cases, and your code, in the next few days.

For [The 'slice' method in abel.tools.center.find_origin()](https://github.com/PyAbel/PyAbel/pull/267#issuecomment-573378527) the idea is to compare the left-side (and top) and right-side (bottom) intensity profiles, shifting one profile, for maximum overlap. The method relies on "dumb" least-squares fitting, which works best with real, structured data, and with the origin in proximity to it's correct value. For your test case there is not much for lsq-fitting to work with:
image

```python
plt.plot(IM[45:55, :].sum(axis=0)[:55], label='left')
plt.plot(IM[45:55, :].sum(axis=0)[:55:-1], label='right')
```

Of course, there could also be bugs in the implementation, but the concept is pretty straightforward.

DanHickstein:
> Do you or your colleagues use PyAbel now? 

Funny enough, by the time that PyAbel came around was the time that I finished photoelectron spectroscopy experiments. So, I've really never actually used PyAbel :)

MikhailRyazanov:
@stggh,
>  My cursory tests work fine. I will look at more test cases, and your code, in the next few days.

I'm glad to hear this and hope to hear more soon. :–)

I've looked at what the "slices" method does internally. Apparently, is has these slices in the opposite direction than your plot shows, but the actual problem is that the internal `_align(offset, sliceA, sliceB)` function, which evaluates the discrepancy, always shifts the "first" (left or top) slice, _filling the vacated portion with zeros_, so when the shift direction is positive, the points that would correspond to the center of the image are zeroed, and the profile looks really weird (and thus cannot be matched against the other slice). If I modify the code to always shift towards the image center (that is, always zero-filling the _outer_ points) by selecting which slice to shift, it seems to work much better and at least converges to correct results for both my examples.

Try to replace this:
https://github.com/PyAbel/PyAbel/blob/a541b33c88cf116f5c24e9b3e002c1058d3a383e/abel/tools/center.py#L484-L485
with this:
```
        if offset < 0:
            A = shift(sliceA, offset)
            B = sliceB
        else:
            A = sliceA
            B = shift(sliceB, -offset)
        diff = A - B
        fvec = (diff**2).sum()

        plt.plot(A, label='A')
        plt.plot(B, label='B')
        plt.legend()
        plt.title('offset = {:f}, fvec = {:f}'.format(*offset, fvec))
        plt.show()
```
and see how the fit converges. And then remove the `if`, leaving only its first branch (that is, always shift `sliceA`, as it is done now) and notice what happens.

If this modification works well with your real data too, I will include it in this PR.

And in any case we need to decide whether to just print a warning and use zero shift:
https://github.com/PyAbel/PyAbel/blob/a541b33c88cf116f5c24e9b3e002c1058d3a383e/abel/tools/center.py#L510-L512
or raise an error:
https://github.com/PyAbel/PyAbel/blob/a541b33c88cf116f5c24e9b3e002c1058d3a383e/abel/tools/center.py#L521
but consistently for both axes.

stggh:
@MikhailRyazanov your new residual function works well. 

Here is some test code, using real data:
```python
import numpy as np
import abel
from scipy.ndimage import shift

def track(im, method='slice', drc=(0, 0), **kwargs):

    print(f'{method} -------------')
    orig = abel.tools.center.find_origin(im, method=method, **kwargs)
    print(f'Ominus image origin: {orig}')

    print(f'offset:              {drc}')
    imo = shift(im, drc)

    new_origin = abel.tools.center.find_origin(imo, method=method, **kwargs)
    print(f'new origin:          {new_origin}')

    print(f'difference:          {np.subtract(new_origin, orig)}\n')

# -----------------------------------

im = np.loadtxt('O2-ANU1024.txt.bz2')
r, c = im.shape
r2 = r//2
c2 = c//2
zoom = 1024/r

drc = (-2.8, -3.3)
track(im, method='com', drc=drc)
track(im, method='convolution', drc=drc)
track(im, method='slice', drc=drc)

prof = im[r2-10: r2+10, c2:].sum(axis=0)
rr = prof.argmax()

print(f'with radial_range=({rr-int(200*zoom)}, {rr+int(200*zoom)})')
track(im, method='slice', drc=drc,
      radial_range=(rr-int(200*zoom), rr+int(200*zoom)))
```

```
python test-center-by-slice.py
com -------------
Ominus image origin: (512.5362783355695, 512.8775734596778)
offset:              (-2.8, -3.3)
new origin:          (509.7363177234742, 509.57760334510886)
difference:          [-2.79996061 -3.29997011]

convolution -------------
Ominus image origin: (512.5, 512.0)
offset:              (-2.8, -3.3)
new origin:          (509.5, 508.5)
difference:          [-3.  -3.5]

slice -------------
Ominus image origin: (512.3507839182547, 512.6404713954993)
offset:              (-2.8, -3.3)
new origin:          (506.98887126245404, 509.27928381941456)
difference:          [-5.36191266 -3.36118758]

with radial_range=(160, 560)
slice -------------
Ominus image origin: (512.3312774429054, 512.632277354586)
offset:              (-2.8, -3.3)
new origin:          (509.51144837374346, 509.2826359259542)
difference:          [-2.81982907 -3.34964143]

```
In general `com` does well! :-(

In my defence, normally, the `radial_range=(a, b)` would be specified with the `slice` call, to limit the fit to useful data.

Nice work, again, @MikhailRyazanov!

I think raising an error is better.

I like your suggested changes.

MikhailRyazanov:
@stggh,
>  My testing has not revealed any issues.

Sounds good, thanks!

> something odd happens with the citation label (generated from `make livehtml`):

As far as I can tell, **readthedocs** do(es) `make html`, which produces a slightly different format. But what is odd to you? If these numbers in parentheses, then they are backlinks and are controlled by the theme settings (I guess, we might be able to disable them, if they look disturbing).

About the modified “slice” method: does it work worse or better that the old version?

stggh:
Oh, okay, I thought the `1(1,2,3,)` was a mistake.

> does it work worse or better that the old version

I will check.

MikhailRyazanov:
> I thought the 1(1,2,3,) was a mistake.

I would say, it is a mistake made by the theme designer. :–)
In `make html` they look differently (not sure that better):

back2

But indeed, they can be disabled:

noback2

by putting a file `docutils.conf` with
```
[general]
footnote_backlinks: off
```
in the `docs` directory (apparently, there is no way to do this from `conf.py`).

I guess, we should disable them, unless somebody thinks that they are useful for something...

DanHickstein:
> I guess, we should disable them, unless somebody thinks that they are useful for something...

I think that the idea is that you can click on the reference number, see the reference, and then quickly go back to what you were reading before. It looks better when the reference is only used on one or two pages. When you have a reference that is being used on three or more pages, it looks a little bit silly. But, I see how they can be kinda cool and fun. You can disable them if you want, but versions seem fine to me.

Otherwise, are we ready to merge this? Is there still a remaining issue with the slice method?

stggh:
Sorry, for the delayed response, my home was hit by a massive hail storm last Monday 20th, skylights, and some windows shattered, flooding. The clean up has kept me (and my family) busy.  Along with the ever present fire threat, and the long drought, we are having a great summer!

No issues with the slice method, @MikhailRyazanov improvement helps handle the more extreme cases. However, in general, for ANU VMI data, `com` appears to give more reliable results, for determining the image center. Although, I have not had time to fully examine this.

DanHickstein:
Wow! Between the hail and the fire, Australia sounds like a wild place these days. I hope that the recovery is going well.

Sounds like we are good to merge this then?

MikhailRyazanov:
@stggh, thanks for testing (and best wishes for a less dangerous summer)!

I'll then do all the discussed changes, update the centering stuff in the GUI examples after @DanHickstein merges #269, and this will be ready merge.

abel/tools/polar.py
MikhailRyazanov: Corrected.

MikhailRyazanov:
OK, I did update the GUI examples (there was also problem with the `'com'` method that was called "center-of-mass" and thus did not work) and made all (?) other corrections discussed here. (I've also moved @DanHickstein's changelog entry for #269 to “Unreleased”, where it should be.)

Should be ready for merging now.

DanHickstein:
Awesome work @MikhailRyazanov!! (and thanks for fixing my mistake with the changelog). I'm merging now.

Issue #221, opened 2018-08-10
Future: simplify the code for Python3 support only


DanHickstein:
As @stggh pointed out in https://github.com/PyAbel/PyAbel/pull/219#issuecomment-411589779, support for Python2 is ending in 2020. As much as I hate to see it go, it seems that it's death is certain, and many projects have pledged to stop supporting it within the next year: http://python3statement.org/

I'm not sure exactly when this change will come to PyAbel, but when it does, it should provide an opportunity to simplify our code, since I recall a few occasions where we went out of our way to provide Python2 support.

Here is the list of changes that we should make when we drop Py2 support:

- We can load `.txt.bz2` files directly using `np.loadtxt` in the examples, basically undoing what was done in https://github.com/PyAbel/PyAbel/pull/219. 

- We can do away with `import six`

- We can get rid of the `from __future` imports that are on top of most of the .py files:
    ```python
    from __future__ import absolute_import
    from __future__ import division
    from __future__ import print_function
   from __future__ import unicode_literals
    ```

I'll amend this list as we come across more examples of simplifications that we can make when we go Py3-only.

MikhailRyazanov:
Done in PR #408!

PR #269, opened 2020-01-17
Making the GUI example work again

DanHickstein:
This PR contains a few tweaks to make the GUI example successfully run. While we should make the GUI much better, I thought that I would start by making the existing GUI example actually work, since it's currently broken.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ Unreleased
 
 v0.8.3 (2019-08-16)
 -------------------
+* Fixed the GUI examples (example_GUI.py and example_simple_GUI.py)
+  so that they work with the lastest versions of tk (#269).
 * New tools.vmi.Distributions class for extracting radial intensity and
   anisotropy distributions (PR #257).
 * Dropped PyAbel version from basex cache files (PR #260).
--- a/examples/example_GUI.py
+++ b/examples/example_GUI.py
@@ -9,10 +9,13 @@ import sys
 if sys.version_info[0] < 3:
     import Tkinter as tk
     from tkFileDialog import askopenfilename
+    import ttk
+    
 else:
     import tkinter as tk
     from tkinter.filedialog import askopenfilename
-import tkinter.ttk as ttk
+    import tkinter.ttk as ttk
+    
 import tkinter.font as tkFont
 from tkinter.scrolledtext import *
 #from ScrolledText import *
@@ -20,7 +23,7 @@ from tkinter.scrolledtext import *
 import matplotlib
 matplotlib.use('TkAgg')
 from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg,\
-                                              NavigationToolbar2TkAgg
+                                              NavigationToolbar2Tk
 from matplotlib.backend_bases import MouseEvent
 from matplotlib.figure import Figure
 from matplotlib.pyplot import imread, colorbar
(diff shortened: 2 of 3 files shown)


DanHickstein:
Good call. I made those changes and then made the examply_GUI.py mostly pep8 compliant just to test my pep8-ing skills :)

stggh:
All good. 

Since you are on a roll, don't forget to apply the same changes to `example_simple_GUI.py`  ;-)

DanHickstein:
Oh, I forgot that there was a simple GUI! I made similar changes to that one too. :)

stggh:
Thanks, good work. Both GUI examples work well, within their programming limitations.

There is a little more styling required for `example_simple_GUI.py`:
```python
[Gibbo:.+UI_fun/examples]$ pycodestyle example_simple_GUI.py 
example_simple_GUI.py:111:27: E231 missing whitespace after ','
example_simple_GUI.py:111:80: E501 line too long (100 > 79 characters)
example_simple_GUI.py:113:8: E111 indentation is not a multiple of four
example_simple_GUI.py:113:26: E231 missing whitespace after ','
example_simple_GUI.py:113:80: E501 line too long (81 > 79 characters)
example_simple_GUI.py:194:46: W605 invalid escape sequence '\p'
example_simple_GUI.py:208:21: E116 unexpected indentation (comment)
```

MikhailRyazanov:
Seems to work in Python 3 under Debian.

Python 2 cannot find some modules (they were renamed between Python 2 and Python 3). If we want to drop Python 2 support completely, then the rest (`if sys.version_info[0] < 3:`) should be removed also. Otherwise these renamed imports should go into the same version selector. Or maybe [six](https://six.readthedocs.io/) might help (I did not use it myself, but saw it among packages installed on my system, so it's probably not uncommon).

BTW, “center image” does trigger a deprecation warning. ;–) If you want to merge this PR after #267, the names there will need to be updated also (I did not, since the whole example was not working at all), otherwise I will need to update them in #267 after this is merged.

stggh:
I believe PR's are with respect to the current origin (master).

I am in favour of dropping python2 support, to allow cleaning of the header baggage for each python file:
```python
# testing strings with Python 2 and 3 compatibility
from six import string_types
```

MikhailRyazanov:
Since I don't use PyAbel myself, I don't really care about Python 2 support, but just dropping it for no good reason does not look nice...

@DanHickstein, I forgot that `six` is already used in some files, so we _can_ rely on it.

Basically, if you replace all `tkinter`-related imports (including the whole `import sys` + `if` part) with
```
from six.moves import tkinter as tk
from six.moves import tkinter_ttk as ttk
from six.moves import tkinter_font as tkFont
from six.moves import tkinter_scrolledtext as scrolledtext
from six.moves import tkinter_tkfiledialog as filedialog
```
(and change `ScrolledText()` and `askopenfilename()` calls accordingly), then everything starts working fine in both Python 2 and 3.

Dropping Python 2 later will require only removing “`from six.moves `” prefixes and replacing `_` by `.`.

DanHickstein:
Thanks guys! 

Here is an update which addresses the remaining pep8 issues (except for a few "line too long" that I don't intend to fix).

I also used the `from six.moves` code from @MikhailRyazanov, which works great. These examples seem to work with both py2 and py3 on OSX.

Regarding py2 compatibility, we can discuss more over on this issue:

https://github.com/PyAbel/PyAbel/issues/221

I don't have a good sense of what fraction of python users are still on py2. But, I would probably wait another 6 months or so before we actively remove any py2 support. (Heck, I've only been using py3 for a few months :) )

MikhailRyazanov:
Thanks, @DanHickstein! Please add something to the changelog and merge this. I'll then update the centering part in #267.

DanHickstein:
Okay, I added a sentence in the changelog about this. I'll go ahead and merge now!

PR #270, opened 2020-02-01
rBasex, a better pBasex

MikhailRyazanov:
Here is a new method for analyzing velocity-mapping images, similar to pBasex, but implemented more efficiently.

The main idea is that the Abel transform of _f_(_r_) cos<sup>_n_</sup> θ is _F<sub>_n_</sub>_(_r_) cos<sup>_n_</sup> θ, with some other radial part _F_<sub>_n_</sub> (different for different powers _n_) but the same angular dependence. And if a suitable basis is chosen for the radial part, the forward transforms _f_(_r_) → _F_<sub>_n_</sub>(_r_) can be computed analytically for all _n_. The rest is as usual in all “BASEX”-type methods.

That is, to perform a forward or inverse Abel transform of an image, it is sufficient to extract from it the radial profiles of all cos<sup>_n_</sup> θ terms, transform these radial profiles (which is much faster that transforming the whole image) and, if needed, construct the transformed image from the transformed profiles. The transformed radial profiles also directly give various radial distributions needed in VMI studies, thus if the transformed image itself is not needed, the last step can be skipped, making the analysis even faster.

An example of speed comparison with `two_point`, the fastest of current methods:
```
PyAbel benchmark run on AMD64 Family 23 Model 17 Stepping 0, AuthenticAMD

time in milliseconds

=== Basis generation ======================================================

Method              n = 513    n = 1025    n = 2049    n = 4097    n = 8193
---------------------------------------------------------------------------
rbasex                 11.7        34.8       116.7       598.9      3248.7
two_point               2.6        13.7        57.7       234.6      1019.7


=== Forward Abel transform ================================================

Method              n = 513    n = 1025    n = 2049    n = 4097    n = 8193
---------------------------------------------------------------------------
rbasex                  2.7        14.6        62.7       302.2      1167.8
rbasex(None)            1.4         6.5        31.7       132.7       516.2


=== Inverse Abel transform ================================================

Method              n = 513    n = 1025    n = 2049    n = 4097    n = 8193
---------------------------------------------------------------------------
rbasex                  2.7        14.7        63.0       302.6      1170.6
rbasex(None)            1.4         6.6        31.8       132.8       516.8
two_point               1.3         9.9        73.7       561.5      4343.1
```
(here `rbasex(None)` means that the output image is not constructed)
And, for completeness, the time that must be added to the `two_point` transform to extract radial distributions from the image:
```
order = 2, time in milliseconds
=== Full image, cached ======================================================
Method                n = 513    n = 1025    n = 2049    n = 4097    n = 8193
-----------------------------------------------------------------------------
linear
   MIN, none              1.8         7.6        31.0       125.7       498.1
```
So when only the radial distributions are needed, the rBasex transform is almost free. Otherwise, when the transformed image is also constructed, `rbasex` is competitive with `two_point` (and other fast methods) at 1 megapixel (the typical image size) and wins at larger image sizes. The best thing is that the time complexity scales only quadratically with the image radius, in contrast to the cubic scaling in all other methods (except maybe Hansen–Law).

Now let's see how the documentation will be built for this PR...

(More testing, proofreading and fact-checking about pBasex are still neded.)

### DIFF
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -474,7 +474,7 @@ class AbelTiming(object):
         # benchmark the basis generation (default parameters)
         def gen_basis():
             rbasex.cache_cleanup()
-            rbasex.get_bs_cached(self.w)
+            rbasex.get_bs_cached(self.w - 1)  # Rmax = half-width - 1
         self._benchmark('bs', 'rbasex',
                         gen_basis)
 
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -9,9 +9,10 @@ import abel
 from . import basex
 from . import dasch
 from . import direct
-from . import linbasex
 from . import hansenlaw
+from . import linbasex
 from . import onion_bordas
+from . import rbasex
 from . import tools
 
 from timeit import default_timer as timer
@@ -203,6 +204,7 @@ class AbelTiming(object):
         ])
         need_whole = frozenset([
             'linbasex',
+            'rbasex',
         ])
         # all available methods (= union of the above sets)
         all_methods = need_half | need_whole
@@ -466,13 +468,37 @@ class AbelTiming(object):
     def _time_onion_peeling(self):
         self._time_dasch('onion_peeling')
 
-    @_skip((['bs', 'inverse'], 'two_point'))
-    def _time_two_point(self):
-        self._time_dasch('two_point')
+    @_skip((['bs', 'inverse', 'forward'], 'rbasex'))
+    def _time_rbasex(self):
+        # benchmark the basis generation (default parameters)
+        def gen_basis():
+            rbasex.cache_cleanup()
+            rbasex.get_bs_cached(self.w)
+        self._benchmark('bs', 'rbasex',
+                        gen_basis)
+
+        # benchmark all transforms
+        for direction in ['inverse', 'forward']:  # (default first)
+            # warm-up run to cache "distributions" (basis is already cached)
+            rbasex.rbasex_transform(self.whole_image)
+            # benchmark the transform
+            self._benchmark(direction, 'rbasex',
+                            rbasex.rbasex_transform,
+                            self.whole_image)
+
+            # discard the unneeded transform matrix
+            basex.cache_cleanup(direction)
+
+        # discard all caches
+        rbasex.cache_cleanup()
 
     @_skip((['bs', 'inverse'], 'three_point'))
     def _time_three_point(self):
         self._time_dasch('three_point')
+
+    @_skip((['bs', 'inverse'], 'two_point'))
+    def _time_two_point(self):
+        self._time_dasch('two_point')
 
     # (End of benchmarking functions.)
 
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -468,7 +468,8 @@ class AbelTiming(object):
     def _time_onion_peeling(self):
         self._time_dasch('onion_peeling')
 
-    @_skip((['bs', 'inverse', 'forward'], 'rbasex'))
+    @_skip(('bs', 'rbasex'),
+           (['inverse', 'forward'], ['rbasex', 'rbasex(None)']))
     def _time_rbasex(self):
         # benchmark the basis generation (default parameters)
         def gen_basis():
@@ -485,6 +486,10 @@ class AbelTiming(object):
             self._benchmark(direction, 'rbasex',
                             rbasex.rbasex_transform,
                             self.whole_image)
+            # same without output image
+            self._benchmark(direction, 'rbasex(None)',
+                            rbasex.rbasex_transform,
+                            self.whole_image, out=None)
 
             # discard the unneeded transform matrix
             basex.cache_cleanup(direction)
--- a/abel/rbasex.py
+++ b/abel/rbasex.py
@@ -107,8 +107,8 @@ def rbasex_transform(IM, origin='center', rmax='MIN', order=2, odd=False,
             channels. Not implemented for odd orders > 1.
 
             Notice that this method is nonlinear, which also means that it is
-            considerably slower than the linear methods and might produce
-            slightly biased results.
+            considerably slower than the linear methods and the transform
+            operator cannot be cached.
 
         In all cases, `strength` = 0 provides no regularization. For the
         Tikhonov methods, `strength` ~ 100 is a reasonable value for megapixel
@@ -134,7 +134,7 @@ def rbasex_transform(IM, origin='center', rmax='MIN', order=2, odd=False,
             half for ``odd=True``
         ``None``:
             no image (**recon** will be ``None``). Can be useful to avoid
-            unnecessary calculations if only the transformed radial
+            unnecessary calculations when only the transformed radial
             distributions (**distr**) are needed.
 
     Returns
@@ -374,7 +374,7 @@ def get_bs_cached(Rmax, order=2, odd=False, direction='inverse', reg=None):
         include odd angular orders
     direction : str: ``'forward'`` or ``'inverse'``
         type of Abel transform to be performed
-    reg : None or tuple (str, float)
+    reg : None or str or tuple (str, float)
         regularization type and strength for inverse transform
     Returns
     -------
--- a/abel/rbasex.py
+++ b/abel/rbasex.py
@@ -220,7 +220,7 @@ def get_bs_cached(Rmax, direction='inverse'):
         return _tri
 
 
-def cache_cleanup():
+def cache_cleanup(select='all'):
     """
     Utility function.
 
@@ -230,7 +230,15 @@ def cache_cleanup():
 
     Parameters
     ----------
-    None
+    select : str
+        selects which caches to clean:
+
+        ``all`` (default)
+            everything, including basis;
+        ``forward``
+            forward transform;
+        ``inverse``
+            inverse transform.
 
     Returns
     -------
@@ -238,9 +246,12 @@ def cache_cleanup():
     """
     global _prm, _dst, _bs_prm, _bs, _trf, _tri
 
-    _prm = None
-    _dst = None
-    _bs_prm = None
-    _bs = None
-    _trf = None
-    _tri = None
+    if select == 'all':
+        _prm = None
+        _dst = None
+        _bs_prm = None
+        _bs = None
+    if select in ('all', 'forward'):
+        _trf = None
+    if select in ('all', 'inverse'):
+        _tri = None
--- a/abel/rbasex.py
+++ b/abel/rbasex.py
@@ -4,7 +4,7 @@ from __future__ import division
 from __future__ import print_function
 
 import numpy as np
-from scipy.linalg import inv
+from scipy.linalg import solve_triangular
 
 from abel.tools.vmi import Distributions
 from abel.tools.symmetry import put_image_quadrants
@@ -36,23 +36,23 @@ from abel.tools.symmetry import put_image_quadrants
 
 
 # Caches and their parameters
-_prm = None  # [shape, origin, rmax, weights]
+_prm = None  # [shape, origin, rmax, order, odd, weights]
 _dst = None  # Distribution object
-_bs_prm = None  # [Rmax]
-_bs = None  # [P0, P2] — projected functions
-_trf = None  # [Af0, Af2] — forward transform matrices
-_tri = None  # [Ai0, Ai2] — inverse transform matrices
+_bs_prm = None  # [Rmax, order, odd]
+_bs = None  # [P[n]] — projected functions
+_trf = None  # [Af[n]] — forward transform matrices
+_tri = None  # [Ai[n]] — inverse transform matrices
 
 
-def rbasex_transform(IM, origin='center', rmax='MIN',
+def rbasex_transform(IM, origin='center', rmax='MIN', order=2, odd=False,
                      weights=None, direction='inverse'):
     """
     This function takes the input image and outputs its forward or inverse Abel
     transform as an image and its radial distributions.
 
-    The **origin**, **rmax** and **weights** parameters are passed to
-    :class:`abel.tools.vmi.Distributions`, so see its documentation for their
-    detailed descriptions.
+    The **origin**, **rmax**, **order**, **odd** and **weights** parameters are
+    passed to :class:`abel.tools.vmi.Distributions`, so see its documentation
+    for their detailed descriptions.
 
     Parameters
     ----------
@@ -63,6 +63,10 @@ def rbasex_transform(IM, origin='center', rmax='MIN',
         string
     rmax : int or string
         largest radius to include in the transform
+    order : int
+        highest order in the angular distributions, ≥ 0
+    odd : bool
+        include odd angular orders (enabled automatically if **order** is odd)
     weights : m × n numpy array, optional
         weighting factors for each pixel. The array shape must match the image
         shape. Parts of the image can be excluded from analysis by assigning
@@ -79,27 +83,41 @@ def rbasex_transform(IM, origin='center', rmax='MIN',
         the object from which various distributions for the transformed image
         can be retrieved, see :class:`abel.tools.vmi.Distributions.Results`
     """
+    if order == 0:
+        odd = False  # (to eliminate additional checks)
+    elif order % 2:
+        odd = True  # enable automatically for odd orders
 
-    p0, p2 = _profiles(IM, origin, rmax, weights)
+    # extract radial profiles from input image
+    p = _profiles(IM, origin, rmax, order, odd, weights)
 
     Rmax = _dst.rmax
 
-    A0, A2 = get_bs_cached(Rmax, direction)
+    # get appropriate transform matrices
+    A = get_bs_cached(Rmax, order, odd, direction)
 
-    c0 = A0.dot(p0)
-    c2 = A2.dot(p2)
+    # transform radial profiles
+    c = [An.dot(pn) for An, pn in zip(A, p)]
 
-    distr = Distributions.Results(np.arange(Rmax + 1), np.array([c0, c2]),
-                                  2, False)
-    Q = _image(c0, c2)[::-1]
-    height, width = _dst.Qheight, _dst.Qwidth
-    recon = put_image_quadrants((Q, Q, Q, Q),
-                                (2 * height - 1, 2 * width - 1))
+    # construct output (transformed) distributions
+    distr = Distributions.Results(np.arange(Rmax + 1), np.array(c), order, odd)
+
+    # construct output (transformed) image from transformed radial profiles
+    if odd:
+        Q = _image(c)  # actually right half
+        # combine with left half (mirrored without central column)
+        recon = np.hstack((Q[:, :0:-1], Q))
+    else:
+        Q = _image(c)[::-1]  # flip to upper right (Q0)
+        # assemble full image
+        height, width = _dst.Qheight, _dst.Qwidth
+        recon = put_image_quadrants((Q, Q, Q, Q),
+                                    (2 * height - 1, 2 * width - 1))
 
     return recon, distr
 
 
-def _profiles(IM, origin, rmax, weights):
+def _profiles(IM, origin, rmax, order, odd, weights):
     """
     Get radial profiles of cos^n theta terms from the input image.
     """
@@ -108,75 +126,92 @@ def _profiles(IM, origin, rmax, weights):
     # image
     global _prm, _dst
 
-    prm = [IM.shape, origin, rmax, weights]
+    prm = [IM.shape, origin, rmax, order, odd, weights]
     if _prm != prm:
         _prm = prm
-        _dst = Distributions(origin=origin, rmax=rmax, weights=weights,
-                             use_sin=False, method='linear')
+        _dst = Distributions(origin=origin, rmax=rmax, order=order, odd=odd,
+                             weights=weights, use_sin=False, method='linear')
 
     return _dst(IM).cos()
 
 
-def _image(c0, c2):
+def _image(c):
     """
-    Create transformed image from its cos^n theta radial profiles.
+    Create transformed image (lower right quadrant for even-only,
+    right half for odd) from its cos^n theta radial profiles.
     """
-    rbin, wl, wu, cos2 = _dst.bin, _dst.wl, _dst.wu, _dst.c[1]
+    rbin, wl, wu, cos = _dst.bin, _dst.wl, _dst.wu, _dst.c
 
-    c0l = np.append(c0, 0)
-    c2l = np.append(c2, 0)
-    c0u = np.append(c0[1:], [0, 0])
-    c2u = np.append(c2[1:], [0, 0])
+    # 0th order (isotropic)
+    IM = (wl * np.append(c[0], [0])[rbin] +  # lower bins
+          wu * np.append(c[0][1:], [0, 0])[rbin])  # upper bins
+    # add all other orders
+    for cn, cosn in zip(c[1:], cos[1:]):
+        IM += (wl * np.append(cn, [0])[rbin] +
+               wu * np.append(cn[1:], [0, 0])[rbin]) * cosn
+    # (weighting for each order is somehow faster than processing lower and
+    #  upper bins separately and then combining)
 
-    return wl * (c0l[rbin] + c2l[rbin] * cos2) + \
-           wu * (c0u[rbin] + c2u[rbin] * cos2)
+    return IM
 
 
-def _bs_rbasex(Rmax):
+def _bs_rbasex(Rmax, order, odd):
     """
     Compute radial parts of basis projections for R and radii up to Rmax.
     """
-    def psqrt(x):
-        return np.sqrt(x, out=np.zeros_like(x), where=x > 0)
+    # all needed orders (even or all) from 0 to order
+    orders = range(0, order + 1, 1 if odd else 2)
 
-    def plog(x):
-        return np.log(x, out=np.zeros_like(x), where=x > 0)
+    # list of matrces for projections: P[i][R, r] = p_{R:n_i}(r),
+    # sampled at r = 0, ..., R, for R = 0, ..., Rmax and orders n
+    P = [np.eye(Rmax + 1, order='F') for n in orders]  # ('F' is faster here)
+    # (p_{0;0}(0) = 1 indeed, but p_{0;n}(0) = 0 for all other orders,
+    #  so P[i][0, 0] are also set to 1 to make P[i] nondegenerate)
 
-    def F1(n, r, z):
-        if n == 0:
-            return z
-        elif n == 2:
-            return r * np.arctan2(z, r)
+    # fill p_{R>0;0}(0) = 2 (all other p_{R>0;n}(0) = 0 already)
+    P[0][1:, 0] = 2
 
-    def Frho(n, r, z):
-        rho = np.sqrt(r**2 + z**2)
-        if n == 0:
-            return (z * rho + r**2 * plog(z + rho)) / 2
-        elif n == 2:
-            return r**2 * plog(z + rho)
+    # fill all other r > 0 columns (only functions with R >= r are non-zero)
+    for r in range(1, Rmax + 1):
+        # all needed R - 1, R, R + 1 for current r
+        R = np.arange(r - 1, Rmax + 2, dtype=float)
 
-    def Z(R, r):
-        return psqrt(R**2 - r**2)
+        # rho = max(r, R)
+        rho = R.copy()
+        rho[0] = r
 
-    def p(R, n, r):
-        ZR = Z(R, r)
-        ZRm1 = Z(R - 1, r)
-        ZRp1 = Z(R + 1, r)
-        return 2 * (2 * (Frho(n, r, ZR) - R * F1(n, r, ZR)) +
-                    (R - 1) * F1(n, r, ZRm1) - Frho(n, r, ZRm1) +
-                    (R + 1) * F1(n, r, ZRp1) - Frho(n, r, ZRp1))
+        # since z = sqrt(R^2 - r^2) for R >= r, otherwise 0,
+        # it is z = sqrt(max(r, R)^2 - r^2) = sqrt(rho^2 - r^2)
+        z = np.sqrt(rho**2 - r**2)
 
-    r = np.arange(Rmax + 1, dtype=float)
-    P0 = np.empty((Rmax + 1, Rmax + 1))
-    P2 = np.empty((Rmax + 1, Rmax + 1))
-    for R in range(Rmax + 1):
-        P0[R] = p(R, 0, r)
-        P2[R] = p(R, 2, r)
+        f = r / rho  # "f" means "fraction r / rho"
 
-    return P0.T, P2.T
+        rln = r * np.log(z + rho)
+
+        # define F_n for all needed n
+        F = {-1: (z / f + rln) / 2, 0: z}
+        if order >= 1:
+            F[1] = rln
+        if order >= 2:
+            F[2] = r * np.arccos(f)
+        if order >= 3:
+            F[3] = z * f
+        if order >= 4:
+            fn = f.copy()  # current (r / rho)^n
+            for n in range(2, order - 1):  # from 4 - 2 to order - 2
+                fn *= f
+                F[n + 2] = (z * fn + (n - 1) * F[n]) / n
+
+        # compute p_{R;n}(r) for all needed R and n
+        for i, n in enumerate(orders):
+            rFRF = r * F[n - 1] - R * F[n]
+            P[i][r:, r] = 2 * (2 * rFRF[1:-1] - rFRF[2:] - rFRF[:-2])
+            #    R >= r                 at R         R - 1      R + 1
+
+    return P
 
 
-def get_bs_cached(Rmax, direction='inverse'):
+def get_bs_cached(Rmax, order=2, odd=False, direction='inverse'):
     """
     Internal function.
 
@@ -188,21 +223,25 @@ def get_bs_cached(Rmax, direction='inverse'):
     ----------
     Rmax : int
         largest radius to be transformed
+    order : int
+        highest angular order
+    odd : bool
+        include odd angular orders
     direction : str: ``'forward'`` or ``'inverse'``
         type of Abel transform to be performed
 
     Returns
     -------
-    A : tuple of 2D numpy arrays
+    A : list of 2D numpy arrays
         (**Rmax** + 1) × (**Rmax** + 1) matrices of the Abel transform (forward
         or inverse) for each angular order
     """
     global _bs_prm, _bs, _trf, _tri
 
-    prm = [Rmax]
+    prm = [Rmax, order, odd]
     if _bs is None or _bs_prm != prm:
         _bs_prm = prm
-        _bs = _bs_rbasex(Rmax)
+        _bs = _bs_rbasex(Rmax, order, odd)
         _trf = None
         _tri = None
 
@@ -212,11 +251,12 @@ def get_bs_cached(Rmax, direction='inverse'):
         return _trf
     else:  # 'inverse'
         if _tri is None:
-            A0, A2 = _bs
-            A2[0, 0] = 1  # (to avoid degeneracy)
-            Ai0 = inv(A0)
-            Ai2 = inv(A2)
-            _tri = (Ai0, Ai2)
+            # P[n] are triangular, thus can be inverted faster than general
+            # matrices, however, NumPy/SciPy do not have such functions;
+            # nevertheless, solve_triangular() is ~twice faster than inv()
+            # (and ...(Pn, I, lower=True).T is faster than ...(Pn.T, I))
+            I = np.eye(Rmax + 1)
+            _tri = [solve_triangular(Pn, I, lower=True).T for Pn in _bs]
         return _tri
 
 
--- a/abel/rbasex.py
+++ b/abel/rbasex.py
@@ -3,13 +3,16 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import os.path
+from os import listdir
+import re
+
 import numpy as np
 from scipy.linalg import inv, solve_triangular, svd, pascal, invpascal
 from scipy.optimize import nnls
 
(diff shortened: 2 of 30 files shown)


MikhailRyazanov:
Addendum: some small changes to the `abel.tools.vmi.Distributions` class are also included. Some are directly related to `rbasex`, and some are just usability/consistency improvements.

stggh:
This is neat!  I will examine the method for some VMI images that have an ion-beam (MCP depleted gain) hole.

DanHickstein:
This is super rad! I will proofread the code more when time allows.

For now, I'm going to close and re-open this PR to see if I can trigger the readthedocs check.

MikhailRyazanov:
@DanHickstein, great! Now the docs are available.

> I will proofread the code more when time allows.

Proofread the _documentation_, please! :–) Grammar, awkward wording and if something isn't clear...

In the meantime I'll think about adding more unit tests (basis, caching, image shape and that higher orders work correctly).

MikhailRyazanov:
@DanHickstein and @stggh, if you have time, please review this PR...

stggh:
Away 10/3-20/3 in New Zealand (North island). The trip was originally planned for Hokkaido (Japan), but cancelled due to the State of Emergency there. 

I have been running some of my VMI data that have "holes" due to MCP ion-beam gain depletion, but haven't yet come across an example that demonstrates the power of your method, or caused a problem. Your code and methods are very good.

I will review on my return, but otherwise, happy to accept this PR if you feel the delay is too long.

MikhailRyazanov:
@stggh, this is not urgent, so looking at it after your return is fine.

The ability to analyze data with “holes” is just a bonus; you'll probably need very poor images (with a hole disturbing a significant angular range) to see a clear difference in real data (but it would be interesting if you can find an example!). The main benefit of this method is that it produces clean images, like lin-BASEX, but actually even cleaner and works faster. Plus, it is a high-resolution substitution for pBasex for those who want it.

MikhailRyazanov:
ping

stggh:
Thanks for the reminder! 

I have been caught up with 2 weeks home isolation, following our return from NZ, which might appear ideal for providing time for a code review, except I caught a flu. The associated COVID-19 test was negative. 

I like your excellent synopsis:
> The main benefit of this method is that it produces clean images, like lin-BASEX, but actually even cleaner and works faster. Plus, it is a high-resolution substitution for pBasex for those who want it.

I will get on to it!

DanHickstein:
This looks awesome Mikhail! Sorry to take so long to review it. I should have some time over the next few days.

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+            corresponds to no interference between different multiphoton
+            channels. Not implemented for odd orders > 1.
+
+            Notice that this method is nonlinear, which also means that it is
+            considerably slower than the linear methods and the transform
+            operator cannot be cached.
+
+        In all cases, `strength` = 0 provides no regularization. For the
+        Tikhonov methods, `strength` ~ 100 is a reasonable value for megapixel
+        images. For truncated SVD, `strength` must be < 1; `strength` ~ 0.1 is
```
DanHickstein, line 122: For the SVD method, is the effect of the `strength` independent of image size? For the Tikhonov methods, should we try to find a way to scale the strength of the regularization so that the effect is independent of image size? It might be something simply like simply multiplying the strength by the image size...  It would be nice if we could just recommend a strength and it would more-or-less work for all images.

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+        a reasonable value; `strength` ~ 0.5 can produce noticeable ringing
+        artifacts.
+    out : str or None
+        shape of the output image:
+
+        ``'same'`` (default):
+            same shape and origin as the input
+        ``'fold'`` (fastest):
+            Q0 quadrant (for ``odd=False``) or right half (for ``odd=True``) up
```
DanHickstein, line 131: change "Q0" to "Q0 (upper left)"?

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+            to **rmax**, but limited to the largest input-image quadrant (or
+            half)
+        ``'unfold'``:
+            like ``'fold'``, but symmetrically “unfolded”
```
DanHickstein, line 135: I'm not really sure what this means. Basically, you calculate just Q0 and then duplicate it to all of the other quadrants? Isn't this what the transform always does? Or does it normally allow for asymmetry in the top/down and left/right dimensions?

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+    for their detailed descriptions.
+
+    Parameters
+    ----------
+    IM : m × n numpy array
+        the image to be transformed
+    origin : tuple of int or str
+        image origin, explicit in the (row, column) format, or as a location
+        string
+    rmax : int or string
```
DanHickstein, line 74: What string values can be entered? Case sensitive?

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+        largest radius to include in the transform
```
DanHickstein, line 75: Say what the default value is.

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+    order : int
+        highest angular order present in the data, ≥ 0
```
DanHickstein, line 77: Say what the default value is.

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+    odd : bool
+        include odd angular orders (enabled automatically if **order** is odd)
```
DanHickstein, line 79: Default is False.

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+    weights : m × n numpy array, optional
+        weighting factors for each pixel. The array shape must match the image
+        shape. Parts of the image can be excluded from analysis by assigning
+        zero weights to their pixels.
```
DanHickstein, line 83: "The default is None, which applies equal weight to all pixels."

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+    direction : str: ``'forward'`` or ``'inverse'``
+        type of Abel transform to be performed
```
DanHickstein, line 85: default is 'inverse'

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+    reg : None or str or tuple (str, float), optional
+        regularization to use for inverse Abel transform. ``None`` means no
```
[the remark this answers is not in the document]
DanHickstein: "default is none"

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+            half for ``odd=True``
+        ``None``:
+            no image (**recon** will be ``None``). Can be useful to avoid
+            unnecessary calculations when only the transformed radial
+            distributions (**distr**) are needed.
+    basis_dir : str, optional
+        path to the directory for saving / loading the basis set (useful only
+        for the inverse transform without regularization; time savings in other
+        cases are small and might be negated by the disk-access overhead).
+        If ``None``, the basis set will not be loaded from or saved to disk.
```
DanHickstein, line 149: Default is None.

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+    verbose : bool
+        print information about processing (for debugging)
```
DanHickstein, line 151: default is False.

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+
+    Returns
+    -------
+    recon : 2D numpy array or None
+        the transformed image. Is centered and might have different dimensions
+        than the input image.
+    distr : Distributions.Results object
```
DanHickstein, line 158: This is fundamentally different from all of the other `_transform` functions, which only return the image array. I see why it needs to work this way, but it's potentially confusing.

DanHickstein:
The comparison with the other methods looks nice! 

image

https://gist.github.com/DanHickstein/5874a3bee2dc0eb37cc724049937fcc4

DanHickstein:
Wow. I am more and more impressed with this work. I was going to ask if you were going to write a paper about it, but it looks like you've already written one! :)  https://pyabel--270.org.readthedocs.build/en/270/transform_methods/rbasex-math.html#rbasexmath

MikhailRyazanov:
Thanks, Dan! I'll wait for you to finish your review (please also make any corrections to the text!) and for Steve's review (@stggh, it's sad that you've got a flu, but hopefully it is less dangerous than this new coronavirus. Get well soon!), and then I'm going to make the edits.

> I was going to ask if you were going to write a paper about it, but it looks like you've already written one! :)

Well, this was a practice for a real one, which I hope to publish eventually (about the original method and this faster version). When I get time to work on it...

stggh:
Amazing and very impressive work @MikhailRyazanov!  

The only issue that has appeared in my analysis is a factor of <math>&pi;</math> between the radial intensity yielded by `.distr.rIbeta()` and that from `.angular_integration`, no doubt a fault  with the latter :-(

```python
im = np.loadtxt('O2-ANU1024.txt.bz2')
imc = abel.tools.center.center_image(im, method='slice')

rbasex = abel.Transform(imc, method='rbasex', angular_integration=True)

rb, rp, rp1 = rbasex.distr.rIbeta()
radial, pes = rbasex.angular_integration
```
Here is the O<sub>2</sub><sup>-</sup> PES, that includes the factor <math>1/&pi;</math> for the `distr` intensity (blue line).
Screen Shot 2020-04-03 at 10 51 05
The 'extra' intensity is due to the better resolution of the basis decomposition.
Screen Shot 2020-04-03 at 10 58 03

The same &pi; scaling factor applies for other species, e.g. C<sub>2</sub>H<sup>-</sup>:
Screen Shot 2020-04-03 at 10 55 34

abel/rbasex.py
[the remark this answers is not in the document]
stggh: `reg='L2'` i.e. string only fails
```python
    transform_options=dict(reg='L2'))
       :
   ValueError: Wrong regularization format "L2"
```

abel/rbasex.py
stggh: `distr` is encapsulates `(r)basex` results, much like the `Transform` class contains more than just the Abel transformed image. It is neat and tidy, and e.g. should be used for `linbasex` to contain the `Beta` arrays unique to it.

abel/benchmark.py
```diff
@@ -9,9 +9,10 @@
 from . import basex
 from . import dasch
 from . import direct
-from . import linbasex
 from . import hansenlaw
+from . import linbasex
 from . import onion_bordas
+from . import rbasex
```
stggh, line 15: `rbasex` has a nice fast execution speed!

examples/example_rbasex_multimass.py
```diff
@@ -0,0 +1,131 @@
...
+plt.title('Distributions: right image')
+_, distr = rbasex_transform(im, origin=origin3, rmax=r3, weights=mask3, out=None)
+r, I, beta = distr.rIbeta()
+plt.plot(r, I, c=rgb(2), label='$I(r)$')
+plt.plot(r, I * beta, c='gray', label='$\\beta(r) \\cdot I(r)$')
+plt.legend()
+plt.autoscale(enable=True, tight=True)
+
+plt.tight_layout()
+plt.show()
```
stggh, line 131: Impressive!

abel/rbasex.py
MikhailRyazanov: I don't actually know. For SVD, I use a fraction instead of an absolute number, since it, in theory, should be less size-dependent. But it still has somewhat different effects for different image sizes. The “L2” method essentially folows the BASEX approach, and its `strength` actually gives effects on velocity distributions very similar to that of BASEX `reg`. So:
1. This question of the optimal regularization strength needs deeper research. I have seen some attempts in the literature, but no definitive answer.
2. So if we want to give more specific recommendations, they perhaps deserve a separate documentation section (with references, examples and so on).
3. I actually have some examples in `transform_methods/rbasex-math.html#examples`, but if you take `rbasex-regRMSE.py` and try it with different images and even different random samples of the _same_ image, you'll see that the optimal parameters will be different. (Hence there is a big “Warning” preceding the examples.)
3. If we want to introduce some automatic scaling to reduce the size dependence, it probably should be done for BASEX as well, for consistency.

abel/rbasex.py
MikhailRyazanov: Like lin-BASEX, it assumes that the image is composed of spherical harmonics, so for even-only orders it is totally symmetric, for even-only orders is horizontally symmetric and vertically antisymmetric, and for mixed orders is just horizontally symmetric. In principle, you can transform each quadrant independently, but this will yield 4 different velocity distributions. Averaging them will give the same results as transforming the whole image, so this is not very useful.
The transform itself actually produces only the velocity distribution. The transformed image is then created from that distribution in the form requested (internally, yes, just Q0 of right half is calculated and, if needed, is mirrored to other parts).

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+    The **origin**, **rmax**, **order**, **odd** and **weights** parameters are
+    passed to :class:`abel.tools.vmi.Distributions`, so see its documentation
+    for their detailed descriptions.
+
+    Parameters
+    ----------
+    IM : m × n numpy array
+        the image to be transformed
+    origin : tuple of int or str
+        image origin, explicit in the (row, column) format, or as a location
```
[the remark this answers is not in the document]
MikhailRyazanov: There is a paragraph “The origin, rmax, order, odd and weights parameters are passed to abel.tools.vmi.Distributions, so see its documentation for their detailed descriptions.” (with a hyperlink) just above the “Parameters” section. :–)

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+
+    The **origin**, **rmax**, **order**, **odd** and **weights** parameters are
+    passed to :class:`abel.tools.vmi.Distributions`, so see its documentation
+    for their detailed descriptions.
+
+    Parameters
+    ----------
+    IM : m × n numpy array
+        the image to be transformed
+    origin : tuple of int or str
```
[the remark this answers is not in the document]
MikhailRyazanov: OK.

abel/rbasex.py
MikhailRyazanov: Upper right actually.

abel/rbasex.py
MikhailRyazanov: There is a paragraph “The origin, rmax, order, odd and weights parameters are passed to abel.tools.vmi.Distributions, so see its documentation for their detailed descriptions.” (with a hyperlink) just above the “Parameters” section. :–)

abel/rbasex.py
MikhailRyazanov: It is also described in `abel.tools.vmi.Distributions`. Do you want to include it here as well? (I guess, something like “by default, the largest radius fitting in at least one full quadrant”.)

abel/rbasex.py
MikhailRyazanov: Yes, this is very much like in `linbasex`, except that `rbasex` always returns 2 values instead of 1 or 3 depending on input parameters.

abel/rbasex.py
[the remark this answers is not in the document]
MikhailRyazanov: I have changed `'L2'` to `('L2', strength)` and so on. This should be more clear, I guess.

abel/rbasex.py
```diff
@@ -0,0 +1,751 @@
...
+    weights : m × n numpy array, optional
+        weighting factors for each pixel. The array shape must match the image
+        shape. Parts of the image can be excluded from analysis by assigning
+        zero weights to their pixels.
+    direction : str: ``'forward'`` or ``'inverse'``
+        type of Abel transform to be performed
+    reg : None or str or tuple (str, float), optional
+        regularization to use for inverse Abel transform. ``None`` means no
+        regularization, a string selects a non-parameterized regularization
+        method, and parameterized methods are selected by a tuple (`method`,
```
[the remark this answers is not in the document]
MikhailRyazanov: I have changed 'L2' to ('L2', strength) and so on. This should be more clear, I guess.

abel/rbasex.py
MikhailRyazanov: Added a hyperlink to the end of that paragraph.

abel/rbasex.py
MikhailRyazanov: Done

abel/rbasex.py
MikhailRyazanov: Done.

abel/rbasex.py
MikhailRyazanov: Done.

abel/rbasex.py
MikhailRyazanov: Done.

abel/rbasex.py
MikhailRyazanov: Done.

abel/rbasex.py
MikhailRyazanov: Done.

abel/rbasex.py
MikhailRyazanov: Done.

examples/example_rbasex_multimass.py
MikhailRyazanov: :–)

MikhailRyazanov:
@stggh, we have discussed the π factor in #257. Basically, `rbasex` is based on `Distributions`, so it has all the same properties. My _I_(_r_) follows the usual convention that 4π ∫ _I_(_r_) _r_<sup>2</sup>_dr_ equals the total intensity (the number of events). I don't remember what exactly `angular_integration` calculates and why. But at least the reconstructed images are consistent with other methods, so you can use either `Distributions` or `angular_integration` in any case. :–)

Can you please do C<sub>2</sub>H<sup>−</sup> with `reg='pos'` and show the low-intensity portion of that plot (say, up to 0.2e5 or 0.1e5)?

stggh:
> #257 

 Oh, that is right, thanks for the reminder.

> you can use either `Distributions` or `angular_integration` in any case

 Yes, this works very well.

> C<sub>2</sub>H<sup>−</sup> with reg='pos' 

Note in this example`reg='pos'` affects both `Distributions` and `angular_integration`. `reg` produces some nice smoothing.

Screen Shot 2020-04-05 at 09 24 25

Screen Shot 2020-04-05 at 09 25 05

I will send you the (unpublished) VMI image data via Cloudstor, in case you want to play without the delay of waiting for me to run your nice `rbasex` analysis.

MikhailRyazanov:
Thanks for sending the data! Now I see that the negative dip at small radii is due to a dark spot near the image center:
C2H-center
If I mask it, together with the strange contamination in the lower left quadrant, and the strange vertical line (just in case), as described [here](https://pyabel--270.org.readthedocs.build/en/270/transform_methods/rbasex.html#how-to-use-it):
C2H-mask
I get more reasonable intensity behavior near the center in both cases — no significant negative intensities beyond the noise level without regularization, and no suspicious hard clamping to zero with the positive regularization:
C2H-plot

In any case, it is interesting to see (more obviously in your plots) how the positivity constraints clean up the 300–360 region, such that the peak near 320 becomes apparent. Do you know whether it is real?

stggh:
Excellent! Thanks for taking a look at the data. The centre line is due to the two CMOS sensors of the camera. The MCP gain depletion results from the ion-beam, in our coaxial spectrometer, which is present during set-up, but switched out, when collecting photoelectrons, although, the gain depletion remains.

You make some good observations, thanks to the (r)basex regularization analysis.

Off-hand I cannot easily say anything about the possibility of a transition near 320 radius. There are many transitions in this region, and it is heavily perturbed. However, I will take a look, in relation to the assignments that have been made. It may take a little time.

doc/transform_methods/rbasex.rst
```diff
@@ -0,0 +1,211 @@
...
+
+1. Triangular radial basis functions are used instead of Gaussians. They are
+   more compact/orthogonal (only the adjacent functions overlap) and have
+   analytical Abel transforms.
+
+2. Cosine powers are used instead of Legendre polynomials for angular basis
+   functions. This makes the projected basis functions also separable into
+   radial and angular parts.
+
+3. The basis separability allows decomposition of the problem in two steps:
```
DanHickstein, line 62: I would change the first part to read: "first, radial distributions are extracted without
   intermediate rebinning to polar grid (which is faster and avoids accumulation
   of resampling errors)"

DanHickstein: Most important is to fix the "what" -> "which"

doc/transform_methods/rbasex.rst
```diff
@@ -0,0 +1,211 @@
...
+3. Only the non-negativity constraints are implemented. However, several
+   linear regularization options are added.
+
+4. Odd angular orders can be included.
+
+
+When to use it
+--------------
+
+Although this method can be applied only to velocity-map images, and not just
```
DanHickstein, line 99: There might be other situations where the data looks close enough to a VMI image so that rBasex might be appropriate. So, I'd suggest that we soften the language to something like:

"The rBasex method makes additional assumptions about the data (beyond cylindrical symmetry), so is most applicable for the transform of velocity-map images. In this special case, if offers several benefits:"

doc/transform_methods/rbasex.rst
```diff
@@ -0,0 +1,211 @@
...
+cylindrically symmetric images, it has several benefits in this special case.
+
+1. The reconstructed radial distributions, which are of the primary interest in
```
DanHickstein, line 102: "which are *often* the primary interest".

abel/rbasex.py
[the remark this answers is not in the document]
MikhailRyazanov: I don't know. I think that it is better to keep the API documentation and the method description separated as much as possible. @stggh also said once that he reads these docstrings within the interactive interpreter session, so it's better to keep them readable as plain text and, I guess, concise. Plus, I hope that people read “How it works/When to use it/How to use it” instead of randomly playing with the parameters (or at least before that). So it is probably better to add more information there (this applies to other methods as well; for example, our BASEX documentation tells almost nothing about regularization).

abel/rbasex.py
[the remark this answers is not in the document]
MikhailRyazanov: The [new code](https://github.com/MikhailRyazanov/PyAbel/blob/bd456a171e9810030be2bec8c4454cea57c3352d/abel/rbasex.py#L89-L110) and [how it renders](https://pyabel--270.org.readthedocs.build/en/270/abel.html#abel.rbasex.rbasex_transform).

doc/transform_methods/rbasex.rst
```diff
@@ -0,0 +1,211 @@
...
+rBasex
+======
+
+
+Introduction
+------------
+
+This method resembles the pBasex [1]_ approach of expanding a velocity-map
+image over a 2D basis set in polar coordinates, but uses more convenient basis
+functions with analytical Abel transforms, developed by M. Ryazanov in [2]_.
```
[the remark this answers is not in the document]
MikhailRyazanov: OK, removing “in” should indeed work.

doc/transform_methods/rbasex.rst
```diff
@@ -0,0 +1,211 @@
...
+
+In velocity-map imaging with cylindrically symmetric photodissociation (in a
+broad sense, including photoionization and photodetachment) the 3D velocity
+distribution at each speed (3D radius) consists of a finite number of spherical
+harmonics :math:`Y_{nm}(\theta, \varphi)` with :math:`m = 0`, which are also
+representable as Legendre polynomials :math:`P_n(\cos\theta)`. This means that
+an :math:`N \times N` image has only :math:`N_r \times N_a` degrees of freedom,
+where :math:`N_r` is the number of radial samples, usually :math:`N / 2`, and
+:math:`N_a` is the number of angular terms, a small number depending on the
+number of photons. These are the “radial distribution” extracted from the
```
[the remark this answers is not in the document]
MikhailRyazanov: As far as I know, sometimes the involved orbitals and other factors can lower this number to some effective order < the number of photons, but this is not always known in advance. Can this be worded better without going too far?

doc/transform_methods/rbasex.rst
```diff
@@ -0,0 +1,211 @@
...
+transforms, with the same coefficients. The reverse is also true: finding the
+expansion coefficients of an experimental VM image over the projected basis
+directly gives the expansion coefficients of the initial 3D velocity direction
+and thus the sought radial distributions.
+
+Finding the expansion coefficients is a simple linear problem, and the forward
+Abel transforms of the basis functions can be calculated easily if the basis is
+chosen wisely.
+
+See :ref:`rBasexmath` for the complete description.
```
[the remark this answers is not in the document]
MikhailRyazanov: Well, if you click on [+] to expand “How it works” in the TOC, it will show up there. :–) I'll try to check whether it can be expanded there by default. About including everything in the main document, I don thick that it is a good idea to just put everything in “How it works”. Maybe insert it before “Citation"? But is is really long and heavy.

doc/transform_methods/rbasex.rst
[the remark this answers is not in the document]
MikhailRyazanov: In the laboratory frame the photoelectron angular distribution is directly related to the orbital angular momentum carried out by the electron, which must be equal to the angular momentum change upon the transition, which is in turn limited by the number of absorbed photons (each gives 0 or ±1). This is true for any non-oriented systems, as far as I know (from reading Zare's works). For oriented systems this is more complicated and generally involves infinitely many spherical harmonics. @stggh should know more...

doc/transform_methods/rbasex-coord.py
```diff
@@ -0,0 +1,92 @@
...
+
+mpl.rcParams['lines.solid_capstyle'] = 'round'
+
+xmax, ymax, zmax = 6, 7, 6  # axes
+x, y, z = 3, 6, 5  # vector
+lo = 0.5  # label offset
+an = 11  # angle arc segments
+anc = an // 2  # index of central point
+
+fig = plt.figure(figsize=(3, 3)) #??, frameon=False
```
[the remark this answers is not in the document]
MikhailRyazanov: I think, by default Matplotlib was adding frames around figures, which I do not want, so I had `frameon=False` everywhere. But this one did not have a frame even without this option. I should better include it here also for consistency.

doc/transform_methods/rbasex-math.rst
```diff
@@ -0,0 +1,810 @@
...
+
+.. math::
+    (x, y, z) \mapsto (x, y),
+
+as shown by the dashed line:
+
+.. plot:: transform_methods/rbasex-coord.py
+    :align: center
+
+This figure also illustrated important relations between the 3D and 2D radii:
```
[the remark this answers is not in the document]
MikhailRyazanov: Thanks!

doc/transform_methods/rbasex-math.rst
```diff
@@ -0,0 +1,810 @@
...
+
+.. math::
+    b_R(\rho) = \begin{cases}
+        \rho - (R - 1), & R - 1 < \rho < R, \\
+        (R + 1) - \rho, & R \leqslant \rho < R + 1, \\
+        0 & \text{otherwise}
+    \end{cases}
+
+(and :math:`b_0(\rho)` does not have the inner part, :math:`R - 1 < \rho < R`,
+since :math:`\rho \geqslant 0`). These functions actually form a basis of the
```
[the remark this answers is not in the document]
MikhailRyazanov: What about “These functions form a basis of piecewise linear functions with nodes at each pixel, so for any radial distribution, its piecewise linear approximation on the pixel grid can be represented in this basis.”?

doc/transform_methods/rbasex-peak.py
```diff
@@ -0,0 +1,73 @@
...
+         'ro--', label='1 px')
+P = p(R, n)
+plt.plot(r, P, 'r')
+plt.plot(r[::samp], P[::samp], 'ro:')
+
+plt.plot([-1, R - 2, R - 1,   R, R + 1, R + 2, rmax + 1],
+         [ 0,     0,   0.4, 0.6,   0.4,     0,        0],
+         'go--', label='3 px')
+P = 0.4 * p(R - 1, n) + 0.6 * p(R, n) + 0.4 * p(R + 1, n)
+plt.plot(r, P, 'g')
```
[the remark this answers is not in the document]
MikhailRyazanov: OK, changed to cyan.

doc/transform_methods/rbasex-math.rst
```diff
@@ -0,0 +1,810 @@
...
+        P_n(\mathbf r) &= \mathbf P_n^{\rm T} I_n(\boldsymbol \rho), \\
+        I_n(\boldsymbol \rho) &= \big(\mathbf P_n^{\rm T}\big)^{-1} P_n(\mathbf r)
+    \end{aligned}
+
+for each angular order separately. Since all projected basis functions satisfy
+:math:`p_{R;n}(r \geqslant R + 1) = 0` (see the plots above), the matrices
+:math:`\mathbf P_n^{\rm T}` are upper triangular, and their inversions
+:math:`\big(\mathbf P_n^{\rm T}\big)^{-1}` are also upper triangular for all
+:math:`n`, which additionally facilitates the computations. (This triangularity
+makes the inverse Abel transform similar to the “onion peeling” procedure
```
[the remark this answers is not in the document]
MikhailRyazanov: Maybe, but which? If I refer to `onion_bordas`, then I need to explain that here I “peel” spherical shells instead of circular rings in cylindrical coordinates. And then it alludes to the “polar onion-peeling” (http://dx.doi.org/10.1063/1.3126527), but that method is actually not correct (their equation 6 is valid only for infinitely narrow radial function), although it works more or less fine at sufficiently large radii.

doc/transform_methods/rbasex-math.rst
```diff
@@ -0,0 +1,810 @@
...
+5. If the transformed image is needed, it is constructed from its radial
+   distributions obtained in step 4 using the first formula in this section.
+   This takes :math:`O(N\,R_\text{max}^2)` time.
+
+That is, only step 3 has time complexity that scales cubically with the image
+size, and all other steps have quadratic complexity. However, for the forward
+Abel transform, step 3 is not needed at all, and for the inverse Abel
+transform, its results can be cached. Thus processing a sequence of images
+takes time linearly proportional to the total number of processed pixels. In
+other words, the throughput is independent on the image size.
```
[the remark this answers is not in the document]
MikhailRyazanov: By the “throughput” I mean the number pixels per second, as we define it in RSI. The number of images per second is the “frame rate”. I can write “pixel throughput” to make it unambiguous.

doc/transform_methods/rbasex-math.rst
```diff
@@ -0,0 +1,810 @@
...
+   data using :class:`abel.tools.vmi.Distributions`. This takes
+   :math:`O(N\,R_\text{max}^2)` time, where :math:`N` is the number of angular
+   terms, and :math:`R_\text{max}` is the largest analyzed radius (assuming
+   :math:`N \ll R_\text{max}`).
+2. Radial projected basis functions are computed to construct the
+   :math:`\mathbf P_n` matrices, also in :math:`O(N\,R_\text{max}^2)` total
+   time.
+3. For the inverse Abel transform, the :math:`\mathbf P_n^{\rm T}` matrices are
+   inverted, in :math:`O(N\,R_\text{max}^3)` total time. This step is not
+   needed for the forward Abel transform.
```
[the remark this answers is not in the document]
MikhailRyazanov: Not really “of the same image size”, but most other parameters also must be the same. I discuss caching immediately after describing these steps. Do you think that it should be mentioned in this mathematical part also?

doc/transform_methods/rbasex-math.rst
```diff
@@ -0,0 +1,810 @@
...
+Instead of inverting :math:`\mathbf A` explicitly, the solution of
+:math:`\mathbf y = \mathbf A \mathbf x` can be found as
+
+.. math::
+    \mathbf x = \mathop{\rm arg\,min}\limits_{\mathbf x}
+                (\mathbf y - \mathbf A \mathbf x)^2,
+
+from a quadratic minimization (“least-squares”) problem, which is equivalent to
+the original problem, but makes evident that for ill-conditioned problems the
+minimum is very “flat”, and many different :math:`\mathbf x` can be accepted as
```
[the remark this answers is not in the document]
MikhailRyazanov: Well, **x** is a vector. I think, “values of **x**” is formally correct, but sounds misleading. Maybe “...many different vectors **x** can be accepted as a solution”?

doc/transform_methods/rbasex-math.rst
```diff
@@ -0,0 +1,810 @@
...
+Unlike the :math:`L_2`-norm regularization, which tends to avoid sign-changing
+functions and oscillating functions in general, this regularization can produce
+noticeably overshoots (including negative) around sharp features in the
+distribution. However, it tends to preserve the overall intensity.
+
+
+Truncated SVD
+^^^^^^^^^^^^^
+
+This is the method mentioned in the pBasex article (but not actually used in
```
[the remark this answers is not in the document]
MikhailRyazanov: Done.

doc/transform_methods/rbasex-sim.py
```diff
@@ -0,0 +1,53 @@
...
+
+# test distribution
+source = SampleImage(n=2 * rmax - 1).image / scale
+Isrc, _ = Ibeta(source)
+Inorm = (Isrc**2).sum()
+
+# simulated projection fith Poissonian noise
+proj, _ = rbasex_transform(source, direction='forward')
+proj[proj < 0] = 0
+proj = np.random.RandomState(0).poisson(proj)  # (reproducible, see NEP 19)
```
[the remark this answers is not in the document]
MikhailRyazanov: NumPy Enhancement Proposal 19 — Random Number Generator Policy:
https://numpy.org/neps/nep-0019-rng-policy.html
This link is probably stable, so I can include it instead of “NEP 19”.

doc/transform_methods/rbasex-regRMSE.py
```diff
@@ -0,0 +1,83 @@
...
+plt.yticks([0.079, 0.080])
+
+plt.xlim((0, 0.055))
+plt.xlabel('strength')
+
+plt.legend()
+
+# finish
+plt.tight_layout()
+
```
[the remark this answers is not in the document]
MikhailRyazanov: Done.

doc/transform_methods/rbasex-math.rst
```diff
@@ -0,0 +1,810 @@
...
+The reconstruction error does not decrease monotonically, but exhibits several
+local minima before starting to grow. Notice that even the global minimum is
+only slightly better than no regularization.
+
+The actual reconstructed images for each regularization method at its optimal
+strength are shown below with their radial intensity distributions:
+
+
+Using ``reg=None``
+""""""""""""""""""
```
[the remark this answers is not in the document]
MikhailRyazanov: This will be difficult. The tick labels already overlap with the image to save space. This is not very noticeable, but the vertical-axis label will be drawn over the image and look very bad. Making the panels separated will squeeze the plots too much, and the whole figure cannot be made wider. Maybe this is not a big deal, since everything is explained immediately below the figure?

MikhailRyazanov:
Thanks, @DanHickstein! I have made some changes and asked about less trivial ones, please take a look.

doc/transform_methods/rbasex-math.rst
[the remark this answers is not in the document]
DanHickstein: Maybe offer the layman's explanation afterwards: "In other words, this basis set allows us to draw straight lines between pixel values, which is the simplest way to interpret a pixel-based dataset." Or something like this.

doc/transform_methods/rbasex-math.rst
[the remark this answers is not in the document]
DanHickstein: Well, which "onion peeling" procedure did you have in mind when you wrote this? Just the general idea of working from the outside in?

doc/transform_methods/rbasex-math.rst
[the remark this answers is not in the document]
DanHickstein: Good point. Probably the discussion below is sufficient.

doc/transform_methods/rbasex-math.rst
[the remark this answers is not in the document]
DanHickstein: Okay, I understand,. Maybe write: "In other words, the throughput (measured in pixels per second) is independent on the image size."

doc/transform_methods/rbasex-math.rst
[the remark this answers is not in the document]
DanHickstein: that wording also works for me.

doc/transform_methods/rbasex-math.rst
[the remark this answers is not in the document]
DanHickstein: okay, we can skip the labels.

doc/transform_methods/rbasex-sim.py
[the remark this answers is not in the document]
DanHickstein: Cool!

DanHickstein:
Okay, I went through and replied to your questions and comments. Please let me know if I missed any. It looks like there are still some changes that you still need to make.

MikhailRyazanov:
@DanHickstein, I have more or less implemented your suggestions: restructured “How to use it” to start from `Transform`; added some words (and a reference) about regularization to “When to use it”; added a reference to `onion_peeling` with some more details; plus did other, minor corrections.

Most importantly, I have added a check to `Transform` that raises an error if its `origin` is used with `method='rbasex'` in a wrong way. I think, it is better to fail and show a useful message instead of producing unexpected results (with or without a warning).

If everything looks _fine_, maybe let's merge this? The few remaining issues, like restructuring the documentation, can be addressed later (I'm planning to do some other additions to the method anyway).

DanHickstein:
@MikhailRyazanov, I had a look at the most recent changes that you've made and they look great!

I think that having Transform raise an error is the right way to go. The error messages are descriptive and should be quite helpful for users. 

I think that you can go ahead and merge. Thank you for this fantastic addition!

MikhailRyazanov:
Thanks!
Merging...

DanHickstein:
Great work!

Issue #84, opened 2016-02-12
unit test to test examples


DanHickstein:
As we are making changes to PyAbel, it seems that we often break the examples. Is it possible to make a unit test that runs the examples? Or, at least the examples that can run quickly (i.e., not basex).

rth:
The fact that we break examples when the interface changes is indeed a problem. Still I do not think that integrating this it in the main test suite, designed to test that PyAbel code works fine is optimal. I think adding a separate test suite would be better (on Travis we can run both one after the other), and I will look into it.

DanHickstein:
That makes sense. A separate test suite would be fine with me. I think that it would help us spot issues with the code, and make sure that the users have reliable examples to start from.

MikhailRyazanov:
If we want to test that the examples just do not fail, rather than that they produce reasonable results, then it might be reasonable to add a test that builds the documentation — it includes all the examples, and checking that the documentation does not break is useful by itself. It does not seem to have any large or unusual dependencies, builds in a reasonable amount of time, and now (after #241) does so without errors and warnings.

rth:
Usually one would just build the documentation in a seperate CI job.

It's not great the examples to the test suite, because the test suite is intended to be fast (TDD development and all that) while examples can take longer to run. Also the way pytest collects test failures etc. is just not great for running examples.

MikhailRyazanov:
I did not mean to include this in the main tests. As you said before, "adding a separate test suite would be better".

Regarding the speed, currently Travis and especially AppVeyor take much more time setting the environment than doing the actual tests. Sometimes AppVeyor takes >10 minutes per test, and it has 6 of them, so this does not look like particularly "fast"...

> Also the way pytest collects test failures etc. is just not great for running examples.

Well, Travis and AppVeyor are not bound to pytest. Not even to Python. So, I suppose, using `make html` would work. Optionally, with the `'warning-is-error'` option for Sphinx to catch the warnings also.

MikhailRyazanov:
By the way, how/when readthedocs builds it? I do not see it being updated after merging #241. Does it need any additional actions? (From you?) Or it has already tried and encountered some fatal error (which I cannot see anywhere)?

DanHickstein:
Circling back to this, I think that we should try to build the documentation on Travis (or appveyor). I'm working on a PR for matplotlib, and they build the documentation and the check fails if there are warnings. It seems like a good idea. I'm not sure exactly how to do it, but I'll look into it further.

[...]

PR #271, opened 2020-02-03
Testing if readthedocs will autobuild the documentation for a pull request

DanHickstein:


DanHickstein:
Oh good! It seems like read-the-docs is at least attempting to build the docs for this PR. I think that I didn't have the webhook set up correctly the first time. Now I set the webhook to send read-the-docs a webhook for all activity on the repo. I think that only pull requests actually do anything.

DanHickstein:
To see the docs from this PR, just click on the "Details" text next to the **continuous-documentation/read-the-docs** text. The, click on the little text on the right of the screen that says "View Docs" (not the green "View Docs" button).

I think that this is pretty cool! This takes care of four separate things:

1. Ensures that the docs build without errors.
2. Allows us to see the docs for each PR for proofreading.
3. Ensures that all of the examples can run without errors. 
4. Let's us see the output of all the examples for each build.

We can close this PR now, I'm just leaving it open so that people can see how cool this is :)

DanHickstein:
Ah, darn. When the examples break, we want the readthedocs build to fail.

I think that we need to do something like this: https://stackoverflow.com/questions/38048945/how-to-turn-warnings-into-errors-when-building-sphinx-documentation-with-setupto/42426616#42426616

MikhailRyazanov:
Will it stop after the first error then?

By the way, is there an easy way to access the logs (even on success)? They are available on https://readthedocs.org/projects/pyabel/builds/, but having a direct link would be convenient.

DanHickstein:
> Will it stop after the first error then?
I'm not sure. It would be best if it showed all of the errors and then failed at the end. You could try running Sphinx with the `-W` flag and see what it does. 

> by the way, is there an easy way to access the logs (even on success)?
If you click the "Details" button before it's finished building, then it takes you to the Log. But, it looks like if you click the button after it succeeds, then it takes you straight to the documentation instead. Looks like you need to go to the builds page and look at the log for that particular PR.

MikhailRyazanov:
> It would be best if it showed all of the errors and then failed at the end.

This is what I was hoping for.

> You could try running Sphinx with the `-W` flag and see what it does.

Apparently, fails immediately.
But the next comment at your SO link suggests adding `--keep-going`, which seems to work as needed: sphinx does as much as it can, reports all errors and "fails" at the end (`make` reports an error).

> If you click the "Details" button before it's finished building, then it takes you to the Log. But, it looks like if you click the button after it succeeds, then it takes you straight to the documentation instead. Looks like you need to go to the builds page and look at the log for that particular PR.

Hm, I did not see what happes during the build process, but if they have the log link there, maybe suggest them to keep it after the build succeeds? (I guess, at least if the build fails, the link to logs will remain there instead of the docs link?)

DanHickstein:
Okay, so we want both the -W and the -keep-going flags. I'm not sure how to implement that on readthedocs through...

Yes, I assume that if the readthedocs build fails, then the "Details" link likely takes us to the logs. If it succeeds, then it takes us straight to the docs. This actually seems like the desired behavior, right?

MikhailRyazanov:
> I'm not sure how to implement that on readthedocs through...

If they are using our Makefile, then put `-W --keep-going` here:
https://github.com/PyAbel/PyAbel/blob/8f8d7e864475ae823837b0925a88eaac7b3ffdc7/doc/Makefile#L5
Otherwise, I have no idea.

> If it succeeds, then it takes us straight to the docs. This actually seems like the desired behavior, right?

Well, there can be some warnings that might be useful to see even if the build succeeds... But in principle, if it fail on serious errors and then show the log, this would be close enough. Anyway, much better than nothing. :–)

If you find out how to add these `-W --keep-going` to the build options, it would be interesting to test how all this works.

DanHickstein:
Hmm, I tried including this in the makefile, but I think that readthedocs must not be using our makefile. I wish there were some way to include this in the conf.py file.

MikhailRyazanov:
Then what exactly they are using?
According to [sphinx-build man page](http://www.sphinx-doc.org/en/master/man/sphinx-build.html#environment-variables), it should also take options from the `SPHINXOPTS` environment variable. And [apparently](https://docs.readthedocs.io/en/stable/builds.html#build-environment) you can define your own environment variables in the RTD project settings. Give it a try?

DanHickstein:
Yeah, I was able to define a "SPHINXOPTS" environment variable on readthedocs, but it doesn't seem to be respected by sphinx-build on readthedocs. 

Now, I'm trying to include a `.readthedocs.yml` file...

DanHickstein:
Okay, I was able to get Sphinx to fail on warnings! The key is to include
```
sphinx:
  fail_on_warning: true
```
in a `.readthedocs.yml` file.

This causes the build to fail on the first warning. It would be better if we could have all of the warnings lists and then have the build fail. But, I think that this behavior is better than nothing.

MikhailRyazanov:
> Yeah, I was able to define a "SPHINXOPTS" environment variable on readthedocs, but it doesn't seem to be respected by sphinx-build on readthedocs.

Really strange. I tried it locally, and if `SPHINXOPTS` is not defined in the `Makefile`, then `sphinx-build` uses the `SPHINXOPTS` environment variable and works as expected.

> But, I think that this behavior is better than nothing.

Sure. :–)
If there is no way to pass `-W --keep-going`, this at least does the most important part.

DanHickstein:
According to the readthedocs docs, adding the SPHINXOPTS = -W --keep-going to the make file should work. I'm going to try that again...
image
from page 66 here: https://readthedocs.org/projects/docs/downloads/pdf/latest/

DanHickstein:
Nope, for some reason readthedocs doesn't respect the `SPHINXOPTS = -W --keep-going` in the doc/Makefile. I think that including the `fail_on_warning: true` in the readthedocs.yml file is the best we can do.

DanHickstein:
Okay, so, I think that we should merge this PR, since it enables the doc build to fail on an error. This will finally let us close #84.

MikhailRyazanov:
Yes, will be helpful and probably the best we can do so far.

By the way, we should probably replace Python 3.6 with 3.8, since 3.6 is already “old” (it's full support ended more than a year ago), and even in conservative Debian “Python 3” has recently changed from 3.7 to 3.8 (that is, 3.8 is now the default version, so we really need to test with it).

DanHickstein:
Good point about Py3.8. I can update our builds on Travis and Appveyor. I'll do that as a separate PR. I'm going to go ahead and merge this one.

PR #279, opened 2020-04-22
included python 3.8 tests

DanHickstein:
As @MikhailRyazanov pointed out (#271 (comment)), we should be testing with the latest version of Python (3.8). I removed the tests for Python 3.6, since I think that testing on three Python versions is enough (2.7, 3.7, 3.8). 

(I know that @stggh will argue that we drop the Py2.7 test. He's probably right. Maybe we keep Py2 compatibility and testing through 2020 and we can celebrate New Years by going all Py3?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,14 @@ language: python
 sudo: false
 matrix:
   include:
-  - python: '3.7'
+  - python: '3.8'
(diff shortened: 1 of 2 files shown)


DanHickstein:
Cool, I'll merge this. Thanks!

Issue #268, opened 2020-01-15
PEP8/pycodestyle enforcement?


DanHickstein:
Many of the changes that @MikhailRyazanov made in #267 were tweaking the code to conform to the python coding style as describe in [PEP8](https://www.python.org/dev/peps/pep-0008/) and often checked using the [pycodestyle](https://pypi.org/project/pycodestyle/) tool. This brings up two issues:

1) We should probably make the decision to have PyAbel meet PEP8. Pycodestyle reports about 700 PEP8 violations for the whole PyAbel directory. So, it would be quite a project. But, we could work through each file at a time and break it into manageable chunks. (BTW, I think that the 79 character line limit is a bit too restrictive, and would suggest that we relax it to ~100 characters.)

2) A lot of projects use an automatic PEP8 checker for pull requests. I'm not sure what is considered the best way to do this, but "pep8speaks" looks pretty cool: https://pep8speaks.com/

MikhailRyazanov:
In principle, I support the idea of moving towards PEP8, since its recommendations _in most cases_ improve the readability. But I also have some concerns.

If we want to enforce it strictly, then all these 700 violations need to be addressed before enabling the automatic checker. This is actually not so difficult — meaningful PRs have hundreds of deliberated lines changed, and these changes will be almost automatic, as the the tool itself says what to do.

But in _some cases_ I consider `pycodestyle` warnings detrimental. For example, it often complains about “continuation line over-indented for visual indent” where the indent actually makes structural sense. Complaining about “ambiguous variable name 'I'” for `I = np.identity(...)` is also counterproductive (any other name for this variable will be worse). And, to my taste,
```
if isinstance(rmax_in, int):
    rmax = rmax_in
elif rmax_in == 'hor':
    rmax = hor
elif rmax_in == 'ver':
    rmax = ver
elif rmax_in == 'HOR':
    rmax = HOR
elif rmax_in == 'VER':
    rmax = VER
elif rmax_in == 'min':
    rmax = min(hor, ver)
elif rmax_in == 'max':
    rmax = max(hor, ver)
elif rmax_in == 'MIN':
    rmax = min(HOR, VER)
elif rmax_in == 'MAX':
    rmax = max(HOR, VER)
elif rmax_in == 'all':
    rmax = int(np.sqrt(HOR**2 + VER**2))
else:
    raise ValueError('Incorrect radial range "{}"'.format(rmax_in))
```
is less readable and worse structured than
```
[...]

PR #281, opened 2020-04-23
PEP8 checking for new PRs

DanHickstein:
Following up on the discussion in https://github.com/PyAbel/PyAbel/issues/268, I think that it would be a good idea to activate pep8speaks on our repository: https://pep8speaks.com/

You can see a demo of pep8speaks in operation over on my PyAbel fork: https://github.com/DanHickstein/PyAbel/pull/1

Basically, it just makes a comment that identifies all of the pycodestyle errors for the lines added or altered by the current PR. It doesn't fail the tests, so we are free to ignore it's advice as we please. But, it should serve as a good reminder for us to write code that is pycodestyle compliant. We could use the default behavior of pep8speaks, but I think that it's good to have the `.pep8speaks.yml` file in place so that we can alter the behavior in the future.


DanHickstein:
The docs build failed, and I got this fairly unhelpful error:

```
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyabel/envs/281/lib/python3.7/site-packages/sphinx/cmd/build.py", line 304, in build_main
    app.build(args.force_all, filenames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyabel/envs/281/lib/python3.7/site-packages/sphinx/application.py", line 341, in build
    self.builder.build_update()
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyabel/envs/281/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 347, in build_update
    len(to_build))
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyabel/envs/281/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 360, in build
    updated_docnames = set(self.read())
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyabel/envs/281/lib/python3.7/site-packages/sphinx/util/logging.py", line 261, in pending_warnings
    memhandler.flushTo(logger)
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyabel/envs/281/lib/python3.7/site-packages/sphinx/util/logging.py", line 224, in flushTo
    logger.handle(record)
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/logging/__init__.py", line 1529, in handle
    self.callHandlers(record)
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/logging/__init__.py", line 1591, in callHandlers
    hdlr.handle(record)
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/logging/__init__.py", line 901, in handle
    rv = self.filter(record)
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/logging/__init__.py", line 762, in filter
    result = f.filter(record)
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyabel/envs/281/lib/python3.7/site-packages/sphinx/util/logging.py", line 410, in filter
    raise SphinxWarning(location + ":" + message)
sphinx.errors.SphinxWarning: /home/docs/checkouts/readthedocs.org/user_builds/pyabel/checkouts/281/doc/transform_methods/rbasex-math.rst:266:Exception occurred in plotting rbasex-pRn
 from /home/docs/checkouts/readthedocs.org/user_builds/pyabel/checkouts/281/doc/transform_methods/rbasex-pRn.py:
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/site-packages/numpy/core/function_base.py", line 117, in linspace
    num = operator.index(num)
TypeError: 'float' object cannot be interpreted as an integer

  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/site-packages/matplotlib/sphinxext/plot_directive.py", line 503, in run_code
    exec(code, ns)
  File "<string>", line 19, in <module>
  File "<__array_function__ internals>", line 6, in linspace
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/site-packages/numpy/core/function_base.py", line 121, in linspace
    .format(type(num)))
TypeError: object of type <class 'float'> cannot be safely interpreted as an integer.

Warning, treated as error:
/home/docs/checkouts/readthedocs.org/user_builds/pyabel/checkouts/281/doc/transform_methods/rbasex-math.rst:266:Exception occurred in plotting rbasex-pRn
 from /home/docs/checkouts/readthedocs.org/user_builds/pyabel/checkouts/281/doc/transform_methods/rbasex-pRn.py:
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/site-packages/numpy/core/function_base.py", line 117, in linspace
    num = operator.index(num)
TypeError: 'float' object cannot be interpreted as an integer

  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/site-packages/matplotlib/sphinxext/plot_directive.py", line 503, in run_code
    exec(code, ns)
  File "<string>", line 19, in <module>
  File "<__array_function__ internals>", line 6, in linspace
  File "/home/docs/.pyenv/versions/3.7.3/lib/python3.7/site-packages/numpy/core/function_base.py", line 121, in linspace
    .format(type(num)))
TypeError: object of type <class 'float'> cannot be safely interpreted as an integer.
```

I think that there is a float/int issue in rBasex-pRn.py, but I can't reproduce it on my machine. It's quite unhelpful that the error message on readthedocs doesn't give the linenumber of the error in a straightforward way.

DanHickstein:
I found a [similar problem](https://github.com/pydata/numexpr/issues/354) that was solved by making sure that the third argument to np.linspace was an integer. But, I fixed this issue in rBasex-pRn.py and the problem persists.

DanHickstein:
Oh, actually I think that I fixed the issue in rBasex-pRn.py, but rBasex-peak.py had the same problem. @MikhailRyazanov, it turns out you don't know how to use np.linspace :)

This issue must have popped up because readthedocs upgraded to numpy 1.18, which treats passing linspace a float for the third argument as an error.

MikhailRyazanov:
Well, #271 already helps! :–)

> it turns out you don't know how to use np.linspace :)

Somebody's NumPy is broken. :–/ Before v1.18, it was supposed to print a warning ([1](https://github.com/numpy/numpy/blob/v1.16.0/numpy/core/function_base.py#L128), [2](https://github.com/numpy/numpy/blob/v1.16.0/numpy/core/function_base.py#L20-L29)), but I never saw it (I have v1.16 now). What is also amusing, Python knows that `(8.5 * 20 + 1).is_integer()` is `True`, so why this number “cannot be interpreted as an integer”?..

In any case, thanks for noticing and correcting!

.pep8speaks.yml
```diff
@@ -0,0 +1,14 @@
+# File : .pep8speaks.yml
+ 
+scanner:
+    diff_only: True  # If False, the entire file touched by the Pull Request is
+                     # scanned for errors. If True, only the diff is scanned.
+    linter: pycodestyle  # Other option is flake8
+ 
+pycodestyle:  # Same as scanner.linter value. Other option is flake8
+    max-line-length: 90  # Default is 79 in PEP 8
```
MikhailRyazanov, line 9: Can we keep the default here?
In your example [here](https://github.com/PyAbel/PyAbel/issues/268#issuecomment-574984497) I can see only 88 characters:
longline
We can always ignore this warning when needed, but using the standard value is probably better than some random number.

doc/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -W --keep-going
```
MikhailRyazanov, line 5: This is useful for local builds (I actually use these options in my local copy), even if currently it does not work on readthedocs.

DanHickstein: Yeah, I realized that I wanted it for my local builds as well, so I am sneaking it in here :)

.pep8speaks.yml
DanHickstein: Agreed. We should also tell people that it's okay to break the rules sometimes. I think that we can change the message from pep8speaks, so maybe we can change it to something that refers folks to our style policy: "try to follow PEP8 unless breaking the rules makes your code better."

DanHickstein:
Following up on the line-length discussion, we should probably include a blurb in the CONTRIBUTING.rst that sets out our style policy. I'm not sure if we really have much of a style policy, but maybe we can work on it. Here is what I would write so far:

"**Code Style:** We want to code of PyAbel to be understandable, hackable, and maintainable for years to come. So, please include plenty of comments and docstrings for functions. 

PyAbel attempts to follow [PEP8 ](https://www.python.org/dev/peps/pep-0008/) style when possible, since it typically produces code that is easier to read. We encourage PRs to follow PEP8 style. You can check your code using [pycodestyle](https://pypi.org/project/pycodestyle/), which can often be incorporated into your text editor. We have also implemented automated pycodestyle checking of all Pull Requests using [pep8speaks](https://github.com/OrkoHunter/pep8speaks). However, [producing readable code is the primary goal](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds), so please go ahead and break the rules of PEP8 when doing so improves readability. For example, if a section of your code is easier to read with 120-character lines instead of 79-character lines, then use the longer lines."

MikhailRyazanov:
> We want _to_ code of PyAbel to be understandable

_the_?

> if a section of your code is easier to read with 120-character lines instead of 79-character lines, then use the longer lines.

I think, 120 is really too long (at least for me, it will not be easier to read). Maybe change to something milder, like “easier to read with lines slightly longer than 79 character, then use the longer lines”?

Otherwise, looks great.

MikhailRyazanov:
By the way,

> Somebody's NumPy is broken

Apparently, everybody's. Their supposed deprecation warning did not work: https://github.com/numpy/numpy/issues/15821#issuecomment-602949800.

DanHickstein:
> What is also amusing, Python knows that (8.5 * 20 + 1).is_integer() is True, so why this number “cannot be interpreted as an integer”?..

Yeah, that seems reasonable enough to me. It seems that numpy has tightened up its rules and wants it to actually be of type int. I would have expected a lengthier discussion, but this was all that I found: https://github.com/numpy/numpy/issues/7289

And that makes sense about the botched depreciation message. Seems like this caused issues with some other projects as well.

I'll make your suggested changes to the CONTRIBUTING text.

DanHickstein:
Okay, I made a few changes:

1. I changed the max line length to 79 characters, and removed all ignored errors. We are now checking for full PEP8 and can ignore things that we don't care about. The example pep8speaks.yml was ignoring these two, which I agree are pretty stupid most of the time. Let me know if we should go back to ignoring them: 
    ```
        - W504  # line break after binary operator
        - E402  # module level import not at top of file
    ```
2. I included a section in CONTRIBUTING.rst that outlines out style suggestions/policy. Please have a look.

3. **I included `*.svg` in the `.gitignore`.**  Let me know if this poses a problem anywhere. I generated a svg file using on of @MikhailRyazanov's new rBasex doc scripts and I almost accidentally committed it. Probably we should be ignoring all image files that are not already in the repo.

CONTRIBUTING.rst
```diff
@@ -58,6 +58,14 @@ which should launch a browser window displaying the docs. When you save a change
...
 
 
+Code Style
+----------
+
+We hope that the PyAbel code will be understandable, hackable, and maintainable for many years to come. So, please use good coding style, include plenty of comments, use docstrings for functions, and pick informative variable names.
+
+PyAbel attempts to follow `PEP8 <https://www.python.org/dev/peps/pep-0008/>`_ style whenever possible, since the PEP8 recommendations typically produces code that is easier to read. You can check your code using `pycodestyle <https://pypi.org/project/pycodestyle/>`_, which be called from the command line or incorporated right into most text editors. Also, PyAbel is using automated pycodestyle checking of all Pull Requests using `pep8speaks <https://pep8speaks.com/>`_. However, `producing readable code <https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds>`_ is the primary goal, so please go ahead and break the rules of PEP8 when doing so improves readability. For example, if a section of your code is easier to read with lines slightly longer than 79 characters, then use the longer lines."
```
MikhailRyazanov, line 66: "which be called" is probably "which can be called"?

MikhailRyazanov:
Yes, ignoring SVG files by default makes sense. If we need to include handmade images in the documentation, we should be able to add them with `git add -f`.

DanHickstein:
> we should be able to add them with git add -f.

Yup. Or we can add a 
```
!specific_file.svg
```
line to the gitignore.

CONTRIBUTING.rst
DanHickstein: fixed.

DanHickstein:
Okay, I'm going to merge this. I activated pep8speaks for this repo, so future PRs should be checked. Thanks for your help @MikhailRyazanov!

Issue #285, opened 2020-04-25
Something is wrong with the horizontal symmetry axis


leon-vv:
```
import numpy as np
import abel

xx, yy = np.meshgrid(np.linspace(-256, 256, 512), np.linspace(-256, 256, 512))
r = np.sqrt(xx**2 + yy**2)

a = np.exp(-1/5000*r**2)

plt.imshow(a)
plt.colorbar()

t = abel.Transform(a, symmetry_axis=0).transform;

plt.figure()
plt.imshow(t)
```

This results in

![image](https://user-images.githubusercontent.com/4085287/80280891-78d57f80-8707-11ea-80d9-dd38c02446a7.png)

which looks just fine.

Now keeping with the vertical axis of symmetry I move the gauss function along the y axis.
`r = np.sqrt(xx**2 + (yy-80)**2)`. The result is as expected:

![image](https://user-images.githubusercontent.com/4085287/80280932-bd611b00-8707-11ea-826c-737f783576d8.png)

Now move the gauss along the x-axis  `r = np.sqrt((xx-80)**2 + yy**2)` and choose a horizontal axis of symmetry: `t = abel.Transform(a, symmetry_axis=1).transform;`. This should result in the same images, except transposed. But we get:

![image](https://user-images.githubusercontent.com/4085287/80281010-6ad42e80-8708-11ea-83ba-8a301d2b9283.png)

stggh:
@leon-vv thanks for providing the code to illustrate this issue.

Your second case breaks cylindrical symmetry and so the Abel transform is invalid. See the [docs](https://pyabel.readthedocs.io/en/latest/readme_link.html), and the wine glass image figure.

In your example, specifying the symmetry origin facilitates the Abel transform. Passing the `origin` (centre of symmetry) parameter for the transform fixes this. Note, also, the image width should be an odd integer for `PyAbel` transforms.
```python
xx, yy = np.meshgrid(np.linspace(-255, 255, 511), np.linspace(-255, 255, 511))
r = np.sqrt((xx-80)**2 + yy**2)

a = np.exp(-1/5000*r**2)

plt.imshow(a, extent=[-255, 255, -255, 255])
plt.colorbar()

t = abel.Transform(a, symmetry_axis=1, origin=(255, 255+80)).transform

plt.figure()
plt.imshow(t, extent=[-255, 255, -255, 255])

plt.show()
```
[...]

PR #288, opened 2020-04-26
Clarify purpose of symmetry_axis argument

leon-vv:
See issue #285

### DIFF
--- a/abel/transform.py
+++ b/abel/transform.py
@@ -101,7 +101,11 @@ class Transform(object):
 
     symmetry_axis : None, int or tuple
         Symmetrize the image about the numpy axis
-        0 (vertical), 1 (horizontal), (0, 1) (both axes)
+        0 (vertical), 1 (horizontal), (0, 1) (both axes). Note that the
+        abel transform is always performed around the vertical axis.
+        This parameter only affects how the image is modified before
+        (and after) applying the Abel transform. For more information,
+        see the "Quadrant combining" note below.
 
     use_quadrants : tuple of 4 booleans
         select quadrants to be used in the analysis: (Q0, Q1, Q2, Q3).


abel/transform.py
```diff
@@ -101,7 +101,10 @@ class Transform(object):
 
     symmetry_axis : None, int or tuple
         Symmetrize the image about the numpy axis
-        0 (vertical), 1 (horizontal), (0, 1) (both axes)
+        0 (vertical), 1 (horizontal), (0, 1) (both axes). Note that the
+        abel transform is always performed around the y-axis.
```
stggh, line 105: Actually, your original statement of [`z-axis`](https://github.com/PyAbel/PyAbel/issues/285#issuecomment-619503209) is correct, according to our wine-glass figure:
image

MikhailRyazanov: It's probably better to call it “vertical” for clarity and consistency.

leon-vv: Yes I hadn't noticed there is an ambiguity in what I mean with the y-axis. I meant the vertical axis, which contradicts the image. So changing 'y-axis' to 'vertical axis' would be better.

abel/transform.py
```diff
@@ -101,7 +101,10 @@ class Transform(object):
 
     symmetry_axis : None, int or tuple
         Symmetrize the image about the numpy axis
-        0 (vertical), 1 (horizontal), (0, 1) (both axes)
+        0 (vertical), 1 (horizontal), (0, 1) (both axes). Note that the
+        abel transform is always performed around the vertical axis.
+        This parameters only affects how the image is modified
+        before applying the Abel transform (see below).
```
DanHickstein, line 107: instead of "This parameters only affects how the image is modified before applying the Abel transform (see below).', can we write something like:

"This parameter only affects how the image is modified before (and after) applying the Abel Transform. For more information, see the "Quadrant combining" note below."

leon-vv: Sure

leon-vv:
Do you want me to stash the three commits into one?

DanHickstein:
> Do you want me to stash the three commits into one?

Sounds good. I think that this is ready for merging once @MikhailRyazanov or @stggh has a moment to review and approve.

abel/transform.py
```diff
@@ -101,7 +101,11 @@ class Transform(object):
 
     symmetry_axis : None, int or tuple
         Symmetrize the image about the numpy axis
-        0 (vertical), 1 (horizontal), (0, 1) (both axes)
+        0 (vertical), 1 (horizontal), (0, 1) (both axes). Note that the
+        abel transform is always performed around the vertical axis.
+        This parameter only affects how the image is modified before
+        (and after) applying the Abel Transform. For more information,
```
MikhailRyazanov, line 107: In “Abel transform”, the word “Abel” should be capitalized (proper name), and the word “transform” should be not (common noun). Having “abel transform” and “Abel Transform” in the same paragraph looks really bizarre. :–)

DanHickstein: Good point. The capitalization was my fault :)

DanHickstein:
@leon-vv, can you fix the capitalization and then we can merge?

leon-vv:
Yes sure, I'm not at home at the moment but will do it soon. Feel free to go ahead and do it yourself

-------- Original Message --------

leon-vv:
All done

DanHickstein:
Great! Thanks for fixing this @leon-vv!!

Issue #296, opened 2020-07-23
Crop in only 1 axis


Mobbybobby:
Hi,

your crop function always crops in both axis, even when the image is centered is with respect to only one axis.

I fixed it for my case:

center.py line 264:
```ruby
 elif crop == 'valid_region':
        # crop to region containing data
        shift0, shift1 = (None, None)
        if isinstance(axes, int):
            if axes == 0 and delta0 != 0:
                shift0 = 1 + int(np.abs(delta0))
                return centered_data[shift0:-shift0, :]
            elif axes == 1 and delta1 != 0:
                shift1 = 1 + int(np.abs(delta1))
                return centered_data[:, shift1:-shift1]
            else:
                raise ValueError("axes value not 0, or 1")
        else:
            if delta0 != 0:
                shift0 = 1 + int(np.abs(delta0))
            if delta1 != 0:
                shift1 = 1 + int(np.abs(delta1))
            return centered_data[shift0:-shift0, shift1:-shift1]
    else:
        raise ValueError("Invalid crop method!!")
```

I hope this helps you or other users.

Thanks and great package!

Mobby

DanHickstein:
Hi Mobby,

Thanks for the suggestion! I think that this makes sense. If the user specifies `axis=1`, then they are still required to enter some value for the center row in the `origin` keyword argument. And, currently, the image will still be cropped using this nonsense value.

The code the you have seems good to me. Are there additional suggestions from @MikhailRyazanov or @stggh?

@Mobbybobby, can you make a Pull Request for this change?

stggh:
A good point!  

We can use the PR to further review the `set_center` code.  

In the block above, `crop == maintain_data`  the `area` reassignment does not look right:

```python
        area = container[:, :]
        if shift0:
            if delta0 > 0:
                area = area[:-shift0, :]
            else:
                area = area[shift0:, :]
        if shift1:
            if delta1 > 0:
                area = area[:, :-shift1]
            else:
                area = area[:, shift1:]
        area[:, :] = data[:, :]
        data = container
```

Mobbybobby:
Yes, I only fixed it for valid_region at the moment since it is the method i need. I never done a Pull request sorry. I usually don't work with GitHub.

stggh:
@MikhailRyazanov of course I should have included you in the "_will no doubt have the solution_" for the all files lines have changed PR  #298. The torment of Windows OS.

MikhailRyazanov:
[...]

PR #298, opened 2020-07-24
Crop method valid region fix for only 1 axis

Mobbybobby:
Hi

like discussed in [Issue#296 ](https://github.com/PyAbel/PyAbel/issues/296)I fixed the cropping of the valid region method. The other methods still need to be fixed!


MikhailRyazanov:
It seems that all the line breaks were changed from the UNIX-style `CR` to the DOS-style `CR`+`LF`. Selecting “Hide whitespace changes” in the diff setting does display the diff properly, but I don't think that we actually want this to be merged as is. So @Mobbybobby probably needs to discard this commit and make it anew without changing the line breaks (perhaps the easiest way would be to just use the internal GitHub editor).

PR #299, opened 2020-07-25
Fixed crop for valid region

Mobbybobby:
Hi again,

like discussed in [Issue#296 ](https://github.com/PyAbel/PyAbel/issues/296)I fixed the cropping of the valid region method. The other methods still need to be fixed!

I hope it's right now.

### DIFF
--- a/abel/tools/center.py
+++ b/abel/tools/center.py
@@ -264,11 +264,21 @@ def set_center(data, origin, crop='maintain_size', axes=(0, 1), verbose=False,
     elif crop == 'valid_region':
         # crop to region containing data
         shift0, shift1 = (None, None)
-        if delta0 != 0:
-            shift0 = 1 + int(np.abs(delta0))
-        if delta1 != 0:
-            shift1 = 1 + int(np.abs(delta1))
-        return centered_data[shift0:-shift0, shift1:-shift1]
+        if isinstance(axes, int):
+            if axes == 0 and delta0 != 0:
+                shift0 = 1 + int(np.abs(delta0))
+                return centered_data[shift0:-shift0, :]
+            elif axes == 1 and delta1 != 0:
+                shift1 = 1 + int(np.abs(delta1))
+                return centered_data[:, shift1:-shift1]
+            else:
+                raise ValueError("axes value not 0, or 1")
+        else:
+            if delta0 != 0:
+                shift0 = 1 + int(np.abs(delta0))
+            if delta1 != 0:
+                shift1 = 1 + int(np.abs(delta1))
+            return centered_data[shift0:-shift0, shift1:-shift1]
     else:
         raise ValueError("Invalid crop method!!")
 


MikhailRyazanov:
It seems to me that if centering with respect to both axes is specified, but at least one of the shifts turns out to be zero, then this new code will still fail. I think, it might be a better idea to use slice objects instead of shifts, something like
```
        if delta0:
            shift = 1 + int(np.abs(delta0))
            slice0 = slice(shift, -shift)
        else:
            slice0 = slice(None)
        if delta1:
            shift = 1 + int(np.abs(delta1))
            slice1 = slice(shift, -shift)
        else:
            slice1 = slice(None)
        return centered_data[slice0, slice1]
```
However, as @stggh has pointed out in #300, the whole function needs to be revised...

DanHickstein:
> However, as @stggh has pointed out in #300, the whole function needs to be revised...

Do we plan to make this revised function soon? If not, are we better off merging this PR for now and leaving #300 open to remind us to re-write the function later?

MikhailRyazanov:
> Do we plan to make this revised function soon?

Ask @stggh... :–)

> merging this PR for now

I don't really like the current solution because it tries to hide problems inside nested `if`s for some particular cases instead of addressing the root cause. Even with these `if`s, trying to center an image that is already centered (both `delta`s = 0) will fail with a `ValueError`, since all the condition will be false.

As I understand, the only problem is that `-None` raises a `TypeError` instead of giving `None`, and this is what needs to be addressed instead of meddling with `axes`. Perhaps just replacing
```
return centered_data[shift0:-shift0, shift1:-shift1]
```
by
```
return centered_data[shift0:-shift0 if shift0 else None, shift1:-shift1 in shift1 else None]
```
can do the job.

MikhailRyazanov:
And by the way, there should be no `1 +` in `shift`s. Thus the whole branch
https://github.com/PyAbel/PyAbel/blob/f110ba34ad1e8e6bc920c40b402a3963527064b2/abel/tools/center.py#L264-L271
can be replaced by
```python
    elif crop == 'valid_region':
        # crop to region containing data
        shift0 = int(np.abs(delta0))
        shift1 = int(np.abs(delta1))
        return centered_data[shift0 : -shift0 if shift0 else None,
                             shift1 : -shift1 if shift1 else None]
```
There are, however, similar problems with trying to do arithmetic with `None` for `crop='maintain_data'`, so this function indeed needs to be revised substantially (and we must write tests for it). I'll look closer at @stggh proposal in #300, as it seems to be more sensible (in particular, `[pad:size-pad]` without `None`s).

stggh:
> Ask @stggh... :–)

My apologies, my attention to this problem was diverted. 

Cropping is a real pain, in the sense that I do not fully comprehend the cropping options:
```python
    crop : str
        determines how the image should be cropped. The options are:

        ``maintain_size``
            return image of the same size. Some regions of the original image
            may be lost and some regions may be filled with zeros.
        ``valid_region``
            return the largest image that can be created without padding.
            All of the returned image will correspond to the original image.
            However, portions of the original image will be lost.
            If you can tolerate clipping the edges of the image, this is
            probably the method to choose.
        ``maintain_data``
            the image will be padded with zeros such that none of the original
            image will be cropped.
```
given that the function simply calls `centered_data  = scipy.ndimage.interpolation.shift(data, (row_shift, col_shift))`. 
Why not pad the image for all cases and then crop to a `PyAbel` friendly nearest odd-column size? In which case `centered_data = centered_data[padr:-padr, padc:-padc]` will do the right thing, no need for any `crop`-options.

MikhailRyazanov:
@stggh, I've [answered](https://github.com/PyAbel/PyAbel/issues/300#issuecomment-674305950) in #300 (a more general topic).

DanHickstein:
This PR has been superseded by #302, correct?

MikhailRyazanov:
Yes, and @Mobbybobby has [confirmed](https://github.com/PyAbel/PyAbel/issues/296#issuecomment-685440184) that #302 solves the problems reported in #296. So now we can close this PR and that issue.

DanHickstein:
Great! Thanks again to @Mobbybobby for pointing out this issue and for @MikhailRyazanov for improving the code.

Issue #303, opened 2020-09-10
Need help to save the after abel transformed image as a 16 bit tiff file


kfxnightmare2:
I tried few ways to save the tif file but all those saving it as a 32bit?

from PIL import Image
pic = Image.fromarray(recon)
pic.save('test-pillow.tif','TIFF')

but the dtype is float32, I tried to convert it to 16 bit but some forums say 16 bit conversion is not accurate in pillow 

Also,
is there a way to load the raw_data in 16-bit

raw_data = plt.imread(filename).astype('float64')  <------- instead of float64

MikhailRyazanov:
It's not very clear what exactly you want to do, but saving a [16-bit grayscale](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes) TIFF seems to work like this:
```python
pic = Image.fromarray(recon.astype(np.uint16), mode='I;16')
pic.save('test-pillow.tif')
```
You might need to normalize your array to the 16-bit unsigned-integer range (0 to 65 535), and perhaps do proper rounding, before applying `.astype()`.

Regarding loading 16-bit data, as far as I understand, Pillow can load 16-bit grayscale TIFFs, at least those saved as shown above. So
```python
pic = Image.open('test-pillow.tif')
raw_data = np.asarray(pic)
```
creates a NumPy array with `dtype=uint16`, that is 16-bit unsigned integers. However, Abel transform is not designed for integers, neither are other PyAbel tools, so using this loaded raw data without converting it to a floating-point format looks like a very strange idea. Do you really need that and why?

MikhailRyazanov:
Another general remark about using image formats for saving/loading raw data: beware of [gamma correction](https://en.wikipedia.org/wiki/Gamma_correction). Most image formats use or assume nonlinear intensity encoding by default, but all data analysis must be done with intensity in a linear scale. So you must either make sure that linear encoding (γ = 1) is used everywhere in the data path, or do appropriate intensity transforms yourself.

stggh:
An excellent point! This comment should not be lost, perhaps, it could be worked into the `PyAbel` docs?

MikhailRyazanov:
> An excellent point!

Which one? About gamma correction?

stggh:
Yes, the gamma correction.

MikhailRyazanov:
[...]

PR #302, opened 2020-08-29
New set_center() implementation and axes in find_origin()

MikhailRyazanov:
OK, here is the `abel.tools.center.set_center()` rewritten completely:
* Should behave correctly for all reasonable parameters.
* Doesn't lose fractional edge pixels (`scipy.ndimage.shift()` is [going](https://github.com/scipy/scipy/issues/12773) to be corrected eventually, but this works with what we have now).
* Allows selecting the interpolation order.
* Includes a much faster implementation without interpolation.
* Allows specifying `None` in the origin for axes that do not need centering.

Also, the `axes` argument is added to `find_origin...()` methods, such that they don't do unnecessary (and potentially unsuccessful) work. Now they return the image center coordinate for unused axis, so passing this to `set_center()` will have no effect.

PR #305, opened 2020-09-24
About intensities (gamma correction, int/float)

MikhailRyazanov:
The proposed addition to README from [here](https://github.com/PyAbel/PyAbel/issues/303#issuecomment-698072238).
Please check whether the wording can be improved or something added/removed...


stggh (approved):
Looks good to me. Thanks, for "exposing" the subtle gamma intensity correction.

README.rst
```diff
@@ -147,17 +147,20 @@ The PyAbel code adheres to the following conventions:
...
 
+-
+    **Intensity:** Absolute pixel intensities can be arbitrary (withing the floating-point range), but the intensity scale must be linear. Keep in mind that cameras and common image formats can use `gamma correction <https://en.wikipedia.org/wiki/Gamma_correction>`_ and thus provide data with nonlinear intensity encoding. When working with such data, make sure to apply appropriate intensity transformations before the analysis. Most PyAbel methods also assume intensities to be floating-point numbers, and when applied to integer types, can return inappropriately rounded results. The ``abel.Transform`` class recasts the input image to ``float64`` by default, but if you wish to call the transform methods directly or use other tools, you might need to perform the conversion yourself (as ``IM.astype(float)``, for example).
```
DanHickstein, line 162: Bold indicates suggested changes:

"**The pixel intensities can have any value (within the floating-point range). However, the intensity scale must be linear.**"

"...cameras and common image formats **often** use" 

"...nonlinear intensity encoding. **Thus, if possible, it is recommended to disable the gamma correction on cameras used to record images that will be inverse Abel transformed. If this is not possible, then it is necessary to apply the appropriate intensity transformations before the analysis.**"

Also, what rounding issues arise with the int-to-float conversion?

DanHickstein (approved):
Looks good to me! I made some minor suggestions. Take or leave them.

DanHickstein:
>  Thanks, for "exposing" the subtle gamma intensity correction.

Hahahaha!

README.rst
MikhailRyazanov: Thanks!

MikhailRyazanov:
> Also, what rounding issues arise with the int-to-float conversion?

They sometimes arise _without_ this conversion.

I've noticed when working on #302 that, for example, `scipy.ndimage.shift()` with a fractional shift applied to an integer array returns also an integer array, apparently rounded:
```python
import numpy as np
from scipy.ndimage import shift

a = np.array([0, 1, 2, 1, 0])

print(shift(a, 0.5))
# [0 0 2 2 0]

print(shift(a.astype(float), 0.5))
# [0.     0.3125 1.6875 1.6875 0.3125]
```
which is probably not very useful. I haven't checked other functions, but there might be similar surprizes. And, well, the `abel.Transform()` docstring says: “Many images are imported in other formats (such as `uint8` or `uint16`), and this does not always play well with the transorm algorithms.”

DanHickstein:
Okay, I see how this could be an issue. In reality, the issue probably only manifests when the integers are low value. But, it's good that abel.Transform converts them to float. Potentially the individual transform functions (and other functions) should cast them to floats as well. But this would be a lot of work.
Anyway, I think that you can merge this.

Issue #307, opened 2020-11-21
Update testing


MikhailRyazanov:
We probably need to include the recently released Python 3.9 in the unit tests. And also make some clean-up there. For example, this:
https://github.com/PyAbel/PyAbel/blob/afc54ee18c50e68bd55700ce9b0e0ac0f07aadd8/.travis.yml#L5-L12
apparently tests _only_ Python 3.8, but does so _twice_.

And do we really need 32-bit tests for Windows Miniconda3?
https://github.com/PyAbel/PyAbel/blob/afc54ee18c50e68bd55700ce9b0e0ac0f07aadd8/appveyor.yml#L16-L27
Maybe they can be removed? (It [seems](https://docs.conda.io/en/latest/miniconda.html) that Miniconda with Python 3.9 is not available yet, so it should be added later...)

DanHickstein:
You can never be too certain that your software works on Py3.8 - always good to test twice :). 

But yes, you're right, probably we can swap one of those for a 3.9 test.

And I think that you're correct that most people using Py3 are on 64-bit. Maybe some people need 32-bit to interface with legacy hardware drivers? I feel like I've run into this issue before. But, I'm fine to drop all but one of the 32-bit test (Py 2.7) in the interest of faster appveyor builds and less complication.

MikhailRyazanov:
Yes, keeping the 32-bit Python 2.7 is useful for legacy stuff. I'm not sure that 32-bit drivers would have Python 3 interfaces, but we can leave one test just in case and remove another.

By the way, Ubuntu Xenial is rather old (released 2016-04-21, its support will end in half a year), it might be also a good idea to use the [latest LTS release](https://releases.ubuntu.com) (Focal Fossa) at least for Python 3.9.

Would you do the updates? I could, but the testing of this testing :–) might require deeper interaction with Travis and AppVeyor (I don't know what they have available and how it is called)...

DanHickstein:
I opened #309 to update the testing.

PR #308, opened 2020-12-01
State requirements before importing packages

prataffel:
Hey,
thanks for writing PyAbel - it's very convenient. When using ``tox`` to set up my environment, I get a ``ModuleNotFoundError`` (``No module named 'numpy'``) during the installation of the module. 
Defining the requirements before importing ``numpy`` in ``setup.py`` solves this issue for me.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "numpy >= 1.6"]
--- a/setup.py
+++ b/setup.py
@@ -9,13 +9,13 @@ from distutils.errors import CCompilerError, DistutilsExecError, DistutilsPlatfo
 # https://read-the-docs.readthedocs.io/en/latest/faq.html#how-do-i-change-behavior-for-read-the-docs
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 if not on_rtd:
-    import numpy as np
     install_requires=[
           "numpy >= 1.6",
           "setuptools >= 16.0",
           "scipy >= 0.14",
           "six >= 1.10.0"
           ]
+    import numpy as np
 else:
     np = None
     install_requires=[]


DanHickstein:
Hi @prataffel! Thanks for making this PR!

It looks fine to me. @stggh and @MikhailRyazanov, okay to merge this?

DanHickstein (approved):

stggh (approved):
Fine with me. Thanks for contributing.

MikhailRyazanov:
Hm, I don't know how all this works, but if `setup.py` is really a Python script, changing the order of `install_requires` and `import` should make no difference, since `install_requires` is just a declaration, used much later, in `setup()`. And as I see, `np` is used only for building Cython extensions, so perhaps this `import` should not exist there (in the very beginning of the file) at all... Can anybody suggest what to read about setup scripts (is this “setuptools”?), so that we can discuss and correct ours properly?

P.S. @DanHickstein will also need to update Python versions here and in README, in addition to #307. :–)

prataffel:
Hey. Thanks for reviewing this so fast -- but sadly, I was too impatient to test this properly. After setting up a clean virtualenv, ``python setup.py install`` will fail with the ``ModuleNotFoundError`` still. Sorry for that. @MikhailRyazanov, you are right, this does not work as I stated it, I solved my problem by accident.

prataffel:
The correct solution should be given in PEP 518 by specifying requirements in an additional file. 
If I setup a clean environment, ``pip install .`` now sets up PyAbel (and produces an Error without the ``pyproject.toml``). 

Note, that ``python setup.py install`` is still not working. This could be achieved with setuptools (see https://luminousmen.com/post/resolve-cython-and-numpy-dependencies and https://github.com/pypa/pip/issues/5761) but might be more of a workaround. 

Sorry again!

MikhailRyazanov:
@prataffel, adding the `pyproject.toml` file looks like a correct idea, and we can probably merge this as is, since this already helps.

But my impression is that most of the stuff before `setup()`, including that NumPy import, is needed only for Cython and thus should not be run at all if Cython is not available. And if it is available, all this should be executed *after* installing all “requires” (that is, from `setup()`). As I said, unfortunately I don't know how this can be done and never tried to build PyAbel Cython extensions myself, so I can't help here... But if you want to dig into it, it would be great. We can also try to summon @rth, since he wrote most of the setup script.

MikhailRyazanov (approved):

DanHickstein:
Okay, so it looks like the `pyproject.toml` file is indeed how we are supposed to tell pip what the requirements are. (At least according to [this article](https://snarky.ca/what-the-heck-is-pyproject-toml/).)

I can re-create the original issue by making a new environment with conda and then trying `pip install .` in the PyAbel directory. Indeed, numpy is not found. By adding the suggested `pyproject.toml` file, the installation goes ahead. (It is not necessary to modify the setup.py file.)

However, I get the following warning:   
```
WARNING: Missing build requirements in pyproject.toml for file:///Users/danhickstein/Documents/GitHub/PyAbel-DH.
WARNING: The project does not specify a build backend, and pip cannot fall back to setuptools without 'wheel'.
```

I can fix this issue by adding `build-backend = "setuptools.build_meta"`. I'm not really sure what this is doing, but I got the idea from here: https://setuptools.readthedocs.io/en/latest/build_meta.html

Also, using `pip install .`, the cython extensions are not built. I can get them to build if I include `"cython"` as a requirement in the `pyproject.toml` file. Here is what my `pyproject.toml` file looks like:

```
[build-system]
requires = ["setuptools", 
            "numpy >= 1.6",
            "cython"]

build-backend = "setuptools.build_meta"
```

But, maybe this will cause issues if Cython cannot easily be installed by pip. 

Also, I realize that we are being a bit silly about where we are locating the `import numpy as np` line in the `setup.py` file. The only place numpy is used in `setup.py` is `np.get_include()` on line 102 inside an `if _cython_installed` statement. So, we can simply move the numpy import up into the `try:` block (say to line 28). When I do this, it is not necessary to add the `pyproject.toml` file at all. This section of setup.py now looks like this:

``` python
if not on_rtd:
    # import numpy as np  # removed numpy import here
    install_requires=[
          "numpy >= 1.6",
          "setuptools >= 16.0",
          "scipy >= 0.14",
          "six >= 1.10.0"
          ]
else:
    np = None
    install_requires=[]



try:
    from Cython.Distutils import build_ext
    import Cython.Compiler.Options
    import numpy as np  # Moved numpy import to here
    Cython.Compiler.Options.annotate = False
    _cython_installed = True
```

MikhailRyazanov:
> But, maybe this will cause issues if Cython cannot easily be installed by pip.

I would also say that a hard dependence on Cython is very undesirable, since it is heavy, but is used only for making the “direct” method faster, while this method itself “is included mainly for educational and comparison purposes. In most cases, other methods will provide more reliable results and higher computational efficiency”. I mean, those who want to compare `direct_Python` and `direct_C` could install and configure Cython manually, but for everyone else it is just a useless burden.

Anyway, what is the supposed logic of our current `setup.py`? It should try to build Cython extensions when Cython is available, unless run by readthedocs.io? Then we probably can/need to move all this into `try ... Cython`, I guess.

MikhailRyazanov:
Another thought:
According to [“Writing the Setup Script”](https://docs.python.org/3/distutils/setupscript.html#preprocessor-options), the base directory for all Python headers can be obtained from `distutils.sysconfig.get_python_inc()`. And NumPy headers are [always](https://github.com/numpy/numpy/blob/master/numpy/lib/utils.py#L19-L46) (at least, for current versions) in `numpy/core/include` from there. So instead of trying to `import numpy` before it is installed, we can replace
```python
np.get_include()
````
by
```python
os.path.join(distutils.sysconfig.get_python_inc(), 'numpy', 'core', 'include')
```
Of course, if manually specifying these include directories is indeed needed.

Can anybody test whether Cython will work without `np.get_include()`, and if not, whether replacing this with `... get_python_inc() ...` and eliminating `import numpy` will solve the problem and work for all common systems (Linux/macOS/Windows)?

rth:
> whether Cython will work without np.get_include(), and if not, whether replacing this with ... get_python_inc() ... and eliminating import numpy will solve the problem and work for all common systems (Linux/macOS/Windows)?

I don't think it will. Cython needs numpy includes to build those extensions.I'm not aware of other packages that do this compilation conditionally, and I'm not sure if it's currently possible with `pyproject.toml`. In particular, specifying optional dependencies in pyproject.toml is still not supported in pip https://github.com/pypa/pip/issues/6144 

As far as I can tell there are 3 workarounds,
 - keep using the previous system (without pyproject.toml) with conditional compilation
 - add `pyproject.toml` (with numpy and cython build dependencies) and then always compile this extension. To make this user friendly (particularly for Windows users), binary wheels should be created in CI and uploaded to PyPi for releases (using e.g. https://github.com/joerick/cibuildwheel) which means that users won't have to install Cython, nor compile the package when they install it from PyPi. Or just suggest to install from conda-forge. 
 - rewrite the corresponding code in numba and add that as as a dependency, to remove the need for Cython.

MikhailRyazanov:
@rth,
> Cython needs numpy includes to build those extensions.

But it can find “regular” Python includes without explicitly specifying the path, right? And NumPy includes are in the same directory, in the `numpy` subdirectory. Isn't it added automatically by `cimport numpy`?

In any case, using `distutils.sysconfig.get_python_inc()` *should* work.

> keep using the previous system (without pyproject.toml) with conditional compilation

This did not work at all without `numpy` preinstalled, while `pyproject.toml` at least allows it to work with `pip`. I guess that avoiding `np.get_include()`, as described before, should solve the problem for direct execution of `setup.py` as well, but this needs to be tested.

> with numpy and cython build dependencies ... binary wheels should be created in CI and uploaded to PyPi for releases ... Or just suggest to install from conda-forge.

Looks like extra troubles both for us and for users, without noticeable benefits. :–(

> rewrite the corresponding code in numba and add that as as a dependency, to remove the need for Cython

Well, adding an *optional* Numba implementation would be interesting, but adding Numba as a mandatory requirement does not look any better than Cython...

DanHickstein:
@MikhailRyazanov, yes, I agree that we should have the direct-C installation be completely optional.

I think that we can successfully use our setup.py file (with the minor modification moving `import numpy` into the try statement). Then we can install with pip. And we don't even need the `pyproject.toml` file. I think that this is all we really need to do here, right?

We should add something to the documentation (probably [here](https://pyabel.readthedocs.io/en/latest/abel.html#module-abel.direct)?) telling users that they need to have numpy, Cython, and a c-compiler pre-installed in order to use the direct-C method.

In the long run, it might be a fun project to see if we can port direct-C from Cython to Numba. I think that Numba is quite an improvement, because users don't need to install a C-compiler, they just need to install Numba. Big improvement, right? Low priority and not something that we need to deal with soon though.

Good to hear from you @rth! I hope that you're doing well.

DanHickstein:
So, I think what we decided is that we basically like the strategy of using a `try` statement in setup.py to import cython and numpy. If they are installed, then we go ahead and try to build the cython extensions (which is another try statement that is allowed to fail, typically if the user doesn't have a C-compiler installed). Users that want to have direct-C installed need to be savvy enough to have cython, numpy, and a C-compiler installed. 

[This seems okay, since only 0.1% of users should care about direct-C (most users are better served with other, faster, transform methods and if they do want to test the direct methods, the direct-Python should be fast enough for most cases).]

The only real issue is that, currently, the `try` block in setup.py checks only for cython, and not for numpy. Basically we just need to include a numpy import in that `try` block.

Does that sound good? @MikhailRyazanov, @rh?

I can go ahead and do that (and make a separate PR), unless @prataffel would rather modify this PR to do so.

MikhailRyazanov:
@DanHickstein, yes, it would be much better to rewrite `setup.py` with a single `try` block for all Cython-related stuff and explain in [README / Installation](https://github.com/PyAbel/PyAbel#installation) that Cython implementation of the "direct" method requires NumPy and Cython to be preinstalled before installing PyAbel. (By the way, what happens when `setup.py` or `pip install --upgrade --force-reinstall` is executed with already installed package? Will this [re]build Cython extensions if it is available at that time?)

DanHickstein:
> yes, it would be much better to rewrite setup.py with a single try block

Sounds good, I'll have a crack at it.

> By the way, what happens when setup.py or pip install --upgrade --force-reinstall is executed with already installed package? Will this [re]build Cython extensions if it is available at that time?

Yes, I think that it will re-build the Cython extensions.

DanHickstein:
I couldn't find a natural spot in the Readme to include the discussion about the Cython extensions. And, I didn't want to "pollute" the readme with information that will only be useful to 0.01% of users. So, I buried the info about the Cython build requirements in a docstring in direct.py. Happy to move it if there is a better location.

prataffel:
Thanks for looking into this! I will close the PR in favor of #310.

DanHickstein:
Thanks for bringing this to our attention!

PR #69, opened 2016-01-13
Adding Sphinx documentation for PyAbel

rth:
Following issue #25, this PR adds a draft documentation for PyAbel using Sphinx.

As suggested by @DanHickstein this setup uses the readthedocs.org theme, and [Napoleon](https://sphinxcontrib-napoleon.readthedocs.org/en/latest/index.html) for Numpy style docstrings. 

In addition, this setup uses the [`recommonmark`](https://github.com/rtfd/recommonmark) package to process either .rst or .md documentation files. Markdown for some parts of the documentation would be nice, and we could then potentially reuse the same files between Sphinx and the github wiki.

To build the documentation locally you need to have `Sphinx` and `recommonmark` installed, and the html pages can be generated with:
   
     cd PyAbel/doc/
     make html

The html files currently produced with this command are accessible at http://r0h.eu/it/static/d/PyAbel/index.html although more work is necessary in order to ensure that all docstings are compatible with Numpy style and to add some general explanations. 

In addition, more work would be necessary to automate deployment of the documentation somewhere (for instance to github pages).  I would like to it this latter point before merging this PR.

@DanHickstein Sorry for not responding earlier in issue #25, I completely agree with your comments there. 

**Edit:**  [`sphinx_rtd_theme`](https://github.com/snide/sphinx_rtd_theme/)  also needs to be installed.

PR #310, opened 2021-01-05
Moving numpy import into try block in setup.py

DanHickstein:
Following the discussion in #308, we decided that we should move the `import numpy as np` line into the `try` block in setup.py. Numy is required for the `np.get_include()` command to build the Cython extensions.

This should allow pip to successfully install PyAbel, even in situations where numpy was not previously installed.

I'll work on adding something to the readme to discuss the Cython+numpy+C-compiler for building the direct-C.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,6 @@ import os.path
 from setuptools import setup, find_packages, Extension
 from distutils.errors import CCompilerError, DistutilsExecError, DistutilsPlatformError
 
-
 # a define the version string inside the package
 # see 
 # https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package
@@ -13,33 +12,19 @@ VERSIONFILE = "abel/_version.py"
 verstrline = open(VERSIONFILE, "rt").read()
 VSRE = r"^__version__ = ['\"]([^'\"]*)['\"]"
 mo = re.search(VSRE, verstrline, re.M)
+
 if mo:
     version = mo.group(1)
 else:
     raise RuntimeError("Unable to find version string in %s." % (VERSIONFILE,))
-    
-# change behaviour of the setup.py on readthedocs.org
-# https://read-the-docs.readthedocs.io/en/latest/faq.html#how-do-i-change-behavior-for-read-the-docs
-on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
-if on_rtd:
-    np = None
-    install_requires = []
-
-else:
-    install_requires = [
-          "numpy >= 1.6",
-          "setuptools >= 16.0",
-          "scipy >= 0.14",
-          "six >= 1.10.0"]
-
-try:  # try to import numpy and Cython to build Cython extensions
+# try to import numpy and Cython to build Cython extensions:
+try:  
     import numpy as np
     from Cython.Distutils import build_ext
     import Cython.Compiler.Options
     Cython.Compiler.Options.annotate = False
     _cython_installed = True
-
 
 except ImportError:
     _cython_installed = False
@@ -112,7 +97,10 @@ setup(name='PyAbel',
       url='https://github.com/PyAbel/PyAbel',
       license='MIT',
       packages=find_packages(),
-      install_requires=install_requires,
+      install_requires=[ "numpy >= 1.6",
+                         "setuptools >= 16.0",
+                         "scipy >= 0.14",
+                         "six >= 1.10.0"],
       package_data={'abel': ['tests/data/*']},
       long_description=long_description,
       classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -5,42 +5,40 @@ import os.path
 from setuptools import setup, find_packages, Extension
 from distutils.errors import CCompilerError, DistutilsExecError, DistutilsPlatformError
 
-# change behaviour of the setup.py on readthedocs.io
+# change behaviour of the setup.py on readthedocs.org
 # https://read-the-docs.readthedocs.io/en/latest/faq.html#how-do-i-change-behavior-for-read-the-docs
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
-if not on_rtd:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -5,42 +5,40 @@
...
 
-# change behaviour of the setup.py on readthedocs.io
+# change behaviour of the setup.py on readthedocs.org
 # https://read-the-docs.readthedocs.io/en/latest/faq.html#how-do-i-change-behavior-for-read-the-docs
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
-if not on_rtd:
-    import numpy as np
-    install_requires=[
+
+if on_rtd:  # if we're building on readthedocs.org:
```
MikhailRyazanov, line 12: Is this needed? RTD does run examples, so it needs all these modules anyway, so why exclude them from requirements here?

DanHickstein: I don't remember why this is the way it is. I can try removing this and see what happens.

DanHickstein: Huh, seems to build just fine without the RTD difference. Maybe we can remove the RTD-checking from setup.py completely.

DanHickstein: This code was introduced in PR #69. Apparently readthedocs has some issues with c-dependent libraries.
https://docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules

They say: 
>With Sphinx you can use the built-in autodoc_mock_imports for mocking. If such libraries are installed via setup.py, you also will need to remove all the C-dependent libraries from your install_requires in the RTD environment.

So, what we were doing seems to be correct given their advice. I'm unsure why it _doesn't_ cause a problem to include numpy in the `install_requires`, since their docs imply that it should. 

But, the docs seem to build fine anyway. 

PS: Back in 2016, it seems that I tried to do this same thing, and ended up switching back to the way things were. No idea why! :) https://github.com/PyAbel/PyAbel/commit/107170810f0f86aeec21285fc5f6bb116bdd53ad

MikhailRyazanov: If RTD does not like C, then we should disable only Cython extensions (which should happen automatically, unless RTD somehow has Cython preinstalled by default, but even then there is a check `if _cython_installed and not on_rtd` later...), but all other dependences _are_ needed on RTD. So I would say that this can and better be removed.

MikhailRyazanov:
... and can we please move everything else Cython-related
https://github.com/PyAbel/PyAbel/blob/86884e9a4903c592557702678c7ff80e7da2f82b/setup.py#L57-L108
into that `try` block? Perhaps, with `import sys` and `import os...`, which are not needed otherwise.

setup.py
DanHickstein: Sorry, should have clarified: Numpy uses C libraries and is considered a "C-dependent library" in this context.

DanHickstein:
So, I moved everything into the try block. Now I'm re-thinking that. Try-except statements are kinda dangerous, because they make it more difficult to track down errors. I think that the best-practice is to put only the statements that you think are likely to fail in the try block and to keep everything else outside of it where errors can work as normal. I'm going to change it a little bit, but hopefully you'll still like it...

DanHickstein:
Okay, see what you think about the latest draft. I moved some things around to make the logic more obvious, but I kept the try block to only importing numpy and cython, since these are the only lines that we expect to fail. Everything else is outside of the black-magic of the try block.

MikhailRyazanov:
Yes, I think, doing Cython testing in a `try` block and moving everything else inside `if _cython_installed:` makes sense (although there shouldn't be 2 empty lines before `except ImportError:`). Looks much better than before. :–)

Regarding RTD, it indeed seems that they don't allow to install NumPy, but they can provide it preinstalled (which you do, I guess). So including all these `install_requires` should not harm on RTD, since if that environment is prepared correctly, then everything would work as expected. And if something wrong happens with the RTD configuration, it's probably better to get an error at the PyAbel installation step rather than somewhere deep inside when it tries to use unavailable NumPy. (If we want to be able to build the documentation without figures, this must be done more carefully.) I'd suggest to remove that `if` and instead add a short legible comment how RTD needs to be configured.

DanHickstein:
I think that the docs build fine on RTD even if we remove all of the RTD-specific code from setup.py. So, even though some of the documentation on RTD implies that we need to take special steps, it seems that perhaps we no longer need to do this. It would make sense that RTD would improve the situation. Lots of people use numpy after all. So, we can simplify setup.py.

DanHickstein:
Whoops, I left this perfectly good PR sitting for two months. Time flies when you're trying to build and ship laser systems around the world on tight deadlines in the middle of a pandemic :)

As far as I can tell, this PR is good-to-go, clarifies the logic in setup.py, and allows PyAbel to build in situations where numpy is not already installed.

@stggh and @MikhailRyazanov, what do you think? Should we merge this?

stggh (approved):
These changes appear to work. I have built `PyAbel` without`cython`:
image
and with cython:
image

There may be a couple of issues:
(1) Message above states something about setting cython language level?
(2) A depreciation warning:
image

My system: Linux Gibbo 5.8.0-45-generic #51-Ubuntu SMP Fri Feb 19 13:24:51 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux.  Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz


Otherwise, AFAICT all good.

DanHickstein:
Thanks for checking this @stggh! 

I think that those warnings are the same as before this PR, right? I don't think that any of the changes that I made should have affected the compilation process. (Not saying that we shouldn't look into them, just wondering if I changed anything here.)

stggh:
Yes, they were present before this PR. I should have mentioned that.

MikhailRyazanov (approved):

MikhailRyazanov:
Finally! :–)

DanHickstein:
Merged! Thanks :)

I'll open a new issue for us to look into the warnings during the Cython compile.

PR #319, opened 2021-04-28
New functions for angular integration and average

MikhailRyazanov:
Here are the new functions:
* `angular_integration_2D()` and `angular_integration_3D()` to replace `angular_integration()`,
* `average_radial_intensity_2D()` and `average_radial_intensity_3D()` to replace `average_radial_intensity()`.
They are in fact implemented through a new function `radial_intensity()`, which can do all of this.

The old functions are declared deprecated and removed from the rest of the code. Hopefully, completely... This took most of the effort. :–) In this process, I've discovered one **problem**: the `angular_integration` option in `transform.Transform()` was using that incorrect `angular_integration()` function and thus also produced incorrect results. I've [mentioned this in the docstring](https://github.com/MikhailRyazanov/PyAbel/blob/d8173987ac044649216788482be8c34f93150bfa/abel/transform.py#L132-L137), but nothing can be done for backward compatibility in this case... Except renaming this option (and the `angular_integration` attribute, which hold the results), like we did with the `center` → `origin` transition. Any ideas?

P.S. Various minor things that produced warnings (due to incorrectly formatted docstrings and changes in Matplotlib and NumPy) are also corrected. Now  AppVeyor and Read the Docs build without warnings.

### DIFF
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -140,6 +140,16 @@ def angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None):
     Note: the use of ``Jacobian=True`` applies the correct Jacobian for the
     integration of a 3D object in spherical coordinates.
 
+    .. warning::
+        This function behaves incorrectly: misses a factor of π for 3D
+        integration, with ``Jacobian=True``, and for ``Jacobian=False`` returns
+        the *average* (over polar angles) multiplied by 2π instead of
+        integrating. It is currently deprecated and is provided only for
+        backward compatibility, but will be removed in the future.
+
+        Please use :func:`radial_intensity`, :func:`angular_integration_2D` or
+        :func:`angular_integration_3D`.
+
     Parameters
     ----------
     IM : 2D numpy.array
@@ -175,6 +185,8 @@ def angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None):
         integrated intensity array (vs radius).
 
     """
+    _deprecate('angular_integration() is deprecated, '
+               'please see the documentation for appropriate alternatives.')
 
     polarIM, R, T = reproject_image_into_polar(
         IM, origin, Jacobian=Jacobian, dr=dr, dt=dt)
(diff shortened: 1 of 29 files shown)


MikhailRyazanov:
A question, probably to @stggh, which also can be addressed in this PR:
I think, `abel.tools.vmi.radial_integration()` needs somewhat better documentation. I don't understand what exactly is meant by “intensity” and “amplitude” there.

abel/tools/vmi.py
```diff
@@ -60,6 +185,8 @@ def angular_integration(IM, origin=None, Jacobian=True, dr=1, dt=None):
         integrated intensity array (vs radius).
 
     """
+    _deprecate('angular_integration() is deprecated, '
```
DanHickstein, line 188: Why not suggest angular_integration_3D or angular_integration_3D specifically here?

DanHickstein (approved):
Wow! Lots of good stuff in here. Thanks Mikhail!

abel/tools/vmi.py
MikhailRyazanov: Done!

stggh (approved):
Sorry for the delay, I am busy preparing for a June outback trip, and as I am not the sole Australian undertaking this task, it is hard to source equipment.

The changes look good to me, and a quick check finds the O<sub>2</sub><sup>-</sup> spectrum radial intensities (`example_O2_PES_PAD.py`) appear to match those of Lineberger's non-imaging spectrometer spectrum, so all good (no missing radial factor).

In relation to `vmi.radial_integration`: `intensity` refers to the image pixel intensity (left subfigure), and also the integrated intensity (middle), and `amplitude` is the height of the <math>1 + &beta; P<sub>2</sub>(cos&theta;)</math> function (right subfigure):
Screenshot-20210501130721-1494x435

I agree this could be explained better. I will think about this later, otherwise feel free to include your own correct terminology.

MikhailRyazanov:
Thanks, @stggh!
I'll open a separate issue to discuss `radial_integration()` (need to collect some more information)...

DanHickstein:
Well done Mikhail!

> Sorry for the delay, I am busy preparing for a June outback trip, and as I am not the sole Australian undertaking this task, it is hard to source equipment.

What does an "outback trip" entail? Is it an extended camping trip?

stggh:
Yes,  4 wheel-drive "car" camping in central Australia, away from everything. 
e.g.
https://traveloutbackaustralia.com/oodnadatta-track.html/

PR #324, opened 2021-07-20
Corrections to Transform()

MikhailRyazanov:
1. Some `dict` parameters of `abel.transform.Transform` had incorrectly documented types (tuples really don't work there).
2. With `symmetry_axis=(0, 1)`, instead of transforming 1 quadrant, it was calling underlying transform methods for 3 quadrants (but using only 1 of them).


DanHickstein (approved):
Looks good to me! Thanks Mikhail!

Issue #245, opened 2019-03-01
Adding Tikhonov regularization 


jahdreyer:
Hi,

I started using the PyAbel package and it works very well. I was just wondering if you considered adding Tikhonov regularization to the other Abel transforms available in the PyAbel package. My understanding is that currently only BASEX uses Tikhonov regularization. I previously applied Tikhonov regularization to Onion Peeling following the description by Daun et al.:
https://www.osapublishing.org/ao/abstract.cfm?uri=ao-45-19-4638

For the flame images I work with, this significantly improved the reconstruction (i.e., less noisy result). For me, BASEX works fine but for others Tikhonov regularization of a faster transform might be useful.

Thanks for the great package and regards,
Jochen

MikhailRyazanov:
Now BASEX is among the fastest transforms (see Fig. 5 in https://arxiv.org/abs/1902.09007). Its only problem is the relatively slow generation of the basis sets, but this needs to be done only once.

I need to read the Daun's article carefully, but from a quick glance it seems that they are using a Dasch-like approach, but with the regularized inversion of the forward-transform matrix instead of directly computing the inverse-transform matrix. In principle, adding this to the Dasch methods would not be difficult, but the resulting speed will be the same as in BASEX (except the faster basis generation).

MikhailRyazanov:
@jahdreyer, you are welcome to test #326!

MikhailRyazanov:
OK, the Daun et al. method has been implemented in #326, so I'm closing this issue. Thanks, @jahdreyer, for suggesting it!

DanHickstein:
Yes! Thanks a lot to @jahdreyer for suggesting this and to @MikhailRyazanov for implementing it! Good work!

Issue #306, opened 2020-10-16
Propragation of uncertainty


gh4ag:
Hello,

I started looking at your code and doc. I have seen the impressive amount of work done on improving the robustness and the speed of the code.

I couldn't find much information about the possibility to propagate uncertainties through the code though.

>In what follows, I am going tu reuse the same notation as the one used in the article your team has published. Capital letter for the projection (forward Abel tranform) and small caps for the radial distribution (inverse radial transform). 

**Question**

Let's say I measure <img src="https://render.githubusercontent.com/render/math?math=F(y,z) \pm \Delta F(y,z)">. How would this uncertainty <img src="https://render.githubusercontent.com/render/math?math=\Delta F(y,z)"> affect the output <img src="https://render.githubusercontent.com/render/math?math=f(r,z)">  of the transform ? Especially where <img src="https://render.githubusercontent.com/render/math?math=\frac{dF}{dy}(y,z)"> is close to zero.

**Remark**

If we take the functions <img src="https://render.githubusercontent.com/render/math?math=G_{\pm}(y,z) = F(y,z) \pm  \Delta F(y,z)"> and apply the inverse transform on each, I am not sure that the quantity <img src="https://render.githubusercontent.com/render/math?math=\lvert g_{%2B} (r,z) - g_{-}(r,z) \rvert"> would be representative of the uncertainty that is induced on the computation of <img src="https://render.githubusercontent.com/render/math?math=\frac{dF}{dy}(y,z)">... Indeed the derivative may be computed or approximated using different approaches for each algorithms.

> It is not really an issue, more a question.

Regards

MikhailRyazanov:
Since PyAbel works with discrete pixels instead of functions (and their derivatives, which are not even well-defined for noisy data), it's better to consider the transform as a matrix problem and use the well-known formula (see [here](https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Linear_combinations)) for propagation of uncertainty through linear transforms.

For the raw original data, in many cases the covariance matrix can be assumed diagonal (no correlations among pixels), but the Abel transforms introduce huge correlations (positive for the forward and negative for the inverse) for adjacent output pixels, so these output uncertainties must be treated very carefully. I mean, considering only “standard deviations” for each pixel value would be misleading, and their full covariance matrix must be taken into account.

If you provide more specific details about your data, methods and final goals, I would probably also suggest something more practical...

DanHickstein:
[...]

PR #326, opened 2021-07-24
New method by Daun et al. + extensions

MikhailRyazanov:
Implementation of the method described in

K. J. Daun, K. A. Thomson, F. Liu, G. J. Smallwood, “Deconvolution of axisymmetric flame properties using Tikhonov regularization”, Appl. Opt. 45, 4638–4646 (2006), [doi:10.1364/AO.45.004638](https://doi.org/10.1364/AO.45.004638).

The original method basically applies Tikhonov regularization to the “onion peeling” approach. Conceptually similar to BASEX, but instead of the _L_<sub>2</sub> norm, uses the derivative operator (in fact, the finite-difference matrix, since the problem is discrete), which also helps to suppress oscillations, although does this somewhat differently, maybe even better.

This addresses #245, but more importantly, implements some useful extensions:
1. Different basis functions can be used, for direct comparison of their effect on the results.
2. Both “difference” (from the original) and “_L_<sub>2</sub>” (from BASEX) regularizations are implemented for direct comparison.
3. The non-negative least-squares solution can be used as a special “regularization” approach. It is very slow compared to everything else in PyAbel, but produces impressively clean results for noisy data:
nonneg
(and the efficiency grows as the SNR drops; look also at `daun (nonneg)` in the [“comb” comparison](https://pyabel--326.org.readthedocs.build/en/326/transform_methods/comparison.html#fig-comb)).

With default parameters (without regularization) it is identical to `onion_peeling`, but should run about twice faster (at least, does so for sufficiently large images on my system) due to using back substitution instead of matrix multiplication (here comparing with `two_point` as the fastest current method):
benchmarks
With Tikhonov regularization, it should behave like other matrix-multiplication methods (`dasch` and `basex`).
(I'll need @stggh's help to run the benchmarks on his i7 — after we make sure that everything works properly...)


MikhailRyazanov:
A technical question... The readthedocs theme renders equation numbers _above_ the equation instead of on the right (see [here](https://pyabel--326.org.readthedocs.build/en/326/transform_methods/daun.html#equation-forward), for example). This is apparently a known issue (https://github.com/readthedocs/sphinx_rtd_theme/issues/301), but nobody cares. The default sphinx theme in fact has a CSS rule to render this properly:
```CSS
span.eqno {
    float: right;
}
```
but it is not used. Any ideas how to add it to our documentation on RTD?

stggh:
@MikhailRyazanov this is impressive. I am still travelling Oz, currently in the Northern Territory, and just happen to be at Yallara/Uluru where there is an internet connection. My return to Canberra will now not be until early September, so the i7 speed testing may need to wait.

MikhailRyazanov:
Return safely, @stggh! :–) This is not urgent at all and surely can wait until September. Will you have a chance (and interest) to do some testing and reviewing before that?

DanHickstein:
Wow! This is really exciting! It seems like this new method may be the fastest and most versatile method in PyAbel.

Looking quickly through the code, it looks excellent. I should have time to do some more detailed code review next week.

MikhailRyazanov:
Yes, I also like it (even in the original form) more than BASEX. And was again surprised how different fields are disconnected: this method it is not (widely) known in the VMI community, and Daun doesn't even mention BASEX...

stggh:
I have finally returned to Canberra, into 2 weeks of quarantine. The ACT is also under COVID lockdown, along with NSW and Vic., so no more travel for a while. My Outback trip was fantastic. Google photo heat map gves some indication of the journey:
              googlephotos_shotmap
Edith falls, NT (a favourite location):
Edith_falls

I will send some details via email (later).


@MikhailRyazanov, @DanHickstein you have been very busy. The new Daun method is fast and neat. I will review the code and run the benchmarks on my systems.


Here is a quick benchmark `abel.benchmark.AbelTiming(n=[301, 501, 1001, 2001])` on my Dell PC:
system

running Ubuntu 21.04, python 3.9.5
```
=== Basis generation ==========================================

Method              n = 301     n = 501    n = 1001    n = 2001
---------------------------------------------------------------
basex                 191.5       634.7      3731.9     25256.2
daun                    1.2         2.1         4.6        10.7
linbasex                4.1        11.2        42.9       170.3
onion_peeling           0.6         1.4         4.7        24.1
rbasex                  4.0         7.2        19.7        56.8
three_point             1.8         4.2        15.3        83.2
two_point               0.5         1.2         4.6        27.8


=== Forward Abel transform ====================================

Method              n = 301     n = 501    n = 1001    n = 2001
---------------------------------------------------------------
basex                   0.1         0.2         1.5        13.1
basex(var.reg.)         1.3         2.2         7.3        42.1
daun                    0.1         0.3         1.8        14.8
direct_C                4.1        16.1       126.5       927.9
direct_Python          44.2       166.1      1168.9     12382.8
hansenlaw               3.1         7.5        27.3        96.3
rbasex                  0.7         1.7         7.3        34.1
rbasex(None)            0.4         1.0         4.2        17.5


=== Inverse Abel transform ====================================

Method              n = 301     n = 501    n = 1001    n = 2001
---------------------------------------------------------------
basex                   0.1         0.2         1.5        13.1
basex(var.reg.)         1.3         2.1         7.7        44.3
daun                    0.1         0.3         1.9        15.0
daun(var.reg.)          0.5         1.3         6.1        40.3
direct_C                4.6        16.8       130.7       946.0
direct_Python          44.7       176.9      1179.7     12782.4
hansenlaw               3.4         8.2        31.4       106.0
linbasex               28.7        78.3       317.2      2144.6
onion_bordas           84.3       264.0      1268.2     11769.6
onion_peeling           0.1         0.2         1.5        16.1
rbasex                  0.7         1.7         7.4        32.7
rbasex(None)            0.5         1.0         4.2        17.5
three_point             0.1         0.2         1.5        13.3
two_point               0.1         0.2         1.5        13.1
```

doc/transform_methods/daun.rst
```diff
@@ -0,0 +1,332 @@
...
+""""""""""""""""
+
+In addition to the original derivative (difference) Tikhonov regularization,
+PyAbel also implements the usual :math:`L_2` regularization, as in
+:ref:`BASEX`, with the identity matrix :math:`\mathbf I` used instead of
+:math:`\mathbf L` in :eq:`tikhonov`. The results are practically identical to
+the BASEX method, especially with **degree** = 2, except that the basis set is
+computed much faster.
+
+Non-negativity
```
stggh, line 189: I have mixed feelings about this option. It is fundamentally wrong, produces incorrect results for the background intensity, only to give the illusion of smoothing. However, it reproduces the old PC `basex` output, so it may have some use.

Here is the `O2-ANU2048.txt` (larger image size to increase noise) example, note incorrect background between transitions for the `nonneg` option:
image

stggh (approved):
@MikhailRyazanov you do not disappoint - amazing coding, with excellent results. Timings to follow.

stggh:
Timing results
===========
All timing benchmarks in this [gist](https://gist.github.com/stggh/128da21f1cbfd30ded6909dbfc0609b0)

Not all methods run (as they should be similar to previous runs).


i7-9700-Linux-Anaconda3.9.7 (Dell OptiPlex 7070)
-------------------------------------------------------------------
```
Name:       Intel Core i7-9700
uArch:      Coffee Lake
Technology: 14nm
Max Freq:   4.700 GHz
Cores:      8 cores
AVX:        AVX,AVX2
FMA:        FMA3
L1i Size:   32KB (256KB Total)
L1d Size:   32KB (256KB Total)
L2 Size:    256KB (2MB Total)
L3 Size:    12MB
Peak Perf.: 1.20 TFLOP/s

Anaconda Python 3.9.7

Linux Gibbo 5.11.0-31-generic #33-Ubuntu SMP Wed Aug 11 13:19:04 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Distributor ID:	Ubuntu
Description:	Ubuntu 21.04
Release:	21.04
Codename:	hirsute

32GB memory
```

i7-9700-Linux-Anaconda3 9 7
e.g.
``` 
# daun            iabel(ms)           basis(ms)
    5             0.01711             0.02373
    9             0.01955             0.03707
   17             0.02057             0.06232
   33             0.02293             0.11695
   65             0.02886             0.22029
  129             0.04329             0.43562
  257             0.10706             0.90285
  513             0.42341             1.89876
 1025             4.23736             4.05734
 2049            19.78891             9.56412
 4097           109.93354            27.52181
 8193           727.86471            74.85047
16385          4903.41616           227.30092
32769         37612.45776           818.95221
```
i7-6700-Linux-Anaconda3.9.7 (Dell OptiPlex #)
--------------------------------------------------------------
```
Name:       Intel Core i7-6700
uArch:      Skylake
Technology: 14nm
Max Freq:   4.000 GHz
Cores:      4 cores (8 threads)
AVX:        AVX,AVX2
FMA:        FMA3
L1i Size:   32KB (128KB Total)
L1d Size:   32KB (128KB Total)
L2 Size:    256KB (1MB Total)
L3 Size:    8MB
Peak Perf.: 512.00 GFLOP/s

Anaconda python 3.9.7

Linux etalon 5.8.0-55-generic #62-Ubuntu SMP Tue Jun 1 08:21:18 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Distributor ID:	Ubuntu
Description:	Ubuntu 20.10
Release:	20.10
Codename:	groovy

32GB memory
```
i7-6700-Linux-Anaconda3 9 7
```
# daun            iabel(ms)           basis(ms)
    5             0.01801             0.02375
    9             0.03927             0.03721
   17             0.04102             0.06304
   33             0.04390             0.11677
   65             0.05894             0.22311
  129             0.08766             0.44174
  257             0.20657             0.92827
  513             0.84640             1.98533
 1025             4.85983             4.27869
 2049            35.90545             9.95490
 4097           201.18627            27.41542
 8193          1272.86124            76.51444
16385          9042.48185           242.99418
32769         69927.85735           857.24778
```

i7-6920HQ-Mac-Anaconda3.9.7 (Macbook pro)
---------------------------------------------------------------
```
Name:              Intel Core i7-6920HQ
Microarchitecture: Skylake
Technology:        14nm
Max Frequency:     3.800 GHz
Cores:             4 cores (8 threads)
AVX:               AVX,AVX2
FMA:               FMA3
L1i Size:          32KB (128KB Total)
L1d Size:          32KB (128KB Total)
L2 Size:           256KB (1MB Total)
L3 Size:           8MB
Peak Performance:  486.40 GFLOP/s

Mac OSX 11.5.2

Anaconda Python 3.9.7

16GB memory
```
i7-6920HQ-Mac-Anaconda3 9 7
```
# daun            iabel(ms)           basis(ms)
    5             0.03110             0.03354
    9             0.04050             0.05106
   17             0.03582             0.08438
   33             0.03876             0.16036
   65             0.04841             0.30764
  129             0.07312             0.55529
  257             0.17056             1.28726
  513             0.71052             2.58375
 1025             4.33280             5.42697
 2049            30.17282            12.79168
 4097           191.89015            33.11970
 8193          1643.85077           152.99867
16385         10593.66686           462.27430
32769         97123.04623          1708.42080
```

doc/transform_methods/daun.rst
MikhailRyazanov: Sorry, what is fundamentally wrong?

stggh: The non-negativity option (`reg='nonneg'` parameter) limits the inverse Abel transform to positive values. This option may result in an under correction for low pixel intensities, as shown in the figure above (blue line vs orange). This demonstration is for good quality VMI data, the effect is much more dramatic for lesser quality images, and the effect varies for the particular spectral structure.  "Fundamentally wrong" in that the option results may produce an incorrect spectral profile. The old PC `basex.exe` produced only positive valued inverse Abel transforms (images). 

My apology if I have misunderstood what this option does.

MikhailRyazanov:
@stggh,
> "Fundamentally wrong" in that the option results may produce an incorrect spectral profile. The old PC `basex.exe` produced only positive valued inverse Abel transforms (images).

The old BASEX program was performing a linear transform and then zeroing the negative values in its results. This is indeed fundamentally wrong.

However, it is important to understand that applying the inverse Abel transform directly to event-counted experimental data is also “fundamentally wrong” because such data _are not a forward Abel transform_ of anything meaningful but instead a _random sample from a probability distribution_ parametrized by the forward Abel transform of some distribution. So the correct approach would be to use some sort of statistical inference. If we don't know anything except that the underlying distribution is non-negative (which is true for the intended use of this method) and that the observed counts are described by Poissonian statistics (which is very close to reality), then it can be shown that the maximum-likelihood estimation of that distribution is obtained from some non-linear equations, which are basically equivalent to a properly weighted non-negative least-squares problem (see section 3.3.3 in [my dissertation](https://digitallibrary.usc.edu/asset-management/2A3BF169XWB4)).

So what this `reg='nonneg'` does is not “fundamentally wrong” but in fact is an approximation to the MLE method — without proper weighting, but reasonably close (for a uniform Gaussian noise, however, this would be the exact MLE).

Here is some example for a lesser-quality image ([code](https://gist.github.com/MikhailRyazanov/0a42709ee5721b53ac39eb8261bf28df)):
images
`source` is a synthetic distribution (`'Ominus'`).
`sample projected` is a random Poissonian sample from its projection, simulating low-intensity experimental data.
`daun` is the usual linear transform, very noisy.
`daun[>0]` is the same but with all negative intensities zeroed (like the old BASEX did), still very noisy and in fact looking very similar to the non-transformed image.
`daun(nonneg)` is the non-negative least-squares solution, now much more resembling the underlying structure of `source` — although also quite noisy, but without so much crazily oscillating background.
`rbasex(pos)` is the rBasex result (also with the non-negativity constraints), for comparison.

So `daun(nonneg)` in fact produces much more reasonable _images_.

If we examine separate rows of these images, there is a similar picture (here for the central row):
1D
`daun(nonneg)` does more or less reproduce the expected peaks and doesn't have the huge oscillations that mask the real structure in the linear transform, although a _small_ number of spurious peaks is still present. Just zeroing negative intensities in the linear results of course doesn't help much. See also the [comb example](https://pyabel--326.org.readthedocs.build/en/326/transform_methods/comparison.html#fig-comb) in the documentation.

The radial intensity distribution
I(r)
for `daun(nonneg)` indeed has the peaks partially spread into spurious background, but even this is significantly less wrong than the radial distribution for `daun[>0]`! (The old BASEX actually did calculate radial distributions before removing negative intensities from the image, so its result would look like the `daun` plot, although with zeroed negative values.)
And the problem is that, as I said, this method assumes that _nothing_ is known about the underlying distribution. But if we know that it was composed of spherical harmonics, we can use the same approach with this additional information and get the `rbasex(pos)` results, which are much better, perhaps even better that the linear results (while `rbasex` produces a relatively clean image even without regularization, its radial distribution (not shown for clarity) is very similar to that of `daun`).

So ideally, when a reasonably detailed model exists, its parameters should be fitted either directly to the experimental data or to the distributions obtained from any of the _linear_ methods (mandatorily using its covariance matrix, which we have [not yet implemented](https://github.com/PyAbel/PyAbel/issues/306)). But when no model is known beforehand, this new non-negative method produces the best available images (or 1D results) to look at an try to see what's going on there...

doc/transform_methods/daun.rst
MikhailRyazanov: I've answered [below](https://github.com/PyAbel/PyAbel/pull/326#issuecomment-915483616).

stggh:
**You are right,** this option is excellent for revealing real structure from noisy data. I have tested it on another anion species, and the results are very good. 

I mislead myself through recalling legacy `basex.exe` behaviour for velocity-map images that had been over corrected for camera noise, resulting in truncated baseline spectra. Not ANU data btw.

MikhailRyazanov:
I should probably add to the documentation some clarification that the non-negative solution is only asymptotically unbiased, but for scarce data can lead to a noticeable bias and thus is not recommended as an intermediate result for further processing (averaging, fitting, and so on)...

stggh:
Yes, your statement is good. It may not mean much to the average `pyabel` user, but at least they have been warned.

DanHickstein:
> I should probably add to the documentation some clarification that the non-negative solution is only asymptotically unbiased, but for scarce data can lead to a noticeable bias and thus is not recommended as an intermediate result for further processing (averaging, fitting, and so on)...

That seems like a good idea. Probably also worth spelling out that: "Using reg='nonneg' typically produces *images* that exhibit less noise, but can introduce bias into the angularly integrated data."

Issue #325, opened 2021-07-20
Default basis_dir for PyAbel


MikhailRyazanov:
Continuing [this discussion](https://github.com/PyAbel/PyAbel/issues/247#issuecomment-478345872):
> Regarding where to store the basis sets: this is an issue for both the GUI and the command-line-interface usage of PyAbel. While we could find a solution that works only for the GUI, I think that this might be an opportunity to fix it globally. Currently, we are saving the basis sets in the current working directory, which litters the users computer with basis sets and most users will likely end up re-generating basis sets needlessly. Saving them in a single user-accessible directory provides a straightforward way for users of both the GUI and the CLI to manage their basis sets.

and #321...

I've looked closer at [appdirs](https://github.com/ActiveState/appdirs) and OS-specific recommendations and thought that we can do something like this:
* Add several new functions to `abel.transform`:
   * `set_basis_dir(...)` to set the default directory for basis files.
   * `get_basis_dir()` to get the currently set default directory for basis files.
   * `default_basis_dir()` to get the system-dependent cache directory suitable for storing basis sets.
* `get_basis_dir()` will be also used internally by all transform methods if no explicit `basis_dir` parameter is given to them, so the user can set the preferred directory (using `set_basis_dir(...)`) once in the program and then do any transforms without worrying about this.
* If `set_basis_dir(...)` was not called explicitly, the value of `default_basis_dir()` will be used. (Later, if we add some global configuration, this could be also altered there.)
* Add `basis_dir_cleanup(...)` utility functions to all transform methods for removing corresponding files (like `cache_cleanup()` currently works for RAM) and `abel.transform.basis_dir_cleanup(..., method='all')` for removing basis files for all or some methods.

This way we can keep all the basis files in one (and appropriate) place, and they can be easily removed when needed.
There are apparently no hooks in Python's distutils/setuptools/whatever that can be used for automatic removal of caches upon module deinstallation, but we can add to README a recommendation to run
```python
import abel
abel.transform.basis_dir_cleanup(method='all')
```
before uninstalling.

Here is a draft: [Default basis_dir for PyAbel](https://gist.github.com/MikhailRyazanov/beaa5d150a8f26c4159964c66b55c740).
Partially tested, but without “cleanup”.

Let's discuss, and if this look good, I'll prepare a PR...

DanHickstein:
Great work Mikhail! I think that this is the right approach. 

[...]

Issue #323, opened 2021-07-06
Enable MacOS and Windows testing on Travis-CI


DanHickstein:
Travis now supports MacOS testing. Seems like a no-brainer that we should enable this and run at least one test on MacOS. Some significant fraction of our users (myself generally included) are on MacOS. 

Currently we are using Appveyor for Windows testing. This works fine, but it's another thing to maintain. We could keep the Windows testing here for now or try switching to Travis. I don't have strong feelings on this one.

MikhailRyazanov:
I guess, adding MacOS testing to Travis definitely should be at least tried.

But I don't really understand their billing model. In [on place](https://docs.travis-ci.com/user/billing-faq/#what-if-i-am-building-open-source) they say:
> Each of the Travis CI Plans contains an amount of special OSS credits <u>per month</u> assigned to run builds only on public repositories.

But in [another](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing#building-on-a-public-repositories-only):
> When your credit allotment runs out - we’d love for you to consider which of our plans will meet your needs.

and
> (should your run out of credits again you can repeat the process to request more or discuss a renewable amount)

So will you need to request renewals every time, or they allow us 25000 credits (=500 macOS minutes) _per month_?

DanHickstein:
I think that smaller open-source projects just need to ask for one-off grants of credits. I don't think that our credits refresh every month. We currently have about 30k credits. Each build on travis takes about 90 seconds, so if we have 3 linux and 1 macOS build, then this is 10 * (4.5 mins) + 50 * (1.5 mins) = 120 credits. This is only 600 or so builds. So, we'll end up asking for more credits every year or two at our current rate. Probably if we pick up the pace we can ask for credits in larger chunks. 

Numpy and Scipy also got hit with the TravisCI paywall earlier this year:
https://github.com/scipy/scipy/issues/13109
https://github.com/numpy/numpy/issues/17824

Scipy seems to have dropped travis completely in favor of some GitHub built-in CI.

In our experience, Travis has been excellent, and I'd suggest that we keep using them unless the credit-requesting process becomes too onerous.

MikhailRyazanov:
Sure, let's just keep it as is (with macOS enabled) for now and see how it goes. Maybe they'll change something again in the meantime... :–)

MikhailRyazanov:
[...]

PR #327, opened 2021-07-29
Default basis_dir

MikhailRyazanov:
Implementation of a reasonable default directory (instead of `'.'`) for storing basis sets (see #325)...

Seems to work and pass tests at least in Linux and Windows, hope to see the same in macOS after @danhickstein enables it (#323).

(This PR can be merged before or after #326, but after merging one, the other needs to be adjusted accordingly.)


DanHickstein:
This seems like the right move. Great work Mikhail!  I'll try to get MacOS testing working and review these changes next week.

README.rst
```diff
@@ -71,6 +71,17 @@ Or, if you wish to edit the PyAbel source code without re-installing each time :
...
     python setup.py develop
 
+Before uninstalling
+~~~~~~~~~~~~~~~~~~~
+
+Some transform methods can save generated basis sets to disk. If you want to uninstall PyAbel completely, these files need to be removed as well. To do so, please *first* run the following script::
+
+    import abel
+    import shutil
+    shutil.rmtree(abel.transform.get_basis_dir())
```
stggh, line 81: neat!

stggh (approved):
A lot of work for a small directory change. Good stuff. 

Works well under Linux (Ubuntu 21.04), and Mac OSX 11.5.2.

DanHickstein (approved):
Awesome work! This is a long-needed feature for PyAbel. This has the potential to massively speed up the workflow of certain users. The only issue I see here is if we have outdated basis sets floating that cause compatibility issues when we update the software. But, we had that problem anyways, so there's no reason not to get this implemented now. I'd say go ahead and merge whenever you like!

PR #328, opened 2021-08-16
Enabling OSX testing on travis

DanHickstein:
Following https://github.com/PyAbel/PyAbel/issues/323, we should enable testing on MacOS is possible. TravisCI seems to support this. 

I think that we just need to edit the `.travis.yml` file. This is my first attempt. Will probably take a few iterations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ jobs:
       python: 2.7
       env: TOXENV=py27, DEPS="numpy scipy cython"
     - os: osx
-      language: generic
+      python: 3.9
       env: TOXENV=py39, DEPS="numpy scipy cython"
     - os: windows
       python: 3.9
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,8 @@ jobs:
       python: 2.7
       env: TOXENV=py27, DEPS="numpy scipy cython"
     - os: osx
-      python: 3.9
+      language: generic
+      osx_image: xcode12.5
       env: TOXENV=py39, DEPS="numpy scipy cython"
     # - os: windows
     #   language: generic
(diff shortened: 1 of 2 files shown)


DanHickstein:
I tried enabling Windows testing on Travis as well, but it's labeled as experimental, and it took longer than 10 minutes for the build to _start_, so I gave up. I think that we should have at least one build run as quickly as possible so that people can spot major errors quickly. So, we can just continue to use AppVeyor for our Windows build needs.

MikhailRyazanov:
Great, nice to see that macOS testing works (and passes)!
As I understand, this is using the default preinstalled Python 2.7. Will it make sense to try testing with the latest Python 3 also?

DanHickstein:
I was trying to test on python 3.9 on MacOS. Let me see if I can get that to happen.

DanHickstein:
Huh, I can't figure out how to set the python version on MacOS on Travis.

MikhailRyazanov:
Well, [here](https://docs.travis-ci.com/user/languages/python/) they show a red warning that “Python builds are not available on the macOS and Windows environments.” :–) So maybe it's not supposed to work at all. On the other hand, it [could](https://stackoverflow.com/questions/56608429) be possible to install [using Homebrew](https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-macos), although I'm not sure this is really worth investigating (but if yes, perhaps all other builds should better be temporary disabled to avoid overusing them)...

MikhailRyazanov:
@DanHickstein, maybe we should merge this as is (but squashed) for now? Testing with the built-in Python is still useful and better than nothing, especially before continuing with #327...

MikhailRyazanov:
Well, I'm going to merge this to see how it works...

DanHickstein:
Awesome! Thanks for moving this forward.

Issue #332, opened 2022-01-20
Releasing v0.8.5 soon


DanHickstein:
We had a request from the [SCICO](https://github.com/lanl/scico) project to release a new version of PyAbel with the latest features. This seems like a great time for a release anyway, since we don't have any outstanding pull-requests, and we've incorporated lots of new goodies since v0.8.4 was released. 

Any objections if @MikhailRyazanov and I release v0.8.5?

stggh:
None from me. Go for it.

MikhailRyazanov:
OK:
- [x] [RDT](https://pyabel.readthedocs.io/en/v0.8.5/changelog_link.html)
- [x] [PyPI](https://pypi.org/project/PyAbel/0.8.5/)
- [x] [Zenodo](https://zenodo.org/record/5888391)
- [ ] Conda?

DanHickstein:
For Travis, I think that only one of the builds successfully uploads to PyPi, and the rest fail because there is already a pyabel v0.8.5 on PyPi. (Or, at least that's what I think is going on. The error message is unhelpful, but I recall seeing it before.) We are indeed on PyPi: https://pypi.org/project/PyAbel/

![image](https://user-images.githubusercontent.com/1107796/150560601-637eed72-15f2-4b64-84ca-47dd86284df9.png)

I honestly have no idea how the conda-forge thing works. I thought that maybe it automatically took the recent version...

DanHickstein:
@rth: should the conda-forge repo automatically update after we make a release on PyPi?

rth:
Yes, a PR will be made by a bot in https://github.com/conda-forge/pyabel-feedstock but you would need to check and merge it.

DanHickstein:
Awesome, thanks!

I guess we should add this instruction to our release checklist in the contributing.rst.

MikhailRyazanov:
@rth, the [conda PR](https://github.com/conda-forge/pyabel-feedstock/pull/20) has been indeed created and passes all tests! Is it possible to see the resulting files? I can see in the logs that the Cython library is now compiled with the new options, but it would be interesting to examine the actual packages before releasing them.

rth:
You should see the log of `pip install` in the CI logs. It's also possible to add extra arguments to that command in `meta.yaml` for more verbosity.

MikhailRyazanov:
I would like to see the conda *packages* built during the test. Or they are not saved at this stage?

rth:
[...]

PR #333, opened 2022-01-20
Preparations for v0.8.5 release

MikhailRyazanov:
Here are the changes needed to release a new version (see #332).

Draft of the release notes:

```
Some improvements to efficiency and consistency of existing tools, as well as
the compatibility of conda packages. The default path for saving and loading
basis sets for methods that use them has changed from “current directory” to a
single cache location, with possible customization. Please see the details
below and take appropriate actions if necessary.

This PyAbel version also incorporates another fast general-purpose method, from
Daun et al. (2006), and implements several original extensions to it.

Here is a full list of changes in v0.8.5:

* New functions in tools.vmi for angular integration and averaging, replacing
  angular_integration() and average_radial_intensity(), which had incorrect or
  nonintuitive behavior (PR #318, PR #319).
* Avoid unnecessary calculations in transform.Transform() for the
  symmetry_axis=(0, 1) case (PR #324).
* New method by Daun et al. and its extensions (PR #326).
* Basis sets are now by default stored in a single system-specific directory,
  see get_basis_dir() and set_basis_dir() in abel.transform (PR #327).
  **Important!** The current working directory is no longer used by default for
  loading basis sets. It is recommended to execute

    import abel; print(abel.transform.get_basis_dir(make=True))

  and move all existing basis sets to the reported directory.
* Cython optimization flags are changed to make conda packages compatible with
  all CPUs and to improve the direct_C method performance (PR #331). Bitwise
  reproducibility of direct_C transforms might be affected.
```


DanHickstein:
These changes all look good to me! And I like your release notes. 

Unfortunately, the read-the-docs build is failing. And we need to get it working before we make a release, since we definitely want this release to have docs.

Obviously this has nothing to do with your changes, and is likely associated with some change with how readthedocs works. I think that perhaps we need to include numpy in our MOCK_MODULES in conf.py. Basically, can you try uncommenting line 42 here and commenting out the next line: https://github.com/PyAbel/PyAbel/blob/master/doc/conf.py#L42

MikhailRyazanov:
Hm, now readthedocs suddenly broke and “can't find NumPy” anymore...
Comparing the [last successful](https://readthedocs.org/api/v2/build/15612512.txt) log with the [current failing](https://readthedocs.org/api/v2/build/15831616.txt) one shows that the option `--system-site-packages` is now missing, which is likely the cause the of problem. Some old TRD [issue](https://github.com/readthedocs/readthedocs.org/issues/1639#issuecomment-139237619) suggests to:

> go to the *Advanced settings* section in your project's admin panel and select the checkbox in *Use system packages*.

I have no idea why we've stumbled on this only today, but where at least is that “admin panel”? @Danhickstein?

DanHickstein:
Oh, I think that box is checked. Do you have an account on RTD? If you make one, then I think that I can add you as a maintainer to the RTD account for PyAbel. (Also applies to any of the other developers who want access, @stggh)

DanHickstein:
image

DanHickstein:
Oh, I think that I see the problem. I think that the message in the gray box at the top of that image is new. So all of those settings are now ignored since we have a readthedocs.yml file. We need to figure out how to set the site-packages to true in the YML file.

DanHickstein:
https://docs.readthedocs.io/en/stable/config-file/v2.html#python

DanHickstein:
So, I think that we need to add this line to readthedocs.yml:
```
python:
  system_packages: true
```

MikhailRyazanov:
Thanks, Dan. I've just registered on RTD through my GitHub account, and apparently couldn't see the admin page because it doesn't recognize me as a maintainer.

>  I think that we need to add this line to readthedocs.yml

Will this work without `install:` and everything else?

DanHickstein:
I'm guessing that you're correct that the "install:" code is also required.

DanHickstein:
Oh, and I just added you as a maintainer on RTD, so you should be able to change the admin settings now.

MikhailRyazanov:
Thanks! There are really no more settings (I though that maybe we could enable `fail_on_warning` somewhere in the settings and get rid of the YML file). I'll try to add just `system_packages` first, and if this doesn't help, will study how to write a complete config...

MikhailRyazanov:
OK, this indeed helped. Perhaps at some point we should create a proper complete `.readthedocs.yml` config, but probably not right now. Maybe wait until they break something again?.. :–)

DanHickstein:
Yeah, this seems to be working, so I think that you are good to go. I approve this PR!

DanHickstein (approved):

stggh (approved):
All good

MikhailRyazanov:
OK, merging!..

DanHickstein:
Great! Now, you should be able to hit the "Draft new release" button here: https://github.com/PyAbel/PyAbel/releases

Issue #335, opened 2022-01-22
Test failure in v0.8.5


bwohlberg:
One of the tests is failing in release v0.8.5:
```
$ pytest --pyargs abel
============================= test session starts ==============================
platform linux -- Python 3.8.12, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
```
...
```
tests/test_dasch_methods.py ..F..                                        [ 20%]
```
...
```
=================================== FAILURES ===================================
____________________ test_dasch_deconvolution_array_sources ____________________

    def test_dasch_deconvolution_array_sources():
        im = abel.tools.analytical.SampleImage(101).image
        q = abel.tools.symmetry.get_image_quadrants(im)[0]
    
        # clean up any old deconvolution array files
        fn = 'three_point_basis_{}.npy'.format(q.shape[0])
        if os.path.exists(fn):
            os.system('rm {}'.format(fn))
    
        gb = abel.dasch.three_point_transform(q)
>       np.testing.assert_equal(abel.dasch._source, 'generated')
E       AssertionError: 
E       Items are not equal:
E        ACTUAL: 'file'
E        DESIRED: 'generated'
```

MikhailRyazanov:
Thanks for reporting!

This is related to #327, which changed the default `basis_dir` from `'.'` to another location, but this particular test still tries to delete its files from the current working directory. When working on that change, I forgot to rerun *all* unit tests locally, and our CI tests pass because they run on a clean system...

I need to go through the unit-test files and make sure that they all use the `tests/data` subdirectory and don't interfere with the system cache...

However, this shouldn't affect the operation of PyAbel itself.

MikhailRyazanov:
PR #336 should solve this. @bwohlberg, could you please test it?

bwohlberg:
All tests pass in the branch for PR #336.

MikhailRyazanov:
Thanks!

MikhailRyazanov:
OK, I'm closing this as solved.

PR #336, opened 2022-01-23
Repairs to basis_dir (amending #327)

MikhailRyazanov:
Some changes related to `basis_dir`, overlooked in the previous PR #327:
* The `dasch` unit tests were failing unless ran on a clean system (the disk-cache test tired to delete a temporary file using a wrong path; reported in #335). Now they avoid using disk caching unless absolutely necessary, and when needed, use the local `tests/data` subdirectory for these temporary files, like all other tests do, and also clean-up after themselves.
* There was a strange error in `basex` when using a relative path for `basis_dir` and running Python 2 (in this case `glob()` yields relative paths). I'm surprised how it was working and passing all tests...

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+* Correct behavior of relative basis_dir in basex under Python 2 (PR #336).
+
 v0.8.5 (2022-01-21)
 -------------------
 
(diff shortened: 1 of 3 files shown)


DanHickstein (approved):
Thanks for taking care of this!

stggh (approved):

MikhailRyazanov:
OK, merging...

PR #339, opened 2022-02-12
Spherical polynomials and improved SampleImage

MikhailRyazanov:
Here is an addition to `abel.tools.polynomial`, now implementing the analytical Abel transform of polynomials in spherical coordinates (bivariate polynomials in *r* and cos *θ*). A special case of them was already used for rBasex basis functions, but here they are generalized to arbitrary *r* powers and arbitrary radial limits.

This can be useful if somebody wants to model smooth VMI-like distributions, for example, by approximating them with cubic splines. But the main goal, also implemented here, was to provide high-accuracy Abel transforms of our sample images (`abel.tools.analytical.SampleImage`). The existing images are composed of Gaussian peaks, which don't have analytical Abel transforms, but they can be approximated by piecewise polynomials with any desired accuracy, and thus an arbitrarily accurate Abel transform can be obtained.

I've also added, for completeness, an isotropic Gaussian “blob” image (2D analog of `abel.tools.analytical.GaussianAnalytical`), which does have an analytical Abel transform. And an image resembling the “classic” calibration image for photofragment VMI (O<sup>+</sup> from O<sub>2</sub> 4-photon photodissociation/ionization), familiar from the Eppink–Parker work. It is composed of polynomials peaks, with maybe less experimentally looking shapes, but having an exact Abel transform.

Partial backward-incompatible changes to `SampleImage` (discussed in #337):
* `.image` attribute is deprecated in favor of `.func` (still works but prints a warning).
* `sigma` parameter is changed to absolute value in pixels instead of something strange. Default values reproduce the old behavior.
* Radial scaling is corrected: `.r_max` is now in pixels according to `n`, and all peaks radii scale proportionally to it (previously it was also doing something strange; see 8fb9b96e for old–new comparison, 39e39636 for the change). `'dribinski'` at its default `n=361` is not affected, `'Ominus'` at its intended `n=501` is also not affected but is slightly changed at other `n` values (I had to adjust the PES unit test slightly to account for this).

Issue #289, opened 2020-04-26
Make the methods documentation work better


DanHickstein:
Following up on #286, I tried looking at our documentation while putting myself in the shoes of a new user. Mainly, I was concerned with this: how would I find information about each transform method?

So, I would start by looking at the docs for [abel.Transform](https://pyabel.readthedocs.io/en/latest/abel.html#module-abel.transform). Here, we have very short descriptions for the [`method` argument](https://github.com/PyAbel/PyAbel/blob/master/abel/transform.py#L49). These summaries generally just say the authors of the method, which is not particularly helpful. If we only have one sentence to describe the methods, shouldn't we give some most useful information about the usage? Also (and I know this will be controversial), should we give some idea about which methods are most recommended? Currently we just list them in alphabetical order. Most importantly, we need to include links for the users to get more information.

Next, if the user happens to scroll down in the `abel.Transform` docs (not guarenteed), then they will find this [Notes section](https://github.com/PyAbel/PyAbel/blob/master/abel/transform.py#L198) which provides some additional information about the transform methods. This Notes section seems like a bit of a duplication (in intent, not in content) of what we are doing in our [Transform Methods doc](https://pyabel.readthedocs.io/en/latest/transform_methods.html).

We also have a dedicated .rst file for each transform method. Finally, we have the docstring for the individual transform method.

So, for basex, I can find information in many places:

1. abel.Transform[ `method` docstring](https://github.com/PyAbel/PyAbel/blob/master/abel/transform.py#L53). 
2. abel.Transform [Notes section of docstring](https://github.com/PyAbel/PyAbel/blob/master/abel/transform.py#L213). 
3. [Transform Methods doc.](https://pyabel.readthedocs.io/en/latest/transform_methods.html)
4. [basex.rst. ](https://pyabel.readthedocs.io/en/latest/transform_methods/basex.html)
5. [basex computational details](https://pyabel.readthedocs.io/en/latest/transform_methods/basex-comp.html#basexcomp)
6. [abel.basex.basex_transform docstring](https://pyabel.readthedocs.io/en/latest/abel.html#module-abel.basex)

And few of these are linked together with hyperlinks.

We should think more carefully about how we structure the docs to make information easier for users to find. I think that I would suggest the following:

[...]

PR #343, opened 2022-03-16
PDF documentation and updates

MikhailRyazanov:
An attempt to enable downloadable PDF documentation on Read the Docs (even if this doesn't work, at least building it locally would become possible). See #340 for some history.

On my system it now builds without errors and produces usable results, although there is a general problem with pagination: since most images are inserted into the text flow, they cannot be floated without creating a mess, thus when a relatively tall image doesn't fit to the rest of the page, a page break is forced, creating a big gap. And since stupid Sphinx doesn't allow overriding LaTeX float options for individual figures (without resorting to black magic), even those large figures with captions, which should have been floated, are contribution to the problem.

There is also one noticed issue that multiple references to the same footnote *sometimes* don't work. It's a mystery why, because some of them do work, and I don't know what to do about this. However, this seems to be a minor nuisance.

I've also made many minor improvements in formatting and linking (particularly, between method invocations and descriptions, to further address #289) and updated some URLs (with the help from ``make linkcheck``!).

A shameful overlook that correcting the ``SampleImage`` scaling in PR #339 broke the rBasex regularization examples is also addressed here.

P.S. Unfortunately, almost every file is affected, so maybe considering each commit successively instead of looking at the overall change would be easier, or maybe not... Please review carefully!


MikhailRyazanov:
Hey, where are all the “checks”?!

stggh (approved):
@MikhailRyazanov your attention to detail is mind blowing. Excellent work as usual.

`livehtml`, `latexpdf` work on my Ubuntu 21.10 system, once `latexmk` was installed. 

`linkcheck` fails with 
```
(transform_methods/two_point: line   43) broken    https://github.com/PyAbel/PyAbel/pull/155#issuecomment-200630188 - Anchor 'issuecomment-200630188' not found
```

MikhailRyazanov:
> `livehtml`, `latexpdf` work on my Ubuntu 21.10 system, once `latexmk` was installed.

OK, great! RTD in fact provides no control over the build commands, so they either must have `latexmk` available (if they use default Sphinx tools) or implement some equivalent (maybe just run `pdflatex` [3 times, as usual](https://tex.stackexchange.com/questions/53235/why-does-latex-bibtex-need-three-passes-to-clear-up-all-warnings)...).

> `linkcheck` fails with

Because GitHub [has broken](https://github.com/github/feedback/discussions/7370#discussioncomment-2374779) links to comments in long discussions... I don't think that we can do anything until/unless they repair this somehow.

DanHickstein (approved):
The PDF builds great on my system, and it looks FANTASTIC! Incredible work as always Mikhail!

MikhailRyazanov:
Great! Then I'm going to merge this and see whether RTD can make a downloadable PDF...

MikhailRyazanov:
OK, success!

According to [the log](https://readthedocs.org/api/v2/build/16418873.txt), LaTeX ran without any outstanding issues, and I can now see the PDF under “Downloads” at https://pyabel.readthedocs.io/en/latest/. The file looks a little bit different from my local build (due to different fonts and slightly different styles), but also without any obvious issues. HTMLs also look good.

DanHickstein:
Awesome!

Seeing all of the documentation in a single PDF is really impressive. It makes me proud of what we have build here! Good work everyone!

Issue #347, opened 2022-03-21
Offline HTML


MikhailRyazanov:
Offline PDF documentation is now available (#343), but while working on that I was also considering how to build offline HTML documentation properly. I've got rid of most dependences on external images (hosted inside our GitHub issues), but two “online” images  still remain: the “cover” image (3D Object ⇄ 2D Projection), used in README and in “Comparison”, and the output of “Example of use” from README.

The cover image is, frankly, not essential and takes >2 MB (~1/4 of the whole PDF file, by the way!), although it looks appealing. So I've made an attempt to redraw it in SVG (~10 kB and also convertible to ~30 kB PDF for LaTeX):
![cover](https://user-images.githubusercontent.com/12854133/159374600-e0664792-c4c3-4f1a-995e-79bcc75c63dc.svg)
(This is also a test that GitHub now supports SVG images!)
I guess, due to its small size, it can now be included in the repository. I don't know, however, what is the proper way to use it simultaneously in
* README at GitHub
* README in the built documentation
* README at PyPI

Searching the internet shows that this is *not straightforward*. Does anybody know how to do so?

And the image in README example perhaps needs to be updated (it has an old Matplotlib appearance). I guess, we can easily include a script to generate it for offline documentation. But should we then just refresh the online image for GitHub/PyPI or maybe we can instead link to the generated image at RTD?

stggh:
Nice! What software did you use to generate the figure? 

Sorry, no idea(s) on the "proper way", any `LaTeX` processing appears to require .svg conversion.

MikhailRyazanov:
Oy vey! According to “Images” and “Relative links” [here](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax), GitHub can use images from the repository, and this [works](https://github.com/MikhailRyazanov/PyAbel/tree/offline#readme) indeed. However, PyPI doesn't understand this (see a [recent report](https://www.reddit.com/r/learnpython/comments/psoqy7/image_doesnt_render_in_readmemd_on_testpypiorg/) ~6 months ago) and apparently is not going to in any foreseeable future. I guess, the only viable solution would be to do it in two steps: first, merge the image into master, and then change its path to the corresponding `github...raw` URL. This has [obvious drawbacks](https://github.com/StochSS/GillesPy2/issues/213#issue-488308286), but hopefully after that we'll not need to modify this image often or at all...

[...]

PR #348, opened 2022-03-23
Offline HTML (1st iteration)

MikhailRyazanov:
Another attempt to improve offline documentation, now by building HTML without external dependences (see #347).

The README images were redrawn/recreated, and building downloadable zipped HTML on RTD is enabled. Now this should work on RTD, and the README should look fine on GitHub. However, since PyPI doesn't support relative paths for images, after this PR is merged into master, another PR will be needed to change these paths to their absolute URLs.

Two other experimental changes:
* Sphinx relies on MathJax to render TeX  markup, and the problem is that by default MathJax scripts are loaded from jsDelivr CDN, so they *will not work offline*. Thus the docs [recommend](https://www.sphinx-doc.org/en/master/usage/extensions/math.html#confval-mathjax_path) downloading and using a local copy. Here I'm trying to do so automatically (the full MathJax script is ~2 MB, so just including it in our repository wouldn't be great, regardless of its license).
* We are using “latest DOI” from Zenodo everywhere, but in several places it would be more appropriate to use DOI for specific versions, especially since RTD keeps the docs for several previous releases. I couldn't find any clues how to retrieve such DOI from Zenodo by simple means, so here is also another script that tries to do so through their API.

Let's see how it all works...


MikhailRyazanov:
OK, all pieces seem to work so far (except that RTD doesn't build for PRs not only PDF but even zipped HTML).

stggh (approved):
Clever work, beyond my area of expertise.

PR #349, opened 2022-04-02
Offline HTML (2nd iteration)

MikhailRyazanov:
OK, offline HTML at RTD seems to be working (although not without surprises, see https://github.com/readthedocs/readthedocs.org/issues/9068).

This is the second part of #348, replacing README image links with absolute URLs such that they can be used on PyPI (hopefully).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,7 @@ Introduction
 
 .. begin-github-only2
 
-.. image:: doc/overview.svg
+.. image:: https://raw.githubusercontent.com/PyAbel/PyAbel/master/doc/overview.svg
 
 .. end-github-only2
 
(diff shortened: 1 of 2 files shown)


MikhailRyazanov:
Currently the MacOS test on Travis fails completely with the only [message](https://app.travis-ci.com/github/PyAbel/PyAbel/jobs/565666516)
```
An error occurred while generating the build script.
```
This seem to be their problem (other people [had this before](https://travis-ci.community/t/macos-build-an-error-occurred-while-generating-the-build-script/7684)). Restarting it right now gives the same result, so I'll try later...

DanHickstein:
This looks good! (For some reason, the "approve" button isn't appearing today, but I do approve this PR.)

I somewhat prefer my fancier looking 3D rendered image, but I agree that it's nice to keep the repo small and enable the docs to build offline. So, good to transition.

MikhailRyazanov:
Thanks, Dan!

> I somewhat prefer my fancier looking 3D rendered image

It is proudly featured on the RSI cover and indeed looks very nice. :–) I was initially planning to trace it exactly, but then decided to start from a simplified version. Do you have any specific suggestions how to improve it? It shouldn't be difficult to change the geometry, but reproducing the real shading with SVG gradients might be tricky (partially because the common open-source SVG → PDF conversion tools have problems with transformed gradients).

DanHickstein:
> Do you have any specific suggestions how to improve it? It shouldn't be difficult to change the geometry, but reproducing the real shading with SVG gradients might be tricky (partially because the common open-source SVG → PDF conversion tools have problems with transformed gradients).

Yeah, I think that with a few improvements, I will like it similarly to the original. I started #350 to provide a few suggestions. I think that most of them should be reasonably straightforward to implement.

I'd say this PR is uncontroversial and can be merged.

Issue #356, opened 2022-08-11
Strange things with LinBASEX


MikhailRyazanov:
In continuing #355, I was investigating strange things in the `Transform` class, and some of them led me to even stranger things related to LinBASEX.

# Quadrants #
This part:
https://github.com/PyAbel/PyAbel/blob/abe87261730564d76528484cb77a71e79b0b4a2b/abel/transform.py#L580-L585
looks very strange and has an obviously wrong syntax. However, it doesn't cause errors because is _never executed_ ... because in this code:
https://github.com/PyAbel/PyAbel/blob/abe87261730564d76528484cb77a71e79b0b4a2b/abel/transform.py#L504-L505
`_symmetry_axis` is never `None` (passing `symmetry_axis=None` in fact [sets](https://github.com/PyAbel/PyAbel/blob/abe87261730564d76528484cb77a71e79b0b4a2b/abel/transform.py#L470-L475) `symmetry_axis = [None]`, notice the list).

Returning to the “speed/anisotropy”, I guess, the intent was something like
```python
            # speed and anisotropy
            self.Beta = [(Beta0[0] + Beta1[0] + Beta2[0] + Beta3[0]) / 4,
                         (Beta0[1] + Beta1[1] + Beta2[1] + Beta3[1]) / 4]
```
(`self.linbasex_angular_integration` and `self.linbasex_anisotropy_parameter` are in fact not documented — this is how I've strated investigating this). Another problem is that `self.radial` and `self.projection` are not assigned in his case, although they must be, according to the documentation (the question is, however, what to do with them: also average over quadrants or return quadrant-wise 4-tuples?).

The latter problem revealed to an even more strange thing...

# linbasex_transform_full() return even-sized image #

The output of the full-image LinBASEX transform is even-sized, contrary to everything:
```python
import numpy as np
import abel
im = np.ones((9, 9))
tr, _, _, _ = abel.linbasex.linbasex_transform_full(im)
print(im.shape)  # (9, 9)
print(tr.shape)  # (10, 10)
tr = abel.Transform(im, method='linbasex').transform
print(tr.shape)  # (10, 10)
```
You can also plot the resulting image and see that it's shifted as if 1-based indexing was implied somewhere. And the extremely strange thing is that the unit test actually asserts that returning a 22 × 22 output for a 21 × 21 input is the way to go:
[...]

PR #357, opened 2022-08-22
Linbasex changes

MikhailRyazanov:
These are the changes related to `linbasex` as discussed in #356, plus some other minor corrections (repair of `verbose` in `Transform()` and some corrections to the examples).

One thing that I'm not sure about is the docstring for `linbasex_transform()`. Previously it was shared with `linbasex_transform_full()` (see `_linbasex_parameter_docstring`), but there are in fact too many important differences for such mechanical treatment. Thus for `linbasex_transform()` I've simply documented its unique parameters and referred to the main function `linbasex_transform_full()` for the rest. But maybe every parameter must be documented for completeness? Or maybe it would be easier/better to replace them all with `**kwargs` and say that additional arguments are passed to `linbasex_transform_full()`, like we do in some other cases?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   - pip install --upgrade coveralls pytest pytest-cov
   - pip install . -v
 script:
-  - cd tmp/; pytest -v --cov=abel --pyargs abel
+  - cd /tmp; pytest -v --cov=abel --pyargs abel
 deploy:
   provider: pypi
   user: DanHickstein
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,14 +4,14 @@ Changelog
 Unreleased
 ----------
 
-* Offline HTML and PDF documentation is now available at Read the Docs and can
-  be built locally (PR #343, #348, #349).
 * Correct behavior of relative basis_dir in basex under Python 2 (PR #336).
-* Analytical Abel transform of axially symmetric piecewise polynomials in
-  spherical coordinates (PR #339).
 * Improvements in tools.analytical.SampleImage class: more consistent and
   intuitive interface, accurate Abel transform for existing images, additional
   sample images with exact Abel transform (PR #339, #352).
+* Analytical Abel transform of axially symmetric piecewise polynomials in
+  spherical coordinates (PR #339).
+* Offline HTML and PDF documentation is now available at Read the Docs and can
+  be built locally (PR #343, #348, #349).
 
 v0.8.5 (2022-01-21)
 -------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,18 @@ Unreleased
   spherical coordinates (PR #339).
 * Offline HTML and PDF documentation is now available at Read the Docs and can
   be built locally (PR #343, #348, #349).
+* **Important** changes in the **lin-BASEX** implementation (PR #357):
+
+  - It was erroneously producing even-sized images with offset origin. Now the
+    output is odd-sized and properly centered.
+  - Output images for radial_step > 1 were scaled down by the same factor. Now
+    they have the save size as the input.
+  - The sign of Beta for odd Legendre orders is reversed to be consistent with
+    PyAbel conventions (zero angle is upwards, towards smaller row indices).
+
+* Transform() with method='linbasex' now always stores the radial, Beta and
+  projection attributes, so there is no need to pass return_Beta=True
+  in transform_options (PR #357).
 
 v0.8.5 (2022-01-21)
 -------------------
--- a/abel/benchmark.py
+++ b/abel/benchmark.py
@@ -465,12 +465,10 @@ class AbelTiming(object):
         self._benchmark('bs', 'linbasex',
                         gen_basis)
 
-        # get the basis (is already cached)
-        basis = linbasex.get_bs_cached(self.h, basis_dir=None)
-        # benchmark the transform
+        # benchmark the transform (basis is already cached)
         self._benchmark('inverse', 'linbasex',
-                        linbasex._linbasex_transform_with_basis,
-                        self.whole_image, basis)
+                        linbasex.linbasex_transform_full,
+                        self.whole_image)
 
         # discard all caches
         linbasex.cache_cleanup()
--- a/abel/linbasex.py
+++ b/abel/linbasex.py
@@ -256,7 +256,7 @@ def linbasex_transform_full(IM, basis_dir=None, proj_angles=[0, np.pi/2],
         if legendre_orders[i] % 2:
             Beta[i] = -Beta[i]
 
-    radial = np.linspace(clip, cols // 2, len(Beta[0]))
+    radial = np.linspace(clip * radial_step, cols // 2, len(Beta[0]))
 
     inv_IM, Beta_convol = _Slices(radial, Beta, legendre_orders,
                                   smoothing=smoothing)
--- a/abel/linbasex.py
+++ b/abel/linbasex.py
@@ -274,14 +274,11 @@ def linbasex_transform_full(IM, basis_dir=None, proj_angles=[0, np.pi/2],
 def _beta_solve(Basis, bb, pol, rcond=0.0005):
     # set rcond to zero to switch conditioning off
 
-    # array for solutions. len(Basis[0,:])//pol is an integer.
-    Beta = np.zeros((pol, len(Basis[0])//pol))
-
     # solve equation
     Sol = np.linalg.lstsq(Basis, bb, rcond)
 
-    # arrange solutions into subarrays for each β.
-    Beta = Sol[0].reshape((pol, len(Sol[0])//pol))
+    # arrange solutions into subarrays for each β
+    Beta = Sol[0].reshape((pol, len(Sol[0]) // pol))
 
     return Beta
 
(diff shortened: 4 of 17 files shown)


stggh:
@MikhailRyazanov - excellent work on your part, thank you! 

I am testing and the results are a big improvement on the old code. 

Benchmark fails:
```python
(py310) [Gibbo:/home/stg121]$ ipython
Python 3.10.4 (main, Mar 31 2022, 08:41:55) [GCC 7.5.0]
IPython 8.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import abel

In [2]: abel.benchmark.AbelTiming(n=101, select=['linbasex'])
n = 101
  linbasex
    estimated ???---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 abel.benchmark.AbelTiming(n=101, select=['linbasex'])

File ~/.local/lib/python3.10/site-packages/PyAbel-0.8.5-py3.10-linux-x86_64.egg/abel/benchmark.py:276, in AbelTiming.__init__(self, n, select, repeat, t_min, t_max, verbose)
    274 self._vprint(' ', method)
    275 try:
--> 276     getattr(self, '_time_' + method)()
    277 except (KeyboardInterrupt, MemoryError) as e:
    278     self._vprint('\n' + repr(e) + '! Skipping the rest...')

File ~/.local/lib/python3.10/site-packages/PyAbel-0.8.5-py3.10-linux-x86_64.egg/abel/benchmark.py:347, in AbelTiming._skip.<locals>.decorator.<locals>.decorated(self)
    345     return
    346 else:  # otherwise run the benchmark
--> 347     f(self)
    348     # calculate the actual total time and report it
    349     t = (sum(self.res[k][m][-1] for k, m in res_keys) *
    350          self.repeat) / 1000  # [ms] -> [s]

File ~/.local/lib/python3.10/site-packages/PyAbel-0.8.5-py3.10-linux-x86_64.egg/abel/benchmark.py:472, in AbelTiming._time_linbasex(self)
    469 basis = linbasex.get_bs_cached(self.h, basis_dir=None)
    470 # benchmark the transform
    471 self._benchmark('inverse', 'linbasex',
--> 472                 linbasex._linbasex_transform_with_basis,
    473                 self.whole_image, basis)
    475 # discard all caches
    476 linbasex.cache_cleanup()

AttributeError: module 'abel.linbasex' has no attribute '_linbasex_transform_with_basis'
```

MikhailRyazanov:
Thanks, @stggh! I thought that after implementing memory caching in #232, the internal `_linbasex_transform_with_basis()` became unnecessary, but forgot to check that it is still used in the benchmark code... Now I've replaced it with simply `linbasex_transform_full()`, as the overhead should be negligible.

MikhailRyazanov:
I've added very basic unit tests for the benchmarks, to make sure that breaking them again will not go unnoticed. :–)

The tests coverage reporting was also behaving strangely, showing 0% and "No data was collected", supposedly because it was trying to collect it for the current `./abel` directory while using the system-installed `abel` package. Running `pytest` in a temporary directory solves this problem (maybe `cd tmp/` in `.travis.yml` tried to do so, but `tmp/` in fact means `./tmp`, which doesn't exist, so `cd` simply failed...), although the resulting file paths in the report look ugly. This apparently doesn't have a simple solution (https://github.com/pytest-dev/pytest/issues/569 is still open, without any recent activity) but, I guess, ugly paths are still better than useless "0%".

stggh:
@MikhailRyazanov - good stuff. 

The `linbasex` transform radial profile is displaced by 1 pixel. This has been a long standing problem (although I cannot find the appropriate issue). Perhaps, your new insight might expose the soure:

linbasex-test3   gnome-shell-screenshot-gdqafc
 
```python

nm = 'Dribinski'
im = abel.tools.analytical.SampleImage(n=1001, name=nm).func

# projection
imp = abel.Transform(im, direction='forward', method='rbasex').transform

l = abel.Transform(imp, method='linbasex')
r = abel.Transform(imp, method='rbasex')
r_rad, r_beta = r.distr.rIbeta()[0], r.distr.Ibeta()[0]

# plot ---------------------------------------
fig, ax = plt.subplots()
ax.plot(l.radial, l.Beta[0], label='linbasex')
ax.plot(r_rad, r_beta/r_beta.max(), label='rbasex')

ax.legend(labelspacing=0.3, fontsize='small')
ax.set_xlabel('radius (pixels)')
ax.set_title(nm)

plt.savefig('linbasex-test3.svg')
plt.show()
```

MikhailRyazanov:
It is indeed 1/2 pixel, not depending on `radial_step`, and for all radii. I don't see any potential problems in calculating the projections or solving the equations, so there is probably something in the basis. I've simplified the code a little bit, trying to understand what's going on, but still am not sure what causes this shift. The only thing that I suspect is that the basis assumes that whole pixels are covered, whereas the outer pixels (at _r<sub>k</sub>_ from the origin) must be only half-covered. So if I try instead of `np.tri(n)` [here](https://github.com/MikhailRyazanov/PyAbel/blob/6eeb256a039d6e1fac6ace07d461dcb6fee54b5d/abel/linbasex.py#L468-L469) to use `(np.tri(n) + np.tri(n, k=-1)) / 2`, this removes the shift but introduces some strange noticeable ringing when `radial_step=1` (although this works fine for sparse basis sets). Doesn't look like a good solution...

On the other hand, I've also noticed that the fact of applying smoothing, even negligible, can remove the shift when `Beta` length is even (that is, for example, working for `n=1001` but not for `n=1001+2`). Because the convolution with an even-sized array effectively shifts `Beta` values by a half-step. :–) So I've corrected this effect in the last commits, but it also suggests two possible solutions for the shift problem:
1. Shift the `radial` array by 0.5 pixel (and use it to create the image, because the image has the same problem). This is very simple but will always produce half-integer numbers in `radial`, which might cause other problems (for example, in comparisons with other methods).
2. “Shift” `Beta` values by interpolation (using `scipy.ndimage.shift()`). This will not affect the radii, pairs naturally with smoothing, and interpolation is used anyway to construct the output image, so it probably wouldn't harm `Beta`.

MikhailRyazanov:
OK, `scipy.ndimage.shift()` actually works quite well, producing the speed distribution much more similar to other methods (can be seen better for smaller images, like `n=401`, where the peaks have fewer points). The anisotropy example and “Comparison of Transform Results” in the documentation also look more consistent. So I guess, this can be deemed as at least a temporary solution and definitely an improvement.

stggh:
This is looking very good. I am still running a few more tests.

stggh:
Sorry for the delay with my follow up, some reno jobs (painting) have been taking up my time. 

I am flummoxed  with `example_all_O2.py` which for some reason yields a displaced `linbasex` profile, whereas `example_all_dribinski.py` is all good. Simplified code, `linbasex-test6.py`, does not expose any issue between a single quadrant and the whole image transform.

Top image `example_all_O2.py` expanded, illustrating displacement of `linbasex` profile, bottom image `linbasex-test6.py` `linbasex` only (code below image):
image

```python
import numpy as np
import abel

im = np.loadtxt('data/O2-ANU1024.txt.bz2')

imc = abel.tools.center.center_image(im, method='slice', square=True,
                                     radial_range=(370, 390))

# full image
l = abel.Transform(imc, method='linbasex')

# quadrant
q = abel.tools.symmetry.get_image_quadrants(imc)
lq, radial, speed, _ = abel.linbasex.linbasex_transform(q[0], return_Beta=True)
speed = speed[0]
rad, sp = abel.tools.vmi.angular_integration_3D(lq, origin=(-1, 0))

# plot 
plt.plot(radial, speed/speed.max(), label='quad.')
plt.plot(rad, sp/sp.max(), label='int.')
plt.plot(l.radial, l.Beta[0], label='full')
plt.legend()

plt.show()
```

BTW `example_all_dribinski.py` may now include `linbasex` as a method, since your changes to `linbasex` have made the result more comparable. However, `rbasex` cannot be included in this example because `rbasex_transform` returns additional parameters.

MikhailRyazanov:
I suspect that the problem is in centering. Currently the image in `example_all_O2.py` is 1024×1023, which is not what `linbasex` expects. Adding `square=True` makes it 1023×1023 and the results better:

**original**
orig

**with `square=True`**
square

Then changing `method="slice"` to **`method="convolution"`** makes it even better:
conv
and, I would say, roughly (within the noise level) consistent with other methods.

So, I think, we should at least add `square=True`.

> BTW `example_all_dribinski.py` may now include `linbasex` as a method, since your changes to `linbasex` have made the result more comparable.

Yes, I was thinking why `example_all_dribinski.py` has fewer methods (`onion_bordas` is also missing) but left it alone. :–) But I can extend it. Maybe we then should also include `example_all_O2.py` in the documentation?

> However, `rbasex` cannot be included in this example because `rbasex_transform` returns additional parameters.

It's not difficult to make a “shim” for it (using `lambda` in the `transforms` dictionary), but the figure has no room for it anyway...

stggh:
Yes, a square image is definitely required. Perhaps, `linbasex` should test that the image is square. 

The small remaining difference is now related to how well the real data image centre is defined, and the image distortion - so, all good.

Thanks, @MikhailRyazanov   

I will next review.

MikhailRyazanov:
OK, I've added `square` to centering and changed the method to “convolution” for better results.
I've also changes the normalization such that it is the same for all methods, so that they show the same signal level, and the noise levels can be compared directly. For the speed distribution, the broader peaks are correspondingly lower as well (this also makes the plot slightly less messy).

> Perhaps, `linbasex` should test that the image is square.

It does. However, the quadrant obtained from `get_image_quadrants()` was indeed square (512 × 512), although the full image wasn't (1024 × 1023).

stggh (approved):
Excellent work @MikhailRyazanov, some essential, and overall, very nice improvements.

`linbasex` may also be added into `example_all_dribinsky.py`

```
diff example_all_dribinski.py ../../PyAbel-master/examples/.
26d25
<     "basex": abel.basex.basex_transform,
29d27
<     "linbasex": abel.linbasex.linbasex_transform,
30a29
>     "basex": abel.basex.basex_transform,
```
alphabetical order for the transform methods.

MikhailRyazanov:
Thanks, @stggh!

> `linbasex` may also be added into `example_all_dribinsky.py`

I think, then it does something strange with combining the output image. So maybe simplify the whole example?
```python
# -*- coding: utf-8 -*-

# This example compares some available inverse Abel transform methods
# for the Dribinski sample image

from time import time
import matplotlib.pylab as plt
import numpy as np

import abel

fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(12, 6))

# inverse Abel transform methods -----------------------------
transforms = [
    "basex",
    "direct",
    "hansenlaw",
    "linbasex",
    "onion_peeling",
    "rbasex",
    "three_point",
    "two_point",
]
# number of transforms:
ntrans = len(transforms)

# sample image radius in pixels
R = 150

fIM = abel.tools.analytical.SampleImage(n=2 * R + 1, name="Dribinski").abel
# fIM += np.random.normal(0, 5000, fIM.shape)  # try adding some noise

print("image shape {}".format(fIM.shape))

# sectors for combining output images (clockwise from top)
row = np.arange(-R, R + 1)[:, None]
col = np.arange(-R, R + 1)
sector = np.asarray(ntrans * (1 - np.arctan2(col, row) / np.pi) / 2, dtype=int)

# apply each method --------------------

IM = np.zeros_like(fIM)  # for inverse Abel transformed images
ymax = 0  # max. speed distribution

for i, method in enumerate(transforms):
    print("\n------- {:s} inverse ...".format(method))
    t0 = time()

    # inverse Abel transform using 'method'
    recon = abel.Transform(fIM, method=method, direction="inverse",
                           symmetry_axis=(0, 1)).transform

    print("                    {:.4f} s".format(time()-t0))

    # copy sector to combined output image
    idx = sector == i
    IM[idx] = recon[idx]

    # method label for each quadrant
    annot_angle = 2 * np.pi * (0.5 + i) / ntrans
    annot_coord = (R + 0.8 * R * np.sin(annot_angle),
                   R - 0.8 * R * np.cos(annot_angle))
    ax1.annotate(method, annot_coord, color="white", ha="center")

    # polar projection and speed profile
    radial, speed = abel.tools.vmi.angular_integration_3D(recon)

    # plot speed distribution
    ax2.plot(radial, speed, label=method)

    # update limit
    ymax = max(ymax, speed.max())

plt.suptitle('Dribinski sample image')

ax1.set_title('Inverse Abel comparison')
vmax = IM[:, R+2:].max()  # ignoring pixels near center line
ax1.imshow(IM, vmin=0, vmax=0.1 * vmax)

ax2.set_title('Angular integration')
ax2.set_xlabel('Radial coordinate (pixel)')
ax2.set_xlim(0, 150)
ax2.set_ylabel('Integrated intensity')
ax2.set_ylim(-0.1 * ymax, 1.2 * ymax)
ax2.set_yticks([])
ax2.legend(ncol=2, labelspacing=0.1, frameon=False)

plt.tight_layout()
# plt.savefig('plot_example_all_dribinski.png', dpi=100)
plt.show()
```

stggh:
Yes, very clever.

MikhailRyazanov:
Done!

@DanHickstein, your turn. :–) Could you please check the affected documentation (`linbasex` module, examples, comparison) with fresh eyes?

DanHickstein (approved):
This all looks good to me! Great work @MikhailRyazanov and @stggh!

MikhailRyazanov:
Thanks! Merging...

PR #359, opened 2022-09-09
Testing: Python 3.10 and transfrom pairs

MikhailRyazanov:
I've noticed that we don't have CI tests with Python 3.10 (the current version, released almost a year ago), so this PR adds it for Travis testing (3.10 should now also be available at AppVeyor, but it's working so slowly that I'm not sure whether having more jobs there would be justified).

The testing coverage report, repaired in #357, also showed that transform pairs are not tested at all, so I've made basic tests for them. This also highlighted potential problems with the profiles that have discontinuities, so I've also added some notes to their descriptions.

And another small clean-up inspired by type-hinting exercises: `guss_gaussian()` from `tools.math` is renamed to `guess_gaussian()` — I suspect that “guss” was an accidental misprint rather than something meaningful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ sudo: false
 jobs:
   include:
     - os: linux
+      dist: jammy
+      python: 3.10
+      env: DEPS="numpy scipy cython"
+    - os: linux
       python: 3.9
       env: TOXENV=py39, DEPS="numpy scipy cython"
     - os: linux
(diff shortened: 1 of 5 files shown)


DanHickstein (approved):
Looks good to me!

stggh (approved):
All fine with me. Thanks @MikhailRyazanov!

MikhailRyazanov:
OK, merging!

Issue #354, opened 2022-05-01
UnboundLocalError: local variable 'scale' referenced before assignment


adammaj1:
Hi

I want to run ffirst example
```
python3 ./p.py
Traceback (most recent call last):
  File "/home/a/Dokumenty/mandelbrot_planes/log/python/abel/./p.py", line 2, in <module>
    original = abel.tools.analytical.SampleImage(name='Gerber').func
  File "/home/a/.local/lib/python3.10/site-packages/abel/tools/analytical.py", line 486, in __init__
    X, Y = np.meshgrid(self.r*scale, self.r*scale)
UnboundLocalError: local variable 'scale' referenced before assignment
```

What is wrong ? 

My settings
Ubuntu 22.04 LTS
```

python3 -V
Python 3.10.4

 pip install PyAbel
Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: PyAbel in ./.local/lib/python3.10/site-packages (0.8.5)
Requirement already satisfied: setuptools>=16.0 in /usr/lib/python3/dist-packages (from PyAbel) (59.6.0)
Requirement already satisfied: scipy>=0.14 in /usr/lib/python3/dist-packages (from PyAbel) (1.8.0)
Requirement already satisfied: numpy>=1.6 in /usr/lib/python3/dist-packages (from PyAbel) (1.21.5)
Requirement already satisfied: six>=1.10.0 in /usr/lib/python3/dist-packages (from PyAbel) (1.16.0)
```

TIA
Adam

MikhailRyazanov:
I guess, you are trying the example from out development version (called “[latest](https://pyabel.readthedocs.io/en/latest/)” on Read the Docs) but using the “[stable](https://pyabel.readthedocs.io/en/stable/)” PyAbel version [from PyPI](https://pypi.org/project/PyAbel/) — this is what `pip` installs. If you want to use what you see here, in GitHub master branch, you need to install it manually (using `git` or from the ZIP file, click the green button “Code” on the [home page](https://github.com/PyAbel/PyAbel)).

If it's not essential to use that particular sample image, then just use the default (`original = abel.tools.analytical.SampleImage().func`, without `name=...`), it should work.

MikhailRyazanov:
@DanHickstein, @stggh, this makes me think that we need to tweak the README further, such that instead of

> Note: This readme is best viewed as part of the [PyAbel Documentation](https://pyabel.readthedocs.io/en/latest/readme_link.html).

[...]

Issue #334, opened 2022-01-21
Update release checklist in contributing.rst


DanHickstein:
We should make a better release checklist in this section of the CONTRIBUTING.rst file: https://github.com/PyAbel/PyAbel/blob/master/CONTRIBUTING.rst#releasing-on-pypi

Something like:
**Releasing a new version**

First, make a PR that does the following:
- increment the version in `abel/_version.py`
- Update the Changelog so that all of the "unreleased" changes are now part of the version. 
- Use the changelog to write version release notes that can be included as a comment in the PR and will be used later.
- Update copyright years in `doc/conf.py`

After the PR is merged:
- Press the "Draft New Release" button on the releases page. 
- Copy and paste the release notes from the PR into the release notes. 
- Release it!
- PyAbel should be automatically released on PyPi (see PR #161). Check that the new version is on [PyPi](https://pypi.org/project/PyAbel/).
- Check that the new docs are on [Readthedocs.org.](https://pyabel.readthedocs.io/en/v0.8.5/index.html)
- Check that the new version is on Zenodo: https://zenodo.org/record/5888391#.Yerecn3MKgw
Increment the version number in abel/_version.py.
- A bot should automatically make a PR on the conda-forge repo (takes a while). This will need to be manually merged: https://github.com/conda-forge/pyabel-feedstock

https://github.com/PyAbel/PyAbel/issues/334

MikhailRyazanov:
Since all theses instructions are for maintainers only, rather than for regular contributors, maybe there is a better place to keep them? Or at least reorganize CONTRIBUTING such that all this stuff is collected at the bottom in a separate section, marked such that potentials contributors don't need to care about it...

Regarding details:
* Maybe suggest to put the expected release date in the changelog, and if the actual date differs, then update it just before merging the PR.
* "Draft New Release" also needs to create an appropriate tag. It does suggest to do so, but writing this down would be useful, with a reminder to check that it matches the new version number (copy/paste?).
* I think, it might be better to show the links explicitly and make sure that they are correct (currently yours point at v0.8.5), like:
  * ... PyPI: https://pypi.org/project/PyAbel/
  * ... Readthedocs.org: https://readthedocs.org/projects/pyabel/versions/
  * ... Zenodo: https://zenodo.org/search?page=1&size=20&q=conceptrecid:594858&all_versions&sort=-version
    (this one looks strange — is there a human-readable link for “all PyAbel versions” or “latest PyAbel version”?)
[...]

PR #360, opened 2022-09-17
Tweaks for documentation and PyPI

MikhailRyazanov:
I think, it would be a good time to release a new version soon, so here are some changes to address some of the remaining issues.

As reported in #354, linking from the stable distribution to our development documentation ("latest") is confusing. However, the description displayed on PyPI (and included in the package) is defined in `setup.py` by the `long_description` variable and can be adjusted, so I've added some simple code to replace all "latest" GitHub and RTD links in README with links to the particular version. I've checked the results as descibed in https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/, and they seem correct. Hopefully, this will work properly when released (for example, the versioned "example" picture doesn't exist yet, but it should be generated on RTD upon the next release, and PyPI should find it there).

I've also updated the release checklist in CONTRIBUTING as suggested by Dan in #334, somewhat rearranged other sections in a hopefully more logical order (not sure) and added there some links where they can be useful. I've also updated README such that on RTD is links internally to the "Contributing" section instead of a file on GitHub. And added the GitHub link to "If you prefer the development version from GitHub, download it here" because on RTD and PyPI this "here" is quite different. :–)

Finally, Travis config is simplified (we don't need `TOXENV`, right?) and updated according to https://config.travis-ci.com/explore complains and https://docs.travis-ci.com/user/deployment/pypi/ suggestions. In particular, I hope that it will stop trying to upload to PyPI multiple times and thus "fail" all jobs except one (like [here](https://github.com/PyAbel/PyAbel/runs/4898449993)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ env: DEPS="numpy scipy cython"
 jobs:
   include:
     - dist: jammy
+      python: 3.11
+    - dist: jammy
       python: 3.10
     - python: 3.9
     - python: 3.8
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,16 @@
 language: python
-sudo: false
+os: linux
+env: DEPS="numpy scipy cython"
 jobs:
   include:
-    - os: linux
-      dist: jammy
+    - dist: jammy
       python: 3.10
-      env: DEPS="numpy scipy cython"
-    - os: linux
-      python: 3.9
-      env: TOXENV=py39, DEPS="numpy scipy cython"
-    - os: linux
-      python: 3.8
-      env: TOXENV=py38, DEPS="numpy scipy cython"
-    - os: linux
-      python: 2.7
-      env: TOXENV=py27, DEPS="numpy scipy cython"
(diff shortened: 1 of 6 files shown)


DanHickstein:
@MikhailRyazanov, this looks great! Sorry for the huge delay in reviewing. 

I'd say we should merge this and release v0.9.0. Probably we should be incrementing the minor version number more often. I just didn't want to reach v0.10.0 because it looks weird. But I think that it's okay. Eventually I guess we should release a 1.0. But then we need to think about backwards compatibility. Nobody has complained about us breaking backwards compatibility all the time, so perhaps we can wait until people start complaining before v1.0 :).

MikhailRyazanov:
Thanks, @DanHickstein!
Meanwhile Python 3.11 was released and appeared at least on Travis (moreover, our conda package [got rebult](https://github.com/conda-forge/pyabel-feedstock/pull/24)), so I've updated this PR to make 3.11 officially supported.

I'll open another issue to discuss releasing a new version.

Issue #371, opened 2023-01-20
Installation via `setup.py` is depreciated


stggh:
Hi @DanHickstein and @MikhailRyazanov (a belated happy new year!) - I have been playing with my new [Orangepi5]( http://www.orangepi.org/html/hardWare/computerAndMicrocontrollers/details/Orange-Pi-5.html) and was testing Anaconda python on it. However, the installation of `PyAbel` complains about the use of `setuptools`, which appears to be depreciated:
```python
(py310) [orangepi5:.+e/Python/PyAbel]$ python setup.py install --user
================================================================================
Warning: Cython extensions will not be built as Cython is not installed!
         This means that the abel.direct C implementation will not be available.
================================================================================
running install
/usr/lib/python3/dist-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/usr/lib/python3/dist-packages/setuptools/command/easy_install.py:158: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
```
I wonder what is an alternative installation method, apart from `pip`?   I haven't yet searched for a solution - using you guys as a short-cut!

FYI PyAbel benchmarks (default OS had older versions of `setuptools`), Dell has mkl libraries installed:
```
Dell Optiplex                                          Orange Pi 5                                             Raspberry pi 4B       
i7-9700 CPU @ 3.00GHz                                  Cortex-A76/A55 @ 2.4GHz                                 Cortex-A72 @ 1.5GHz       
PyAbel benchmark run on x86_64                         PyAbel benchmark run on aarch64                         PyAbel benchmark run on       
                                                                                                                
time in milliseconds                                   time in milliseconds                                    time in milliseconds       
                                                                                                                
=== Basis generation ==============================    === Basis generation ==============================     === Basis generation ==============================
                                                                                                                
Method              n = 501    n = 1001    n = 2001    Method              n = 501    n = 1001    n = 2001     Method              n = 501    n = 1001    n = 2001
---------------------------------------------------    ---------------------------------------------------     ---------------------------------------------------
basex                 572.4      3466.8     23908.4    basex                1014.0      6031.0     40596.5     basex                4706.2     30988.6    225344.7
daun                    1.9         4.2         9.5    daun                    3.5         7.9        19.3     daun                   17.6        46.6       122.4
[...]

PR #372, opened 2023-01-23
Update installation commands

MikhailRyazanov:
To address #371, the deprecated `python setup.py` instructions are replaced with the recommended `pip install .`.

(My local Sphinx now also complains about `language = None`, so I've put the actual `'en'` there, as it suggested.)


stggh (approved):
Thanks @MikhailRyazanov for the in-depth analysis.

MikhailRyazanov:
I've also added a link to PyPI (just in case, since we didn't have it anywhere) and a small hint about how to download from GitHub (it's not so obvious for those who are not familiar with it).

stggh (approved):
Good work, thanks

MikhailRyazanov:
Thanks @stggh for noticing the problem and testing!
@DanHickstein, what do you think?

DanHickstein (approved):
Fine with me! Thanks Mikhail!

MikhailRyazanov:
OK, thanks! Then merging...

PR #373, opened 2023-08-16
Specify requirements for Read the Docs

MikhailRyazanov:
I've got an email from Read the Docs that they'll soon deprecate the “use system packages” feature, which for us means that `matplotlib` will not be available preinstalled, so we need to install it explicitly. This also applies to `numpy` and `scipy`, but these are already listed as dependencies for `pyabel` itself and hopefully will be installed as needed...

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -6,8 +6,8 @@
 version: 2
 
 python:
- # enable access to preinstalled modules (numpy, scipy, matplotlib)
- system_packages: true
+  install:
+    - requirements: doc/requirements.txt
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -65,10 +65,9 @@ Note that this requires that you have `pytest <https://docs.pytest.org/en/latest
 Documentation
 -------------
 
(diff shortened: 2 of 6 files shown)


MikhailRyazanov:
Hm, it apparently uses `abel` without _installing_ it, so we'll need to preinstall all the dependencies manually...

MikhailRyazanov:
OK, now it seems to work fine.

Some questions:

1. Maybe we should remove the following part, as it seemed to be unnecessary even before and now in fact contradicts the reality?
https://github.com/PyAbel/PyAbel/blob/f0146756b67ea02213d47f830ef39d630639794a/doc/conf.py#L30-L47
2. This also looks unnecessary, since we have no `.md` files anywhere:
https://github.com/PyAbel/PyAbel/blob/f0146756b67ea02213d47f830ef39d630639794a/doc/conf.py#L49-L51
3. Why Travis tests weren't executed for this PR? Something got broken silently without our intervention? Or they are just delayed for some reason?

stggh (approved):
@MikhailRyazanov thanks for actioning this (*). IMO it is time to dump support for python2. 

I don't know the purpose of `Mock`, but support its removal item 1, along with CommonMarkParser, item 2. 

(*) In this vein [MyST-parser](https://github.com/readthedocs/recommonmark/issues/221) appears to be a recommended replacement for `recommonmark` ;-)

MikhailRyazanov:
@stggh, thanks for noticing `recommonmark`. We actually don't use it, so I've removed its import altogether and tested that after uninstalling `recommonmark` everything still builds fine.
My local build also revealed that newer versions of Matplotlib broke one figure, so I've updated it as well.

Issue #26, opened 2015-11-21
Upload PyAbel to PyPi


rth:
At some point in the future it would be nice to [upload](https://python-packaging-user-guide.readthedocs.org/en/latest/distributing/#uploading-your-project-to-pypi) PyAbel on PyPi, so it can be simply installed with,
     
    pip install abel

DanHickstein:
Agreed. Once we are done with the major changes and have something that works reliably, we should upload to PyPi. We should be able to do this fairly soon.

DanHickstein:
Should we do this sooner rather than later in order to save the name "abel" for our project? It would be a shame to have somebody else steal the best name!

rth:
I agree, it would be a shame to start looking for a new name, again...  For now we can probably upload the last stable release. @DanHickstein  would you have time to do that? I think uploading a [Python wheel](http://pythonwheels.com/) is good practice, although someone would need to spend time looking into it and I'm not sure if that's useful in our case.  Otherwise just a plain `tar.gz` will do great.

DanHickstein:
I finally got around to trying this. I think that I was successful!  (Though it listed it as 0.7.0, and really it's more like 0.6.x, but close enough for a first try I suppose.)

https://pypi.python.org/pypi/PyAbel/0.7.0

rth:
Great, thanks for doing this! 

I can confirm that `pip install PyAbel` now works. So if you want we could indicate this in the readme as the preferred method of installation (and leave the `python setup.py develop` line for people wanting to play with the code).

We'll have to figure out what kind of description it would need as it does not seem to like Markdown in `README.md` so much, but it's a start! Version 0.7.0 is fine (it must take the version from `setup.py`), it's more a `0.7.0-alpha` but not a `0.6.x` since there were backward incompatible changes from `0.6.0`, I think  (see http://semver.org/).

DanHickstein:
Yes, I feel good having released PyAbel into the wild! :)

semver.org gives us extra flexibility when we are in version 0:
```Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.```

Release 1.0 is scary:

```Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.```

And yes, we should probably leave the version in setup.py (which gets it from ``abel/_version.py``) as, for example, ``0.7.0-alpha`` until we are ready for the true release. 

[...]

PR #376, opened 2023-10-03
Updated AMD benchmarks

MikhailRyazanov:
I've got a new computer, with AMD Ryzen 5 5600G, and ran all the speed benchmarks on it. I guess, it would be better use them to replace the old Ryzen 3 2200G results (particularly because the 2200G data lacks `direct_C` and `daun`, and they can't be added now, since the CPU in that machine has been changed to 3200G).

The plotting code is also slightly simplified, plus `pyabel-timing-extra.py` is added for “non-systematic” benchmarks (currently only for running `daun` with `reg='nonneg'` on the O<sub>2</sub><sup>−</sup> example).


MikhailRyazanov:
Unexpectedly, RTD has caused too many problems. First, an unannounced requirement to specify the build environment, then the fact that they're still pinning our “old” project to [very old package versions](https://docs.readthedocs.io/en/stable/build-default-versions.html), which at this point became incompatible with everything else... So I had to update the required Sphinx and RTD theme versions to make it working at all (this also brings them closer to local builds). Hopefully, this didn't break the resulting documentation, but please take a look at everything you can (I've looked quickly at some sections and didn't see anything obviously wrong, but I'll also try to check thoroughly).

stggh:
@MikhailRyazanov it is great that you have a new toy. Thanks for delving into all the subsequent issues, your solutions are very informative and useful.

I am running a few tests, and will get back to this discussion later.

DanHickstein:
Awesome! Our best benchmarks yet :). Now all of the "real" methods (which excludes Direct and Bordas) are solidly capable of HD video rates. I looked at the new docs and they don't seem obviously broken. I'd say we can just merge and deal with any small bugs with the docs later.

MikhailRyazanov:
Dan, thanks for checking this! In the meantime, I ran `linkcheck`, which has detected that some external links have moved/disappeared, so I've updated them correspondingly. I've also accidentally noticed some typos and corrected them. @stggh, could you please confirm that `O-ANU1024` should really be `O2-ANU1024`?

As for the “best benchmarks”, it seems that many but _not all_ timings are better that those from i7-9700, and my guess is that some are mostly limited by the RAM speed (these got better), but some depend mostly on the CPU and code optimizations, so they are still faster on i7-9700 with MKL.

abel/tools/vmi.py
```diff
@@ -441,7 +441,7 @@ def toPES(radial, intensity, energy_cal_factor, per_energy_scaling=True,
         The units affect the units of the output. e.g. inputs in
         eV/pixel\ :sup:`2`, will give output energy units in eV.  A value of
         :math:`1.148427\times 10^{-5}` cm\ :math:`^{-1}/`\ pixel\ :sup:`2`
-        applies for "examples/data/O-ANU1024.txt" (with Vrep = −98 volts).
+        applies for "examples/data/O2-ANU1024.txt" (with Vrep = −98 volts).
```
stggh, line 444: No, the discussion and  scaling factor actually apply to O<sup>-</sup> photodetachment, while, in the past, the data for `example_hansenlaw.py` was changed to `O2-ANU1024.txt`, no doubt to reduce the PyAbel package size. I will update the `abel.tools.vmi.toPES` text during today.

MikhailRyazanov: OK, now I see that it was indeed `O-` but then was replaced by `O2-` here: ff35dbf. Perhaps then all the references to it should be adjusted accordingly, or maybe `SampleImage(n=..., name='Ominus')` can be used instead?

On the other hand, we actually don't include any example data in the package, so it shouldn't be a problem to readd that file back, if it's useful at all.

One strange thing that I've just noticed is that the PyPI package does include the `examples` directory, although without `examples/data` and even without `examples/README.txt`, but the conda packages don't include any examples (which is probably expected). @DanHickstein, what do you think we should do about this? I guess, providing `pyabel-examples` as a separate package (with the data files) might be useful, but I have no idea how this can be done easily...

DanHickstein: Thanks for catching the typos!

I think that it's okay if we don't install the examples in the case of PyPi or Conda install, since most users will be interacting with the examples as part of the docs. I'm actually a bit surprised that they are installed with the PyPi install, since I think that we previously observed that they were not installed:
https://github.com/PyAbel/PyAbel/issues/26#issuecomment-170242549
But I don't have strong feelings on the issue.

What is the motivation for moving the examples to a separate repo? I think that the standard practice is to keep them in the same repository. For example, matplotlib has them in the main repo: https://github.com/matplotlib/matplotlib/tree/main/galleries

stggh: Here is a diff for `abel/tools/vmi.py` with the corrected constants for O<sub>2</sub><sup>-</sup>:

```
diff vmi.py ~/Downloads/.
422,424c422,424
<     generated energy scale with published spectra. e.g. for O\ :sub:`2`\ :sup:`−`
<     photodetachment, the origin band occurs at the
<     electron affinity :math:`EA = 3,613` cm\ :math:`^{-1}`. Values for
---
>     generated energy scale with published spectra. e.g. for O\ :sup:`−`
>     photodetachment, the strongest fine-structure transition occurs at the
>     electron affinity :math:`EA = 11\,784.676(7)` cm\ :math:`^{-1}`. Values for
443,444c443,444
<         :math:`1.204\times 10^{-5}` cm\ :math:`^{-1}/`\ pixel\ :sup:`2`
<         applies for "examples/data/O2-ANU1024.txt" (with Vrep = −2200 volts).
---
>         :math:`1.148427\times 10^{-5}` cm\ :math:`^{-1}/`\ pixel\ :sup:`2`
>         applies for "examples/data/O2-ANU1024.txt" (with Vrep = −98 volts).
458c458
<         is :math:`10^{7}/\lambda` (nm) e.g. ``1.0e7/454.5`` for
---
>         is :math:`10^{7}/\lambda` (nm) e.g. ``1.0e7/812.51`` for
466c466
<         is applied. e.g. ``-2200`` (volts), for "examples/data/O2-ANU1024.txt".
---
>         is applied. e.g. ``-98`` (volts), for "examples/data/O2-ANU1024.txt".
```

DanHickstein (approved):

abel/tools/vmi.py
MikhailRyazanov: Thanks, @stggh! I've incorporated your corrections (with minor formatting changes, to avoid line breaks between “cm” and “⁻¹”), please check that now they all are as expected.

stggh:
A few things have changed over time:
- [daz](https://github.com/chainer/daz) no longer installs, due to a missing `imp` python module.
- [MKL](https://anaconda.org/anaconda/mkl) "should" be included within Anaonda Python, but doesn't appear in `.show_config()`,

and I just had to re-run some of the timings. For some reason the i7-6700 timings have improved (not listed here). 
Any timing updates may be included later. 

The new Ryzen5 starts off faster, but then converges to the i7-9700 result. The OrangePi5 does well in the bang/$ category.

cf
cf

```
(py312)[Gibbo:.+/fig_benchmarks]$ cat working/hansenlaw.dat 
# hansenlaw       iabel(ms)           
    5             0.10410
    9             0.12103
   17             0.16169
   33             0.24174
   65             0.43798
  129             0.99764
  257             2.79352
  513             9.03836
 1025            33.32467
 2049           121.10773
 4097           510.34561
 8193          1574.43416
16385          5987.98083
32769         36042.93715

(py312)[Gibbo:.+/fig_benchmarks]$ cat Ryzen5-5600G_Linux/hansenlaw.dat 
# hansenlaw       iabel(ms)           
    5             0.09188
    9             0.10680
   17             0.14250
   33             0.21161
   65             0.37621
  129             0.75968
  257             1.80213
  513             5.06276
 1025            19.95768
 2049            76.44536
 4097           300.99862
 8193          1443.36033
16385          8271.27610
32769         31234.13253

(py312)[Gibbo:.+/fig_benchmarks]$ cat OrangePi5_armbian/hansenlaw.dat 
# hansenlaw       iabel(ms)           
    5             0.21973
    9             0.25346
   17             0.33597
   33             0.49767
   65             0.89308
  129             1.82993
  257             4.82953
  513            15.61482
 1025            59.63636
 2049           309.54884
 4097          1379.63771
 8193          6892.70024
16385         35458.90235
```

.readthedocs.yml
```diff
@@ -5,6 +5,11 @@
 # Required
 version: 2
 
+build:
```
stggh, line 8: Good to know this stuff. Thanks

doc/transform_methods/comparison/fig_benchmarks/pyabel-timing-extra.py
```diff
@@ -0,0 +1,38 @@
+import os
+from time import time
+import numpy as np
+
+import abel
+
+# The following should be used for Intel CPUs, since they work very slowly
+# with subnormal floating-point numbers (see PyAbel issue #246):
+try:
+    import daz
```
stggh, line 10: Current [daz](https://github.com/chainer/daz) does not install

stggh (approved):
Good work @MikhailRyazanov. This PyAbel section is difficult to maintain, due to the many system variables: `daz`, `MKL`, and python developments in each new version.

MikhailRyazanov:
@stggh, thanks for noticing the problem with `daz`! I was using it recently without installing (it did compile and work). Unfortunately, the original repository doesn't seem to be maintained any more, so I'll need to investigate whether there are now any reliable alternatives (maybe even in NumPy or BLAS)...

Let's merge this now, then see what else can be done...
Do you think that adding Orange Pi benchmarks would be worthy?

stggh:
> Do you think that adding Orange Pi benchmarks would be worthy?

After my raspberry pi 5 arrives, in a few weeks, we will see.

Merge for now.

PR #382, opened 2024-05-21
Updates for NumPy 2 and Matplotlib 3.9

MikhailRyazanov:
NumPy is planning to release v2.0 soon (conda already wanted to rebuild for it, see https://github.com/conda-forge/pyabel-feedstock/pull/27). I've tested locally with `numpy 2.0.0rc2` (from PyPI) and found that small modifications are needed to ensure compatibility with it:
* `numpy.trapz()` was renamed to `numpy.trapezoid()`, so we'll need to use the one that is available (`numpy.trapezoid()` isn't available in current versions, so we can't just switch to it; `scipy.integrate.trapezoid()` wasn't present in `scipy 1.2` — the last available for Python 2).
* `numpy.NaN` (used only in one documentation figure) was removed, so `numpy.nan` should be used instead.

The newest Matplotlib (v3.6.0) also complains about `plt.colorbar()` called from wrong axes (in another documentation figure), so this is also corrected here.

The rest still looks fine.

I suspect that when `numpy 2.0.0` is released, we'll also have to make a minor maintenance release...

### DIFF
--- a/abel/direct.py
+++ b/abel/direct.py
@@ -5,6 +5,10 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import numpy as np
+if hasattr(np, 'trapezoid'):  # numpy >= 2
+    trapezoid = np.trapezoid
+else:
+    trapezoid = np.trapz
 from .tools.math import gradient
 
 try:
@@ -53,7 +57,7 @@ def _construct_r_grid(n, dr=None, r=None):
 
 
 def direct_transform(fr, dr=None, r=None, direction='inverse',
-                     derivative=gradient, int_func=np.trapz,
+                     derivative=gradient, int_func=trapezoid,
                      correction=True, backend='C', **kwargs):
     """
     This algorithm performs a :doc:`direct computation
@@ -94,8 +98,8 @@ def direct_transform(fr, dr=None, r=None, direction='inverse',
         with respect to r. (only used in the inverse Abel transform).
     int_func : function
         This function is used to complete the integration. It should resemble
-        np.trapz, in that it must be callable using axis=, x=, and dx=
-        keyword arguments.
+        np.trapz/np.trapezoid, in that it must be callable using axis=, x=, and
+        dx= keyword arguments.
     correction : boolean
         If False the pixel where the weighting function has a singular value
         (where r==y) is simply skipped, causing a systematic under-estimation
@@ -155,7 +159,7 @@ def direct_transform(fr, dr=None, r=None, direction='inverse',
         return out
 
 
-def _pyabel_direct_integral(f, r, correction, int_func=np.trapz):
+def _pyabel_direct_integral(f, r, correction, int_func=trapezoid):
     """
     Calculation of the integral  used in Abel transform
     (both direct and inverse).
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -6,6 +6,10 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import numpy as np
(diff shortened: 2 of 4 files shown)


DanHickstein:
Thanks Mikhail!

I suppose one option would be to switch to [`scipy.itegrate.simpson`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.simpson.html#scipy.integrate.simpson) and avoid the trapezoidal rule issue altogether. Probably it gives a slightly better result and avoids the extra lines of code?

DanHickstein (approved):

MikhailRyazanov:
`scipy.integrate.simpson` has the same problem with Python 2 (it was called [`scipy.integrate.simps`](https://docs.scipy.org/doc/scipy-1.2.0/reference/generated/scipy.integrate.simps.html#scipy.integrate.simps) in SciPy 1.2).

But it also seems to me that the `direct` method was really designed assuming trapezoidal integration (for the [“correction”](https://github.com/PyAbel/PyAbel/blob/b10237dd24170d694765a7137c2c79b2c5c01eb9/abel/direct.py#L207-L214) and the fact that the trapezoidal rule is [hard-coded](https://github.com/PyAbel/PyAbel/blob/b10237dd24170d694765a7137c2c79b2c5c01eb9/abel/lib/direct.pyx#L64-L66) in the Cython implementation). Maybe @rth can tell more about this.

DanHickstein:
Oh, yeah, good point. We shouldn't mess with the integration. It's annoying that numpy changed the name somewhat suddenly, requiring that we include extra code to check the numpy version. But, what you have seems like the best solution.

abel/direct.py
```diff
@@ -5,6 +5,10 @@
 from __future__ import unicode_literals
 
 import numpy as np
+if hasattr(np, 'trapezoid'):  # numpy >= 2
```
stggh, line 8: I agree with this code. However, IMO support for python2 should be dropped.

abel/direct.py
```diff
@@ -5,6 +5,10 @@
...
+    trapezoid = np.trapezoid
```
stggh, line 9: The function name change is similar to what occurred for `scipy.integrate.simp` to `simpson`. Makes more sense to use the full name.

stggh (approved):
All good for me. Thanks @MikhailRyazanov 

BTW, as a separate issue, on my Ubuntu 24.04 (Dell OptiPlex 7070) `direct_C` quietly does not build. I am not sure how to force the compile.

abel/direct.py
DanHickstein: This is the *numpy* version, not python version. In this case, v2 is the upcoming version. (But otherwise I agree with you. No need to support Py2 anymore. It, and my beloved print statement without parentheses, are dead.)

stggh: Opps! I must be getting old.

>  beloved print statement without parentheses

 It tried to make a come back:
 [LWN: The (non-)return of the Python print statement](https://lwn.net/Articles/823292/)

DanHickstein: OMG! Guido has returned to bring us back our print statement :)

MikhailRyazanov: Dropping Py2 is planned (#221), but not for right now. Basically, it will allow simplifying the code (and adding type annotations), but I don't think that just declaring that we don't support Py2 anymore without actually doing any improvements is a good thing. :–)

DanHickstein: Agreed!

rth:
> Maybe rth can tell more about this.

Sorry, I have no recollection of any of this anymore.

PR #387, opened 2024-11-13
Update docs: remove misleading notation surrounding quadrant averaging

David1one:
The current documentation notation describing how the symmetrize_method=="average" function works, is misleading. It says things like "Q01 = Q0 + Q1", indicating that the combination is a summation of Q0 and Q1. Instead it is either or, if only one of the quadrants is specified by use_quadrants, or it is the average of the two quadrants is specified. This commit expands the documentation to eliminate the misleading language.

### DIFF
--- a/abel/transform.py
+++ b/abel/transform.py
@@ -199,13 +199,19 @@ class Transform(object):
         1) symmetry_axis = 0 (vertical)::
 
             Combine:  Q01:
-                        = average(Q0, Q1) if use_quadrants[0] is True and use_quadrants[1] is True
-                        = Q0 if use_quadrants[0] is True and use_quadrants[1] is False
-                        = Q1 if use_quadrants[0] is False and use_quadrants[1] is True
+                        = average(Q0, Q1) if use_quadrants[0] is True 
+                            and use_quadrants[1] is True
+                        = Q0 if use_quadrants[0] is True 
+                            and use_quadrants[1] is False
+                        = Q1 if use_quadrants[0] is False 
+                            and use_quadrants[1] is True
                       Q23: 
-                        = average(Q2, Q3) if use_quadrants[2] is True and use_quadrants[3] is True
-                        = Q2 if use_quadrants[2] is True and use_quadrants[3] is False
-                        = Q3 if use_quadrants[2] is False and use_quadrants[3] is True
+                        = average(Q2, Q3) if use_quadrants[2] is True 
+                            and use_quadrants[3] is True
+                        = Q2 if use_quadrants[2] is True 
+                            and use_quadrants[3] is False
+                        = Q3 if use_quadrants[2] is False 
+                            and use_quadrants[3] is True
             inverse image   AQ01 | AQ01
                             -----o----- (left and right sides equivalent)
                             AQ23 | AQ23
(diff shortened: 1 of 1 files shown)


MikhailRyazanov:
Looks too wordy to me...

It seems that @stggh and @DanHickstein were those who designed/implemented the logic behind this, so they probably can explain it better, but my understanding is that basically `use_quadrants` selects the input quadrants, and then that subset is used to produce all 4 output quadrants according to `symmetry_axis` by averaging among the symmetry-equivalent ones (or, if only one symmetry-equivalent was used, copying it as appropriate).

I agree that “Q0 + Q1” is misleading and should be better replaced with something like “average(Q0, Q1)”. But instead of the bunch of “if”s for each case, a general note that only those selected by `use_quadrants` are actually used in calculating the average, would be clearer.

In any case, the docstring for [`tools.symmetry.get_image_quadrants()`](https://github.com/PyAbel/PyAbel/blob/3799cc798cfd31b61b768bf7ec6dd0a7f5f9ef7b/abel/tools/symmetry.py#L13) needs to be updated in the first place.

MikhailRyazanov:
(In the meantime I'll take a look why the RTD build fails...)

DanHickstein:
Thanks @David1one and @MikhailRyazanov! I agree that the original documentation is not accurate, because we're averaging, not adding, and if only one quadrant is used, then we are ignoring the other quadrant. And while I think that your proposed change is totally correct, I wonder if we can make it more concise. Like, can we just change "Q01 = Q0 + Q1" to "Q01 = avg(Q0, Q1)", and then explain that "avg indicates the average (mean) of the two quadrants or either Q0 or Q1 if the `use_quadrants` keyword indicates that only one of the quadrants should be used".

stggh:
>  is not accurate, because we're averaging, not adding

For the purpose of the transform the symmetry related quadrant intensities are added. The role of the averaging is purely to give the correct image total intensity when the image is reassembled, it is a numerical factor. 

IMO the use of `avg(Q0, Q1)` , although code correct, hides that the quadrant intensities are being added. i.e. we are making a big deal out of a normalization factor, rather than that a simple quadrant intensity addition is occurring.

David1one:
I have updated the documentation as suggested to improve clarity and conciseness. @MikhailRyazanov @DanHickstein @stggh let me know what you think.

MikhailRyazanov:
I've merged the master branch into this PR to enable building the docs (see #389). Now the change can be seen: https://pyabel--387.org.readthedocs.build/en/387/abel.html#abel.transform.Transform (scroll down to the “Note”).

Looks good to me, except maybe remove the backticks around ``use_quadrants`` in the “cases” list (these are formatted verbatim, so backticks there don't help).

stggh (commented):
To me the text is verbose, `Qij = Qi + Qj` is simpler, IMO. However, it is not a big deal, so I am happy to go with the flow.

MikhailRyazanov:
@stggh, the problem with “``Qij = Qi + Qj``” is that in fact ``Qij = (Qi + Qj)/2`` when both quadrants are used, and ``Qij = Qi`` when only ``Qi`` is in ``use_quadrants``. In other words, “``Qij = Qi + Qj``” is always technically wrong (and, as we see, misleading to al least some users).

stggh (approved):
Fair enough, point taken. I approve the changes. Thanks @David1one

David1one:
I have updated the docs to remove backticks as recommended by @MikhailRyazanov

MikhailRyazanov (approved):

MikhailRyazanov:
I guess, @DanHickstein has no objections...

MikhailRyazanov:
Thanks, @David1one!

DanHickstein:
> I guess, @DanHickstein has no objections...

This looks great to me! Thanks @David1one and @MikhailRyazanov!

Issue #388, opened 2024-11-13
Request for clarification surrounding symmetry axis


David1one:
I wish to ask for clarification surrounding the definition of the symmetry axis for an abel transform.

For my application, I wish to define a symmetry axis that is centred on the image, and vertical. Such an example is in Issue #341. 

But it is also possible to define a symmetry axis centred on the image, but perpendicular to the plane of the image, such as in the [Hansen-Law example](https://pyabel.readthedocs.io/en/latest/example_hansenlaw.html) in the docs.

It's not clear to me why the first example and the second example have symmetry axes pointing in different directions. How do I know which way the symmetry axis points?

MikhailRyazanov:
There is a “Note that the Abel transform is always performed around the vertical axis.” in https://pyabel.readthedocs.io/en/latest/abel.html#abel.transform.Transform, but this question is still raised from time to time, so I've already prepared a README update

> **Image orientation:** The cylindrical symmetry axis in PyAbel is always **vertical**; if your data is instead symmetric around the horizontal axis, the image must be transposed or rotated by 90° before applying any Abel transform (and then back, to recover the original orientation). PyAbel adopts the "television" convention, ...

but didn't have time to submit it yet. Hopefully, will be done soon.

(And by the way, the cylindrical axis for the Abel transform must be _in the plane_ of the image; “perpendicular to the plane of the image” doesn't work.)

stggh:
Yes, the `PyAbel` transform is always with respect to the vertical axis. The symmetry axis parameter refers to (any) image (data) symmetry. This information allows some image quadrants to be combined,  improving signal to noise, before the transform is evaluated. e.g. for velocity-map imaging of photoelectrons, from a linearly polarized laser source, all the image data may be co-added and the image transformed as a single quadrant.

MikhailRyazanov:
@David1one, please take a look at the updated [“Conventions: Image orientation”](https://pyabel.readthedocs.io/en/latest/readme_link.html#conventions) section and at the ``symmetry_axis`` parameter description in [``abel.Transform``](https://pyabel.readthedocs.io/en/latest/abel.html#abel.transform.Transform). Are they sufficiently clear now?

David1one:
@MikhailRyazanov thank you for the clarifications and the README updates, I think it is much clearer now.

Issue #365, opened 2022-12-03
Re-assess CI options


DanHickstein:
I'm slightly unhappy with Travis (Open Source support slipping, weird errors (like #364)). AppVeyor seems fine but slow. So, considering other options. 

@rth, you're more tapped into the community than the rest of us – What's the current best practice for simple CI testing? GitHub actions? As you know, PyAbel just needs to run basic tests on Linux+Mac+Windows, nothing too crazy.

Link to CI with github actions: https://docs.github.com/en/actions/automating-builds-and-tests/about-continuous-integration

rth:
Hi,

yes, GitHub Actions is probably the easiest way to go currently for cross-platform builds. 
Also https://github.com/pypa/cibuildwheel to building binary wheels if necessary is fairly straightforward now in CI.

DanHickstein:
Awesome, thanks @rth! We'll probably try to transition to GitHub Actions for CI testing in the near future.

DanHickstein:
To get the basic testing working, we just need to add a file to `.github/workflows/`. 

Here is an example that runs a single Py3.9 test on Linux: https://github.com/DanHickstein/PyAbel/pull/2

MikhailRyazanov:
@DanHickstein and @stggh, now we can use GitHub Actions to build [sdist](https://github.com/PyAbel/PyAbel/actions/workflows/sdist.yml) and binary [wheels](https://github.com/PyAbel/PyAbel/actions/workflows/wheels.yml) distributions. If you have time, please take a look at them (the wheels should make `direct_C` available for all platforms).

The remaining task is to switch PyPI deployment from Travis to GitHub...

stggh:
Happy New Year @MikhailRyazanov and @DanHickstein, I am on vacation in Adelaide until 13th January, and can only test `PyAbel` on an M3 MacBookPro (36GB Sequoia 15.2, Anaconda Python 3.13.1). My issue with `direct_C` not being built was due to a missing `Cython` package, and not seeing any warning.

@MikhailRyazanov your GitHub Actions build appears to work, clever stuff!

A few `pyabel-timing.py`'s:
```
(py313) steve@M3pro MacBookProM3 % cat working/*                         
# direct_C        iabel(ms)           
    5             0.02285
    9             0.02362
   17             0.03017
   33             0.03691
   65             0.05883
  129             0.15686
  257             0.77400
  513             4.92246
 1025            33.72569
 2049           241.38475
 4097          1796.36880
[...]

PR #389, opened 2024-11-18
Cylindrical axis orientation in README; refreshed external URLs, RTD config, Python versions

MikhailRyazanov:
Let's say explicitly in README that the cylindrical axis for all Abel transforms is vertical, as this is not immediately obvious and keeps confusing some users (for example, #388, #383, #362, and several times before).

There also was a warning in a recent Read the Docs buid (for #387):
> WARNING: Calling get_html_theme_path is deprecated. If you are calling it to define html_theme_path, you are safe to remove that code.

treated as a build error (due to ``fail_on_warning: true``), so I've just removed that ``get_html_theme_path()``, as suggested.

The supported Python versions are also updated up to 3.13 (released this month) — the tests pass locally on my system, and hopefully are already supported by Travis and AppVeyor.

And, last but not least, I've updated the rotten/moved external links in the documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ env: DEPS="numpy scipy cython"
 jobs:
   include:
     - dist: jammy
-      python: 3.11
+      python: 3.13
       # give RTD enough time to build the docs before the PyPI deployment
       before_deploy: sleep 5m
       deploy:
@@ -17,10 +17,10 @@ jobs:
         on:
           tags: true
           repo: PyAbel/PyAbel
+    - python: 3.12
     # SciPy fails to install for current 3.10.8
     #- dist: jammy
     #  python: 3.10
-    - python: 3.9
     - python: 3.8
     - python: 2.7
     - os: osx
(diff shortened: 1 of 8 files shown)


MikhailRyazanov:
OK, RTD seems to work.

@DanHickstein, could you please check what happened with Travis CI? We were already unhappy with it ~2 years ago (#365), but recently it has stopped running out tests completely — if I remember correctly, #373 was the first time, and since then it never worked. If the problem has a simple solution, we should probably include it in this PR (it needs a minor update anyway, as I forgot to include “3.13” in ``setup.py``). But if not, we should really do something to address #365...

stggh (approved):
Look fine, sorry for the delay.

stggh:
> 6-week old baby

Congratulations @DanHickstein !

MikhailRyazanov:
Thanks @stggh and @DanHickstein! Travis started working again (perhaps emailing them indeed helped), but I was also experimenting with GitHub Actions (in my test repository), with encouraging results, so I'll work on migrating from Travis after ensuring that everything works as expected...

DanHickstein:
> Travis started working again (perhaps emailing them indeed helped)

Yup, they emailed me back and said that it should work now. Presumably they added more credits to our account or something. 

> Congratulations @DanHickstein !

Thanks!! :)

PR #390, opened 2024-11-23
GitHub Actions to run pytest

MikhailRyazanov:
Here is the first step to address #365 — using GitHub Actions to run the unit tests on Linux, macOS and Windows (regular Python instead of Anaconda).
* As I understand, this [should be free for public projects](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions), but even in the worst case we should be allowed 2000 minutes per month (not too much, taking into account that macOS counts at 10× rate, but probably enough).
* Python 2.7 is not tested because it's apparently not available (and we should get rid of it anyway), but it's still covered by AppVeyor.
* For Python 3, only the lowest (3.7) and highest (3.13) versions form “PyAbel requires Python 3.7–3.13” are tested. I think, this should be enough, but more can be added later if needed.
* Only 64-bit systems (amd64, plus arm64 for the new macOS) are tested. AppVeyor still covers 32-bit Anaconda for Windows.
* To be sure, installations with and without Cython extensions are tested, but only in Linux, which is probably enough.

If this work fine, the next step will be to abandon Travis CI completely and implement PyPI deployment from GitHub Actions (hopefully, with compiled wheels again, for all the platforms).

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,0 +1,68 @@
+# See https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: Tests
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+env:
+  FORCE_COLOR: 1 # colored output where possible
+
+jobs:
+  pytest:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-13, macos-latest, windows-latest]
+        python-version: ["3.7", "3.13"]
+        ext: [cython, no-cython] # with/without Cython extension
+        exclude:
+          - os: macos-latest
+            python-version: "3.7" # not available
+          - os: macos-13
+            python-version: "3.13" # to save resources
+          # skip no-cython on slower systems
+          - os: macos-13
+            ext: no-cython
+          - os: macos-latest
+            ext: no-cython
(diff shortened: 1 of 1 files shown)


MikhailRyazanov:
This will not run until merged into the master branch, but [here](https://github.com/MikhailRyazanov/Actions/actions/runs/11970991878) is an example of how it should look (using the same configuration file and a very truncated PyAbel repository).

stggh (approved):
All new to me. Thanks for working through this @MikhailRyazanov  - you have some impressive skills.

DanHickstein:
Awesome work! Let's give it a try. My reading of the rules is that it's free for public repositories, so hopefully we don't get hit with the 2000 minutes limit. As far as I'm concerned, you're good to go ahead and merge and we can see how we like it.

DanHickstein (approved):

MikhailRyazanov:
OK, let's see...

MikhailRyazanov:
I've changed “Workflow permissions” in the Actions settings for this repository to _read-only_ and _unchecked_ “Allow GitHub Actions to create and approve pull requests”. For some reason they were enabled (by default?), which doesn't look safe. My repository, where I was testing these workflows, already had them in the read-only and disallow state, so this should not break anything. And I don't think that we need any action to have write permissions or to manage PRs automatically.

DanHickstein:
Yes, I agree that these are good changed to make to the permissions. We don't need to give Github Actions too much power :)

Issue #391, opened 2024-12-03
Non-portable warnings and errors in testing


MikhailRyazanov:
I've previously seen issues with `tests/test_tools_center.py::test_set_center_float` complaining that
```
OptimizeWarning: Covariance of the parameters could not be estimated
```
on _some_ systems (like [here](https://app.travis-ci.com/github/PyAbel/PyAbel/jobs/628654155#L642)) but not others (including my local).

But now the latest macOS, in addition, gives an [**error**](https://github.com/PyAbel/PyAbel/actions/runs/12130828902/job/33821891034#step:9:189) in `tests/test_tools_polynomial.py::test_angular`:
```
          AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=0
E           -> cos^2
...
E            ACTUAL: array([-7.401487e-17,  0.000000e+00,  1.000000e+00])
E            DESIRED: array([0., 0., 1.])
```
which is very surprising because that result comes from multiplying `[1, 0, 2]` by a simple 3 × 3 matrix; for the 0th element of the result this is essentially equivalent to `(1 * 1 + 2 * (-0.5)) * 1/3` and must give exact `0` for any reasonable implementation (`-0.5` is exactly representable, and even though `1/3` is not, it multiplies what already should be exactly zero in any binary implementation).

This is apparently not an issue in NumPy/SciPy because tests in Linux and Windows with the same NumPy 2.1.3 and SciPy 1.14.1 pass perfectly.

So, @stggh (and @DanHickstein, if you'd have time), could you please run the tests on every system you can and report whether you see _any_ errors or warnings? (I'll also try to see whether I can reproduce it on GitHub with a minimal working example to investigate where it comes from.)

In principle, we can hush this strange error for now by adding `atol`, but something is really weird there.

stggh:
**Dell optiplex 7070, Ubuntu 24.10**, Anaconda python 3.13.0, No warnings. 

**Rpi5B, Rasp OS (Debian 12 bookworm),** Anacona python 3.13.0, 1 warning:

>    abel/tests/test_tools_vmi.py::test_anisotropy_parameter
  /home/stg121/Software/Python/PyAbel/abel/tools/vmi.py:397: OptimizeWarning: Covariance of the parameters could not be estimated
    popt, pcov = curve_fit(PAD, theta, intensity)

This arises from the flat data test: `check('ones',  (0, 1), theta, ones)`.

[...]

PR #394, opened 2024-12-04
Control off-range beta in `anisotropy_parameter()`; avoid broken SciPy optimizer

MikhailRyazanov:
Investigation of the spurious errors and warnings in CI tests (#391) revealed that:
1. New Apple's hardware (arm64) have larger floating-point errors.
2. The default method `'lm'` in `scipy.optimize.curve_fit()` is broken (see SciPy [#21995](https://github.com/scipy/scipy/issues/21995)) and thus was producing `OptimizeWarning` in _some_ testing environments; more importantly, it could produce wrong estimates for fitting errors.

So this PR changes the method in `curve_fit()` to a well-behaved `'trf'` and also adds tests to ensure that the error estimates returned by `anisotropy_parameter()` are correct. Some tolerances in the tests are also increased slightly to accommodate this change and the sloppier Apple's arm64 performance.

Now all the tests pass cleanly.

The `'trf'` method also supports constrained fitting, with is useful for keeping the anisotropy parameter in the physical range. Thus the `anisotropy_parameter()` function is extended here such that, instead of always reporting off-range results as `nan`, it also allows to return them as is or to use a constrained fit. By default, the old behavior (returning `nan`) is preserved.

`radial_integration()` is also extended to accept additional parameters and pass them to `anisotropy_parameter()`.

### DIFF
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -394,7 +394,8 @@ def anisotropy_parameter(theta, intensity, theta_ranges=None):
 
     # fit angular intensity distribution
     try:
-        popt, pcov = curve_fit(PAD, theta, intensity)
+        # using 'trf' because default 'lm' is broken, see SciPy Issue #21995
+        popt, pcov = curve_fit(PAD, theta, intensity, method='trf')
         beta, amplitude = popt
         error_beta, error_amplitude = np.sqrt(np.diag(pcov))
         # physical range
--- a/abel/tools/vmi.py
+++ b/abel/tools/vmi.py
@@ -338,7 +338,7 @@ def radial_integration(IM, origin=None, radial_ranges=None):
     return Beta, Amp, radial_midpt, Intensity_vs_theta, theta
 
 
-def anisotropy_parameter(theta, intensity, theta_ranges=None):
+def anisotropy_parameter(theta, intensity, theta_ranges=None, beta_out='nan'):
     r"""
     Evaluate anisotropy parameter :math:`\beta`, for :math:`I` vs
     :math:`\theta` data:
@@ -368,6 +368,21 @@ def anisotropy_parameter(theta, intensity, theta_ranges=None):
         Allows data to be excluded from fit; default (``None``) is to include
         all data.
 
+    beta_out: str
+        behavior when the evaluated anisotropy parameter would be outside the
+        physical range :math:`-1 \leqslant \beta \leqslant 2`:
+
+        ``'raw'``
+            return the results regardless of their values
+        ``'nan'`` (default)
+            return ``(nan, nan)`` in **beta** (useful for excluding such data
+            points from plots)
+        ``'bound'``
+            use constrained fitting with :math:`-2 \leqslant \beta \leqslant
+            1`; return **beta** and **amplitude** corresponding to the "best
+            fit" (useful for noisy data but could potentially hide severe
+            problems)
+
     Returns
     -------
     beta : tuple of floats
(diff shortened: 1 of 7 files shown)


MikhailRyazanov:
Suggestions for better names of the `beta_out` parameter (and values) are welcome!

And a question: should we also add `beta_out` (and maybe `theta_ranges`) to `tools.vmi.radial_integration()`, which calls `anisotropy_parameter()` internally?

DanHickstein:
Looks good to me!

`beta_out` seems like a reasonable parameter name. I suppose `beta_behavior` is another option that you can take or leave.

I don't have strong feelings about adding these parameters to `tools.vmi.radial_integration()`.

MikhailRyazanov:
> I suppose `beta_behavior` is another option that you can take or leave.

Or “behavio**u**r”. :–) Maybe then just `beta='raw'/'bound'/???` with a more meaningful name for `'nan'` (which is the default and thus can be cumbersome without causing inconvenience)?

abel/tools/vmi.py
stggh, line 1: Not your issue, but `pycodestyle` would prefer a shorter line 200.

stggh (commented):
You are very busy @MikhailRyazanov !

> also add beta_out (and maybe theta_ranges) to tools.vmi.radial_integration()

Yes, this has to be done, for consistency

abel/tools/vmi.py
MikhailRyazanov: Where? I see several complaints about other lines (none affected here), but line *200* is empty, and there's nothing long near it...

MikhailRyazanov:
I've renamed this new `beta_out` parameter in `anisotropy_parameter()` to shorter and (I hope) clearer `mode` with options `'raw'`, `'reject'` and `'bound'`.
And also, as agreed, added to `tools.vmi.radial_integration()` the parameters `theta_ranges` and this `mode` to pass directly to `anisotropy_parameter()`.
Please review the documentation for these functions.

Conda is back to life, so now really all the tests pass cleanly everywhere.

abel/tools/vmi.py
stggh: Sorry, that was `test_tools_vmi.py`, so way out of this PR:

```python
(py313)[Gibbo:.+-tmp/abel/tests]$ pycodestyle test_tools_vmi.py 
test_tools_vmi.py:200:80: E501 line too long (83 > 79 characters)
```

doc/conf.py
```diff
@@ -45,8 +45,7 @@
 templates_path = ['_templates']
 
 # The suffix(es) of source filenames.
-# You can specify multiple suffix as a list of string:
-source_suffix = ['.rst']
+#source_suffix = {'.rst': 'restructuredtext'}
```
stggh, line 48: Why the change? Is there no need to specify this parameter?  (= for my own education)

stggh (approved):
Looks fine to me. I like 'mode' for the beta return type parameter, and the clear choices.

Good work @MikhailRyazanov

doc/conf.py
MikhailRyazanov: Sphinx was issuing a warning
```
Converting `source_suffix = ['.rst']` to `source_suffix = {'.rst': 'restructuredtext'}`.
```
because the [preferred format for `source_suffix`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_suffix) has actually [changed years ago](https://www.sphinx-doc.org/en/master/changes/1.8.html#release-1-8-0-released-sep-13-2018). But it's also set to `.rst` by default, so there's no need to specify it explicitly, until we want to add something else (there are many lines in `conf.py` that show the default but are commented out). So just for consistency.

PR #395, opened 2024-12-13
GitHub actions to build sdist and wheels

MikhailRyazanov:
Here is the next step in moving from Travis CI to GitHub Actions (#365):
* Building the source distribution (sdist) and also testing that it can be installed with and without Cython.
* Building binary distributions (wheels, with the Cython extension) again, but now for
   * Linux (amd64),
   * macOS (amd64 and arm64),
   * Windows (amd64).

The wheels are built using [cibuildwheel](https://github.com/pypa/cibuildwheel) (apparently the official tool), as [suggested](https://github.com/PyAbel/PyAbel/issues/365#issuecomment-1336233250) by @rth. Although it supports more platforms, not all are actually provided (for free) by GitHub Actions, plus NumPy and SciPy themselves don't have binary packages for everything (x86 Linux is an example), so probably covering the most popular ones will be good enough.

An advantage over Travis is that these builds can be triggered manually (right now — only manually, just in case) and also for any branch, so we will be able to test them before releasing and, if needed, build and test distributions for other configurations without affecting the master branch.

No upload to PyPI is implemented yet (to avoid possible errors and because I haven't looked at that yet).

These actions will appear in the [Actions](https://github.com/PyAbel/PyAbel/actions) tab as “Build sdist” and “Build wheels”. The results will be available as downloadable workflow run artifacts.

### DIFF
--- a/.github/workflows/info.py
+++ b/.github/workflows/info.py
@@ -1,0 +1,27 @@
+"""
+Print summary information about current OS, Python, and required packages.
+"""
+import platform
+import sys
+import os
+
+print('Platorm:', platform.platform())
+print('Python:', sys.version)
+
+try:
+    import numpy
+    print('NumPy:', numpy.version.full_version)
+except ImportError:
+    print('NumPy not found')
+
+try:
+    import scipy
+    print('SciPy:', scipy.version.full_version)
+except ImportError:
+    print('SciPy not found')
+
+try:
+    import Cython
+    print('Cython:', Cython.__version__)
+except ImportError:
+    print('Cython not found')
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -19,13 +19,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04, macos-13, macos-latest, windows-latest]
-        python-version: ["3.7", "3.13"]
+        py-ver: ["3.7", "3.13"]
         ext: [cython, no-cython] # with/without Cython extension
         exclude:
           - os: macos-latest
-            python-version: "3.7" # not available
+            py-ver: "3.7" # not available
           - os: macos-13
-            python-version: "3.13" # to save resources
+            py-ver: "3.13" # to save resources
           # skip no-cython on slower systems
           - os: macos-13
             ext: no-cython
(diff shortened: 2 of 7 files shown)


MikhailRyazanov:
Like with other actions, these will become available (and thus testable) only after merging the PR. An example of how they should work can be seen in my testing repository [here](https://github.com/MikhailRyazanov/Actions/actions/runs/12307097492) (sdist) and [here](https://github.com/MikhailRyazanov/Actions/actions/runs/12307357584) (wheels).

stggh:
This may be my ignorance, but if I install a wheel, some of the PyAbel packages are missing.  Note, also that `cython_ext = True` pops up, for the wheel:

image

Compared with a git directory install (with does not create `direct_C`):
image

Is my wheel install method incorrect?

MikhailRyazanov:
That repository is only for testing how to run tests and build extensions and distributions. It does use a minimal amount of PyAbel code (and the `abel` module name) but is not supposed to do anything useful. :–) And definitely not supposed to be installed together with PyAbel (because of the name conflict).

MikhailRyazanov:
I forgot to add `recursive-include abel *.pyx` to `MANIFEST.in`, which is required because otherwise Setuptools ignore `*.pyx` files while building the source distribution if Cython is not installed (and doesn't need to be installed). Now everything should work properly, as I've tested in my PyAbel fork: [Build sdist](https://github.com/MikhailRyazanov/PyAbel/actions/runs/12324509727), [Build wheels](https://github.com/MikhailRyazanov/PyAbel/actions/runs/12324529553).

MikhailRyazanov:
And a question about included files (I loosely remember discussing this before but couldn't find where):
We include all `examples/*.py` files but not the data files (`examples/data/*`) needed to run some of the examples, neither the corresponding [`examples/README.txt`](https://github.com/PyAbel/PyAbel/blob/master/examples/README.txt). Not including the data files was a reasonable decision, since they are larger than the library itself, but shouldn't we supply that `README.txt` file and add to it an explanation where to get the data files?

DanHickstein:
Thanks again @MikhailRyazanov for more awesomeness! 

It look like we should merge this, then open a new test PR to see how well it works, right? If we're happy with how it works, then we make a new PR to remove the Travis and AppVeyor stuff, yeah?

I don't have time at the moment to dig into the details with you, but generally this looks fine and if you're happy with how it is, I think you should go ahead and merge and we can see how we like it. 

> Not including the data files was a reasonable decision, since they are larger than the library itself, but shouldn't we supply that README.txt file and add to it an explanation where to get the data files?

Yes, absolutely.

DanHickstein (approved):

MikhailRyazanov:
Thanks, @DanHickstein, and happy New Year!

> It look like we should merge this, then open a new test PR to see how well it works, right?

They are designed to be triggered manually (from the Actions tab), so they can be tested without another PR.

> If we're happy with how it works, then we make a new PR to remove the Travis and AppVeyor stuff, yeah?

We should keep AppVeyor at least for now because it runs tests in Anaconda. It seems that they can be done in GitHub Actions as well (and most probably faster), but let's defer this until dropping Python 2 support...

But Travis must be removed indeed when PyPI deployment through GitHub Actions is ready because otherwise both of them will try to upload...

>> shouldn't we supply that README.txt file and add to it an explanation where to get the data files?
>
> Yes, absolutely.

Done!

Issue #401, opened 2025-08-16
Release 0.9.1


MikhailRyazanov:
@DanHickstein, would you get some time now to work on making a new release?

The last one (v0.9.0) was in 2022 and since then has accumulated some compatibility issues with the current versions of Python/NumPy/Matplotlib, which are corrected in the `master` branch, so it would be nice to release it on PyPI, such that users can install PyAbel “normally” rather than from GitHub.

Another important thing is that soon we’ll really need to update the installation method (see #397), so now would be the best time to make the final release with Python 2 support and then be free to remove all the associated legacy stuff.

The good thing is that now we can build all the compiled packages on GitHub and thus provide them on PyPI again. It seems that there is a standard workflow for this task, but it needs to be studied/configured, which I’ll be happy to do. However, at some point I will need @DanHickstein, as the PyPI project owner, to do something there with the authorization, as currently it’s configured for Travis CI (which we’ve currently disabled and plan to abandon completely, see #365).

stggh:
I am currently travelling Kyrgyzstan, soon onward to Uzbekistan, not sure how much input I might have for this release process in any case. However, the details are of interest to me.

DanHickstein:
Well, turns out that I don't actually know my PyPi password! And apparently, my email address wasn't properly verified, so I can't access the PyAbel PyPi! Oh dear. Anyway, I stated an issue to get us back in: https://github.com/pypi/support/issues/7384

I plan to add the following as maintainers on PyPi: @MikhailRyazanov, @stggh, and @rth. I know @rth doesn't hang out here much anymore, but I'm sure he would come rescue the repo if the rest of us disappear :)

DanHickstein:
Yeah, bus factor zero! Hopefully they can help us out soon.

DanHickstein:
PyPi access regained!! @MikhailRyazanov, I added you as an owner.

DanHickstein:
Probably we should be releasing 0.10.0 since we're added lots of features since the previous release. At some point, I suppose we could release v1.0.0. However, that seems to imply that we're an organized project that doesn't break backwards compatibility all the time, and I don't want to give people the wrong impression ;)

MikhailRyazanov:
Another thought about versions. We can release this as 0.9.1, but once Python 3.14 arrives, finally drop Python 2 support and release PyAbel 0.10.0, which will be indeed sufficiently different from 0.9.x.

MikhailRyazanov:
[...]

PR #402, opened 2025-08-22
Minor documentation updates

MikhailRyazanov:
In preparation for the new release (#401), here are corrections to some problems noticed in the docs:
1. “Able-transformed” → “Abel-transformed” (a common typo, but seems to be the only occurrence in the repository).
2. “PyAbel includes the following transform methods” in README was always linking to the _latest_ RTD version, even for other RTD versions and when built locally. Now it is replaced in `doc/readme_link.rst` accordingly, like other such links.
3. The MathJax download link stopped working, so it had to be updated.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,7 +37,13 @@ PyAbel provides efficient implementations of several Abel transform algorithms, 
 Transform Methods
 -----------------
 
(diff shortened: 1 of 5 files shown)


DanHickstein (approved):

DanHickstein:
Looks good! Some good typos fixed. (At some point, we should probably have an AI go through the code and find the rest of our typos.)

PR #403, opened 2025-09-17
GitHub action for publishing to PyPI

MikhailRyazanov:
This is (almost) the last step to address #365 and enables publishing to PyPI from GitHub, using the official [PyPI publish GitHub Action](https://github.com/pypa/gh-action-pypi-publish), instead of Travis CI.

Like the previous setup with Travis, it should be triggered automatically when a new version tag is created.
However, to simplify testing, it can also be triggered manually and will publish to [TestPyPI](https://test.pypi.org/) instead of real PyPI. I've tested this from my fork (see https://test.pypi.org/project/PyAbel/0.9.1rc1/), and it seems to work.

After this PR is merged, it should be tested on TestPyPI from _this_ repository, then a similar “[Trusted Publisher](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/)” should be set up on the real PyPI, a “release candidate” made there (will need a separate PR), and then `.travis.yml` removed completely before doing the 0.9.1 release.

I've also enabled building wheels for Linux and Windows on ARM64, as they might be useful (for example, some new Microsoft Surface devices are ARM-based).


MikhailRyazanov:
Forgot to mention: TestPyPI is completely separated from PyPI, so if anybody wants to play with it, you’ll need to create an account there (the GitHub and PyPI accounts will not work) and tell me to add it to the project. Unfortunately, it has the same strict “security” as PyPI but at the same time the documentation says that “the database for TestPyPI may be periodically pruned, so it is not unusual for user accounts to be deleted”...

DanHickstein (approved):
Looks good to me! Thanks for making the transition to GitHub Actions happen @MikhailRyazanov!

PR #405, opened 2025-09-21
Preparations for v0.9.1rc3 release (v0.9.1 pre-release)

MikhailRyazanov:
Another attempt, testing whether publishing on Zenodo will succeed after @DanHickstein updated it (https://github.com/PyAbel/PyAbel/issues/401#issuecomment-3315466954).

I've also removed `.travis.yml` completely, since we don't need it anymore.

Should we mark this release on GitHub as “latest” to check that RTD is also triggered?


DanHickstein (approved):
Looks good! And yeah, try marking as "latest" and see how the release goes.

PR #408, opened 2025-10-02
End of Python 2 support

MikhailRyazanov:
This is the long-awaited removal of Python 2 support (see issue #221) in preparations for PyAbel 0.10.0 release (see [the plan](https://github.com/PyAbel/PyAbel/issues/401#issuecomment-3299743764)):
1. Removal of all ``__future__`` and ``six`` imports (plus found unused imports).
2. Removal of ``coding: utf-8`` headers, since UTF-8 is the default source encoding in Python 3.
3. Conversion form deprecated ``distutils`` to ``setuptools``.
4. Adjustments for Cython 3 (which also stops the complaints about using the legacy NumPy API).
5. Updated dependencies versions to mach those tested with Python 3.7.
6. Switching AppVeyor tests to 32 bits only, since 64-bit Windows is now covered by GitHub (where 32-bit platforms are not available).


MikhailRyazanov:
Well, AppVeyor ended up testing only with Python 3.9, since this is the last version available for 32-bit Miniconda.
They do have “regular” [Python 3.13.6 x86 (`C:\Python313`)](https://www.appveyor.com/docs/windows-images-software/#python), but switching to it will require a significant revision of `appveyor.yml`, so I might try this later (not in this PR).

Issue #409, opened 2025-10-06
Release 0.10.0


MikhailRyazanov:
Python 3.14 should be released next week. Although PyAbel 0.9.1 works fine with it (at least without free-threading), it would be nice to release PyAbel 0.10.0 with the official support of Python 3.14 (as discussed in #401) and other changes that must or could be done now:
* [ ] Drop Python 2 support and simplify the code accordingly (#221, #408).
* [ ] Remove all deprecated functions and arguments. The last of them were deprecated in v0.9.0, almost 3 years ago, so it should fine and will allow simplifying the code and documentation further.
* [ ] Convert all string formatting from `'...'.format(...)` and `'...' % ...` to f-strings (mentioned several times, mostly by @stggh). The most useful would be [self-documenting expressions](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging), greatly simplifying the code in many places. They require Python 3.8, so we’ll need to drop the support of Python 3.7 as well (which also should be fine, since even its last security release was [over 2 years ago](https://peps.python.org/pep-0537/#schedule-last-security-only-release)).
* [ ] Switch from the deprecated `setup.py` installation/distribution to PEP 517 and related standards (#397). The decision that “pip 25.0 will enforce this” has been relaxed for _some_ time, but looking at the pace at which `pip` and `setuptools` deprecate and remove various old mechanism, I suspect that we’ll need to adapt sooner than later.
* [ ] [conda-forge/pyabel-feedstock](https://github.com/conda-forge/pyabel-feedstock) will also need some updates, which I’ll describe in a dedicated issue there. Can be done after publishing on PyPI by updating the bot’s PR.

Addressing free-threading, officially supported by Python 3.14 (luckily, in a special build), can/should be done later, as this will need some work to avoid GIL in the Cython extension and considering thread safety in general.

PR #410, opened 2025-10-27
Conversion to PEP 517

### LINKED ISSUE #397: Switch to PEP 517
This is a reminder to update the installation process in light of (https://github.com/PyAbel/PyAbel/issues/391#issuecomment-2513492541)
> DEPRECATION: Legacy editable install of PyAbel==0.9.0 from file:///home/stg121/Software/Python/PyAbel (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457

Most stuff from `setup.py` must be moved to a declarative `pyproject.toml`, leaving only the minimal amount of executable code for dynamic fields and building the Cython extensions (building them still requires `setup.py`).

All this can be done properly only by dropping Python 2 support (see also #221), which is planned after releasing PyAbel 0.9.1.

MikhailRyazanov:
This should close #397 and currently is the most important change that must be done, as the latest versions of Setuptools and pip are really preventing PyAbel installation as it used to be and insist on installing and building distributions according to [PEP 517](https://peps.python.org/pep-0517/).

Basically, this PR moves all the static information from ``setup.py`` to a declarative file ``pyproject.toml``, reducing ``setup.py`` to the bare minimum required to build the Cython extension and provide consistent PyAbel version and README.

The major problem is that PEP 517 uses build isolation by default, so the opportunistic way to use Cython if it’s available doesn’t work anymore. But including it in build dependencies would trigger unconditional reinstallation of NumPy in the temporary environment and installation of Cython even for those who cannot or don’t want to use it. This problem cannot be solved by any standard means (see the comment in the ``[build-system]`` section of ``pyproject.toml``), so it is circumvented by explicitly disabling build isolation to build the extensions. I’ve added corresponding comments and warnings to the files and updates the README accordingly.

It turned out that the development of Setuptools and pip progresses too quickly, such that some features not or barely available in their latest versions for Python 3.7 are deprecated or even removed in the versions for Python ⩾3.12. So the lowest Python version supported by PyAbel had to be increased to 3.8 (which is probably fine, see #409).

Additional notes:
* For “maintainers”, I took the names (of our currently active members) from ``.zenodo.json``. Please check that they look the way you want.
* The list of “keywords” was collected by looking at the titles of the [articles citing](https://scholar.google.com/scholar?q=link:https%3A%2F%2Fpubs.aip.org%2Faip%2Frsi%2Farticle%2F90%2F6%2F065115%2F360138) our RSI publication. Any edits are very welcome (frankly, I don’t know where and how these keywords are used).
* The version is increased to ``0.10.0dev0`` (according to “[version specifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/#public-version-identifiers)”) to indicate that this is a development version and has incompatible changes.

A special request to @stggh: please test editable installs extensively! I’ve done some very basic checks but am not sure that they cover all practical cases.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,8 @@ Unreleased
 ----------
 * Python 2 support is dropped, the library code and examples are simplified for
   Python 3.x (PR #408).
+* The build system is updated to modern standards (PR #410). If you install
+  PyAbel from source, please see the updated instructions in the README.
 
 v0.9.1 (2025-09-22)
 -------------------


DanHickstein (approved):
This looks good to me! Thanks for modernizing the install and looking into all of this.

MikhailRyazanov:
Thanks, @DanHickstein!

@stggh, will you be able to look at this soon? If not, I’ll then merge this PR as is (to enable installation with the latest pip), and if any problems with editable installs are discovered later, we can work on them later...

stggh:
> will you be able to look at this soon?

Unlikely to be soon, sorry. Too many competing tasks I am afraid.

MikhailRyazanov:
Thanks, @stggh!

Have you tried an editable install? Just one system will be enough.

stggh:
> editable install?

Yes, `pip install -e .` works fine

PR #413, opened 2025-11-07
Removal of all deprecated functions and arguments

MikhailRyazanov:
Another cleanup for #409. As mentioned there, all these functions and arguments were already deprecated in v0.9.0, released almost 3 years ago, and were showing deprecation warnings since then.

The deprecation tools themselves are retained because I expect them to be needed at least for the new Direct implementation, but since at this moment there will be no examples in the code, I’ve added a comment explaining how they should be used.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,10 @@ Unreleased
   Python 3.x (PR #408).
 * The build system is updated to modern standards (PR #410). If you install
   PyAbel from source, please see the updated instructions in the README.
+* All functions and arguments deprecated in v0.9.0 are removed (PR #413).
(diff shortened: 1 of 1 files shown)


MikhailRyazanov:
> Just to confirm, we're not removing any functionality here, things have just moved to different function names and we're cleaning up the old functions in this PR, correct?

Yes, those deprecated things were basically aliases/shims or are not needed any more.
Before merging, the Changelog needs to be updated, so here is a draft of the entry:
```
* All functions and arguments deprecated in v0.9.0 are removed (PR #413).
  If your code stops working after upgrading PyAbel, downgrage it to v0.9.1 and
  pay close attention to deprecation warnings (or check v0.9.1 documentation
  for suggested replacements).
```

Issue #412, opened 2025-11-05
Implement faster Direct method


DanHickstein:
As mentioned in #411, @MikhailRyazanov has developed new direct-python and direct-C methods that offer improved performance. This is just a reminder to implement those and quantity the speedup for various image sizes.

MikhailRyazanov:
PR #415 with the new implementation is merged now, but I’ll leave this issue open as a reminder to do some other related things:
* [ ] Investigate the left-over artifact “correct for the extra triangle at the start of the integral”.
* [ ] Ensure implicit zero after the last pixel (#217).
* [ ] Maybe switch from OpenMP to threading (for better portability and compatibility).
* [ ] Update benchmark results (after the rest is done).

PR #415, opened 2025-11-17
Faster Direct method

MikhailRyazanov:
Here is the new implementation of the Direct method (#412), with much better performance and real support of non-uniform sampling. Right now this PR is for demonstration and discussions, since some additional things need to be done (at least, more tests <s>and documentation</s>) before it’s ready for reviewing...

Instead of modifying the existing functions, I’ve added their new equivalents, with a ``_new`` suffix, such that they can be compared directly, and the tests demonstrating that the new implementation reproduces old results. After we make sure that everything works as expected, the plan is to replace the old code, providing backward compatibility and deprecating renamed arguments and unneeded functions.

The new Python implementation is much faster because:
* [``numpy.trapezoid()``](https://numpy.org/doc/stable/reference/generated/numpy.trapezoid.html) is surprisingly inefficient, so I wrote its replacement for our specific case (later I’ll try to see whether it can be generalized to improve the NumPy code itself);
* the integration was done in a weird way (always over the whole image width, with the irrelevant part zeroed), and rewriting it naturally made the code noticeably faster and much simpler.

The new C implementation now accepts non-uniform sampling and is also refactored, such that it runs with about the same speed as the old uniform-sampling code (in fact, slightly faster on Intel, but slightly slower on AMD). However, Cython allows replacing ``range()`` with [``prange()``](https://cython.readthedocs.io/en/stable/src/userguide/parallelism.html``) for automatically making loops parallelized to multiple threads (through [OpenMP](https://en.wikipedia.org/wiki/OpenMP)), and this gives a huge performance improvement. For example, on my 10-year-old i5 laptop with 4 logical processors:
```
time in milliseconds

=== Basis generation ==============================
Method              n = 257     n = 513    n = 1025
---------------------------------------------------
two_point               0.7         2.4        13.6

=== Forward Abel transform ========================
Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                2.7        17.7       112.7
direct_C_new            0.8         5.1        35.3
direct_Py_new           7.6        49.5       447.7
direct_Python          67.1       473.6      4491.6

=== Inverse Abel transform ========================
Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                3.9        21.1       130.4
direct_C_new            1.1         6.1        38.0
direct_Py_new           7.8        50.7       438.8
direct_Python          68.1       478.9      4515.2
two_point               0.2         1.4         9.8
```

In the meantime, take a look at the [new examples](
https://pyabel--415.org.readthedocs.build/en/415/transform_methods/direct.html#examples) (the “Source code” there are clickable; they are supposed to look like the “Example” box [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#collapsible-admonitions), but for some reason the RTD theme displays them very unfriendly).

<s>The most important problem to solve now is how to build for macOS, and after that I’ll add some comments about several things that need to be changed in the code itself.</s> (done)

### DIFF
--- a/doc/transform_methods/basex.rst
+++ b/doc/transform_methods/basex.rst
@@ -46,7 +46,7 @@ The recommended way to complete the inverse Abel transform using the BASEX algor
 
     abel.Transform(raw_image, method='basex', direction='inverse').transform
 
-The additional BASEX parameters are described in :func:`abel.basex.basex_transform` an can be passed to :class:`~abel.transform.Transform` using the ``transform_options`` argument.
+The additional BASEX parameters are described in :func:`abel.basex.basex_transform` and can be passed to :class:`~abel.transform.Transform` using the ``transform_options`` argument.
 
 If you would like to access the BASEX algorithm directly (to transform a right-side half-image), you can use :func:`abel.basex.basex_transform`.
 


MikhailRyazanov:
I’m surprised that apparently Apple’s Clang doesn’t have OpenMP support... In principle, removing the ``-fopenmp`` compiler flag should allow building the extension (single-threaded, thus losing the speedup). But NumPy and SciPy use OpenMP and are somehow able to build for macOS, so it’s not impossible.

I don't have macOS to experiment with (but will try later on GitHub VMs), but can somebody see whether something like [this](https://iscinumpy.dev/post/omp-on-high-sierra/) works? Especially, if you need to install something, will the built extension work on a “clean” system (after uninstalling that something or on a different machine)?

stggh:
Amazing work @MikhailRyazanov. I haven't looked at all the detail, but my i7-9700 Dell (running Ubuntu 25.10, Anaconda python 3.14 *) gives:
```
time in milliseconds

=== Basis generation ==============================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
two_point               0.3         1.1         3.9


=== Forward Abel transform ========================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                1.6        10.4        71.4
direct_C_new            0.2         1.3         9.5
direct_Py_new           3.4        18.6       124.9
direct_Python          27.3       178.7      1256.4


=== Inverse Abel transform ========================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                2.0        11.4        77.7
direct_C_new            1.5         5.3        16.8
direct_Py_new           3.6        19.1       130.1
direct_Python          28.5       179.7      1266.0
two_point               0.1         0.3         1.9
```
(*) Thanks for the Cython install flag provided in the Install information:   --no-build-isolation

You are right about Mac OS compiler missing OpenMP. Here are the results without Cython (Mac OS M3 Tahoe 26.1, Anaconda python 3.14):
```
=== Basis generation ==============================
 
Method              n = 257     n = 513    n = 1025
---------------------------------------------------
two_point               0.2         0.6         2.6
 
 
=== Forward Abel transform ========================
 
Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_Py_new           2.3        13.6       110.4
direct_Python          19.4       104.2      1117.0
 
=== Inverse Abel transform ========================
 
Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_Py_new           2.3        13.7       112.1
direct_Python          19.0       105.6      1122.3
two_point               0.1         0.2         1.7

```

I am impressed with the `direct_Py_new` speed up, a factor of 10.

I will take a look at the Clang suggestion for Mac.

MikhailRyazanov:
> I will take a look at the Clang suggestion for Mac.

Thanks! Try also removing ``'-fopenmp'`` from ``extra_compile_args`` and ``extra_link_args``. This should build without OpenMP, at least enabling to compare single-thread performances.

stggh:
Mac OS without the '-fopenmp':

```
PyAbel benchmark run on arm

time in milliseconds

=== Basis generation ==============================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
two_point               0.2         0.6         2.5


=== Forward Abel transform ========================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                0.7         4.3        28.8
direct_C_new            0.4         2.8        19.8
direct_Py_new           2.1        13.6       109.0
direct_Python          13.9       101.6       921.8


=== Inverse Abel transform ========================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                0.7         4.7        31.6
direct_C_new            0.4         2.9        20.6
direct_Py_new           2.1        13.7       110.2
direct_Python          15.6       101.2       954.1
two_point               0.0         0.3         1.8
```

MikhailRyazanov:
> Mac OS without the '-fopenmp':

Not bad. Still faster than the old code and several times faster than Python, so in the worst case we can simply build without OpenMP on macOS.

stggh:
I have managed to build under Mac OS using (*):
```python
        extra_compile_args = ['-Xpreprocessor', '-fopenmp', '-O3', '-ffast-math', '-g0']
        extra_link_args = ['-lomp']
    ext_modules = [
        # ("Path" below is a workaround for Setuptools bug on Windows,
        #  see https://github.com/pypa/setuptools/issues/5093)
        Extension('abel.lib.direct', [Path('abel/lib/direct.pyx')],
                  include_dirs=['/opt/homebrew/anaconda3/include', numpy.get_include()],
```

with `diff`:
```
39,40c39,40
<         extra_compile_args = ['-Xpreprocessor', '-fopenmp', '-O3', '-ffast-math', '-g0']
<         extra_link_args = ['-lomp']
---
>         extra_compile_args = ['-O3', '-ffast-math', '-fopenmp', '-g0']
>         extra_link_args = ['-fopenmp']
45c45
<                   include_dirs=['/opt/homebrew/anaconda3/include', numpy.get_include()],
---
>                   include_dirs=[numpy.get_include()],
```

The `'-Xpreprocessor',  '-fopenmp'` must be first.

NB:  the `'include_dirs=` directory is one of many on my Mac containing `omp.h`. It is also present in `/opt/anaconda3/envs/py312/include` but not in the current `py314`. 

So your `set-up.py` is almost correct, if my Anaconda `py314` can be fixed.

A `x5` execution speed improvement for `direct_C_new` with multi-processor execution:
```
PyAbel benchmark run on arm

time in milliseconds

=== Basis generation ==============================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
two_point               0.3         0.6         2.5


=== Forward Abel transform ========================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                0.7         4.3        29.2
direct_C_new            0.2         1.0         5.3
direct_Py_new           2.6        14.6       125.0
direct_Python          14.0       103.4      1167.3


=== Inverse Abel transform ========================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                0.8         4.7        31.8
direct_C_new            0.3         1.4         6.1
direct_Py_new           2.7        15.4       127.2
direct_Python          14.1       103.4      1149.6
two_point               0.0         0.2         1.8
```

(*) How did you make that nice 'show code' link in a previous discussion? Does the associated file need to be present in the repository?

MikhailRyazanov:
> the `'include_dirs=` directory is one of many on my Mac containing `omp.h`. It is also present in `/opt/anaconda3/envs/py312/include` but not in the current `py314`.

OK, so this is in Anaconda (for which we then can build conda packages relying on OpenMP). Do you have [llvm-openmp](https://anaconda.org/anaconda/llvm-openmp/) installed, maybe automatically as a dependency for something else? This will explain the presence of `libomp` and the difference between the environments.

> How did you make that nice 'show code' link in a previous discussion? 

Had to search for the documentation again (and recall how they call it)...
https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections

> Does the associated file need to be present in the repository?

That one was just a code block. But, I believe, if the file is anywhere on GitHub and public, a [permalink](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet) should work too.

stggh:
> OK, so this is in Anaconda (for which we then can build conda packages relying on OpenMP). Do you have [llvm-openmp](https://anaconda.org/anaconda/llvm-openmp/) installed

No. I have now installed it using `conda` and removed the `setup.py` "include_dirs" path for homebrew. Mac now finds `omp.h`.

Abel timing:
```
PyAbel benchmark run on arm

time in milliseconds

=== Basis generation ==============================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
two_point               0.2         0.6         2.8


=== Forward Abel transform ========================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                0.7         4.4        29.8
direct_C_new            0.2         0.7         4.1
direct_Py_new           2.2        13.9       111.6
direct_Python          14.2       101.9      1090.4


=== Inverse Abel transform ========================

Method              n = 257     n = 513    n = 1025
---------------------------------------------------
direct_C                0.7         4.7        32.5
direct_C_new            0.3         0.8         4.8
direct_Py_new           2.2        14.0       112.8
direct_Python          14.3       102.8      1087.7
two_point               0.1         0.2         1.7
```

MikhailRyazanov:
> > OK, so this is in Anaconda (for which we then can build conda packages relying on OpenMP). Do you have [llvm-openmp](https://anaconda.org/anaconda/llvm-openmp/) installed
> 
> No. I have now installed it using `conda` and removed the `setup.py` "include_dirs" path for homebrew. Mac now finds `omp.h`.

Interesting. Maybe you had something else that installed OpenMP? Because “clean” macOS definitely doesn’t have it...

I’ve tried on GitHub with regular macOS tools (+Homebrew) that installing
```sh
brew install libomp
```
and then setting environment variables
```sh
export CFLAGS="-I$(brew --prefix libomp)/include -Xpreprocessor -fopenmp"
export LDFLAGS="-L$(brew --prefix libomp)/lib -lomp"
```
allows building with OpenMP (without having any `openmp` options in `setup.py`). But after
```sh
brew uninstall libomp
```
the built package fails to find `libomp.dylib` anywhere. (See the test run [here](https://github.com/MikhailRyazanov/Actions/actions/runs/19486937403/job/55771104689); it’s not real PyAbel but a very trimmed-down subset to test various things: it this case that `import abel.lib.count` should report 5 threads when compiled with OpenMP, 1 thread when without OpenMP, and crash if it fails to load.)


If I understood correctly that if `llvm-openmp` is installed in `conda`, you don’t need to add the paths, could you please test that
* building this PR in its current state (when `setup.py` excludes `'-fopenmp'` for macOS) produces a single-threaded library,
* but setting
   ```sh
   export CFLAGS="-Xpreprocessor -fopenmp"
   export LDFLAGS="-lomp"
   ```
   before building (in a fresh source tree to avoid picking up previously compiled files) produces a library with OpenMP?

stggh:
> If it can be installed with OpenMP by just setting CFLAGS and LDFLAGS

The answer is yes. Both single- (without `-fopenmp`) and multi-thread install correctly. BTW I was/am cloning your repository using `-b direct`.

abel/tools/math.py
MikhailRyazanov: [``np.gradient``](https://numpy.org/doc/2.3/reference/generated/numpy.gradient.html) actually does support non-uniform sampling (since at least NumPy 1.13 — this is how far back its documentation goes) and is noticeably faster than this function in all cases. Most ironically, *this* function does not work correctly for irregular sampling (instead of computing the gradient at the point, it is computed at the midpoint between the neighboring points). So we should deprecate it and suggest to use ``np.gradient`` instead.
The new Direct implementation doesn’t need this function because by default it calls ``np.gradient`` directly.

abel/tests/test_math.py
MikhailRyazanov: I would suggest renaming this file to ``test_tools_math.py`` for consistency.

abel/direct.py
MikhailRyazanov: This should be removed as not needed anymore. Since it’s private and not imported by default, hopefully nobody uses it directly.

MikhailRyazanov: I’m inclined to rename ``int_func`` to ``integral`` (leaving the original argument as a deprecated fallback) for consistency with ``derivative``.

DanHickstein:
This seems like a great PR! Where are you at with it? I'm fine if you want to go ahead and merge if you consider it done. 

Also, did you look into starting an issue over at the numpy github regarding potential improvements to the trapezoid function?

abel/direct.py
DanHickstein: Sounds good!

stggh (approved):
@MikhailRyazanov a very impressive speed-up for the `direct` method, excellent analysis and coding

examples/example_direct_rmin.py
```diff
@@ -0,0 +1,28 @@
...
+plt.ylabel('Intensity')
+plt.plot(ref.r, ref.abel, 'C0:')
+plt.plot(cut_r, cut_abel, 'C0', label='Data')
+plt.plot(ref.r, ref.func, 'C1:')
+plt.plot(cut_r, cut_res, 'C1', label='Transform')
+plt.xlim(0, 1)
+plt.ylim(0, None)
+plt.legend()
+plt.tight_layout()
+plt.show()
```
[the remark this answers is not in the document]
MikhailRyazanov: > Why is it that the other methods cannot treat this case?

Technically, because they do not have the ``r`` argument. :–) Theoretically, matrix methods where the transform matrix is triangular can do this by truncating the matrix accordingly (rBasex effectively does this in `example_rbasex_block.py`). But if the matrix “spills” over the diagonal (like in BASEX with its non-local basis functions), the outer elements always depend on the inner, although very slightly. In any case, this case is apparently not sufficiently important, so that nobody cared to address it in other methods. But here it works automatically, so I though that it would be nice to mention this possibility.

> Is there a `1/n` scale factor missing in this figure?

Since you cannot pass `r` to other methods, you *should* pass `dr=ref.dr`. If I understood the question correctly.

doc/transform_methods/basex.rst
```diff
@@ -42,13 +42,13 @@ According to Dribinski et al., BASEX has several advantages:
 How to use it
 -------------
 
-The recommended way to complete the inverse Abel transform using the BASEX algorithm for a full image is to use the :py:class:`abel.Transform <abel.transform.Transform>` class::
+The recommended way to complete the inverse Abel transform using the BASEX algorithm for a full image is to use the :class:`abel.Transform <abel.transform.Transform>` class::
 
     abel.Transform(raw_image, method='basex', direction='inverse').transform
 
-The additional BASEX parameters are described in :py:func:`abel.basex.basex_transform` an can be passed to :class:`Transform <abel.transform.Transform>` using the ``transform_options`` argument.
+The additional BASEX parameters are described in :func:`abel.basex.basex_transform` an can be passed to :class:`~abel.transform.Transform` using the ``transform_options`` argument.
```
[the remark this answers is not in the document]
stggh: The only typo that I came across

examples/example_direct_rmin.py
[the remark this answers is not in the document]
stggh: Yes, you are correct. I have had too long a period away from Abel transforms

Issue #385, opened 2024-08-02
Final entry of 'direct' inverse transform of a 1D array is always zero?


DavidT3:
Hi,

I'm using your module to enable inverse-abel deprojections of surface brightness profiles for astrophysical objects in my module - I let the user choose between most of the implementations you provide, but I've just noticed that the 'direct' transform (whether I use dr, or pass a spatial mesh (radii in my case), I end up with the final element of the output array being zero.

I am using the python implementation.

And a test data set would be:

`dr = 42.22897612626319`

`vals = array([0.05380597, 0.03950476, 0.02695798, 0.02092113, 0.01460821,
       0.01132293, 0.00891306, 0.00671739, 0.00609382, 0.00533249,
       0.00420262, 0.0041189 , 0.0038018 , 0.00354966, 0.0036527 ,
       0.00417479, 0.00426684, 0.00354484, 0.00393238, 0.00360251,
       0.00338861])`

`rads = array([ 21.11448806,  63.34346419, 105.57244032, 147.80141644,
       190.03039257, 232.25936869, 274.48834482, 316.71732095,
       358.94629707, 401.1752732 , 443.40424933, 485.63322545,
       527.86220158, 570.0911777 , 612.32015383, 654.54912996,
       696.77810608, 739.00708221, 781.23605834, 823.46503446,
       865.69401059])`

`direct_transform(vals, r=rads, backend='python', correction=True)`

or 

`direct_transform(vals, dr=dr, backend='python')`

Am I a) being stupid and misunderstanding something? b) using this wrong? c) or is this an expected behaviour for this function?

Cheers,
David

MikhailRyazanov:
The outer point in the “direct” method currently has fixed zero intensity, see #217. In principle, it doesn't need to, but (from [documentation](https://pyabel.readthedocs.io/en/latest/transform_methods/comparison.html#math)):
[...]

Issue #217, opened 2018-07-27
Investigate the edge pixel in the direct method


DanHickstein:
Following the discussion is https://github.com/PyAbel/PyAbel/pull/214#issuecomment-407919764, we may want to think about how to deal with the last element of the output. Currently, this is strictly zero, which makes some degree of sense, because a basic assumption of the Abel transform is that the intensity goes to zero at the last pixel. However, other transform methods (at least the hansenlaw) allow this pixel to have non-zero values. Potentially we could pad the array by an additional value (either zero, or the same as the last pixel) and then trim the final array.

stggh:
> @stggh, do you agree?

No. Unless I have misinterpreted this discussion. Zero-padding will not repair a faulty image (that has a background intensity). This figure show the inverse able transform radial profile for the 'O-' test image, including an added background intensity, with and without a zero-intensity boundary. The background results in an additional structure (~500 pixels). This pseudo structure occurs for the other methods, except `hansenlaw`:
     
![Image](https://github.com/user-attachments/assets/23c2fe31-7785-4d30-911a-9e5b4a81bfbb)

Code:
```python
import abel
import matplotlib.pyplot as plt

fn = 'O-'
aim = abel.tools.analytical.SampleImage(n:=1001, name=fn).abel
print(f'Image intensity maximum = {aim.max():.0f}')

# pseudo backround
aim5 = aim + 5

# zero boundary
aim50 = aim5
aim50[:, 0] = aim50[:, -1] = 0

# inverse Abel transform -------------
method = 'direct'
im = abel.Transform(aim, method=method, angular_integration=True)
im5 = abel.Transform(aim5, method=method, angular_integration=True)
im50 = abel.Transform(aim50, method=method, angular_integration=True)

# plot -----------
plt.plot(*im.angular_integration,label='no background')
plt.plot(*im5.angular_integration, label='background 5')
plt.plot(*im50.angular_integration, '--', label='background 5, 0 edge')

plt.legend()
plt.title(method)
plt.xlabel('radius')
plt.ylabel('amplitude')
plt.savefig('bkg.svg')
plt.show()
```

MikhailRyazanov:
I was mostly concerned about the following behavior when any noise at the edge pixel shifts the whole transform intensity (demonstrated using the low-level 1D transforms for clarity):
<details>
<summary>[code for the plots]</summary>

```python
import abel
import matplotlib.pyplot as plt

[...]

PR #419, opened 2026-02-18
Implicit background for the direct method

MikhailRyazanov:
This should address the issues reported in #217 and #385:
1. That the edge (outermost) pixel of the transform is always zero.
2. That the inverse transform essentially subtracts the edge-pixel intensity from the whole row.

The solution is basically as proposed there — by padding the input date with zeros and then cropping the output accordingly. This is the most straightforward approach, compatible with custom derivative and integral functions and apparently having negligible impact on performance. The padding is made optional (enabled by default) to allow reproducing the original behavior for those who need it.

Also, since padding with some other, non-zero, value can be done without any extra work and will result in treating that value as a user-defined non-zero background, the corresponding new argument is ``background=val`` (instead of ``pad=True``/``False``, which would also refer to internal technical details instead of user-observable behavior).

Another thing in this PR is the removal of the artifact at the penultimate pixel inherited from the old implementation (see #415). I’ve done some tests using step functions:
step_with
step_without
and it seems that the results without that “correction” are at least more consistent than with it (besides that the code becomes cleaner, 712c55c88ee73e275431b9a1a070434329c33517).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ Unreleased
   for suggested replacements).
 * Improvements to direct method (both C and Python backends): much better
   performance and full support of non-uniform sampling (PR #415).
+* Direct method now, by default, assumes zero intensity outside the image frame
+  but can optionally disregard a constant background, as before (PR #419).
 
 v0.9.1 (2025-09-22)
 -------------------


stggh:
> I can do a follow-up PR

Yes, please hack away, thanks @MikhailRyazanov

PR #420, opened 2026-02-20
Implicit background for Hansen–Law method

MikhailRyazanov:
This is basically a follow-up for PR #419, but for ``hansenlaw``, padding the input data with zeros, such that the derivative in the inverse transform captures absolute intensities (rather than relative to the last pixel, see https://github.com/PyAbel/PyAbel/issues/217#issuecomment-3509139101).

As a bonus, this also solves the problem with the missing last column:
(upper row — old results, without padding; lower row — new results, with zero padding)
hold_order=0

Curiously, ``hold_order=1`` produces unbiased results (although less sharp) for both the forward and inverse transforms:
hold_order=1
My understanding of the internals is not enough to identify the reason. Perhaps @stggh can later take a deeper look at this...

<details>
<summary>(code for the plots)</summary>

```python
import numpy as np
from abel.hansenlaw import hansenlaw_transform
from abel.tools.analytical import TransformPair

n = 11
pairs = (TransformPair(n, 9),
         TransformPair(n, 5),
         TransformPair(n, 8))
steps = ('mid', 'pre', 'mid')
funcs = np.vstack([p.func for p in pairs])
abels = np.vstack([p.abel for p in pairs])
r, dr = pairs[0].r, pairs[0].dr

hold_order = 0
for direction, background, subplot in [('forward', None, 1), ('inverse', None, 2),
                                       ('forward', 0,    3), ('inverse', 0,    4)]:
    if direction == 'forward':
        IM, ref = funcs, abels
    else:
        IM, ref = abels, funcs
    tr = hansenlaw_transform(IM, dr=dr, direction=direction,
                             hold_order=hold_order, background=background)
    plt.subplot(2, 2, subplot)
    plt.title(f'{direction}, {background=}')
    for i in range(3):
        ds = None if direction == 'forward' else 'steps-' + steps[i]
        plt.plot(r, ref[i], f'C{i}', ds=ds, label=f'{pairs[i].label}')
        plt.plot([1, 1 + dr], [ref[i][-1], 0], f'C{i}:', ds=ds)
        plt.plot(r, tr[i], f'C{i}.-', lw=0.5)
        plt.plot([r[-1], r[-1] + dr], [tr[i][-1], 0], f'C{i}:', lw=0.5)
    plt.ylim((0, 2.3 if direction == 'forward' else 1.35))
    plt.legend()

plt.suptitle(f'{hold_order=}')
plt.tight_layout()
#plt.savefig(f'{hold_order=}.svg')
plt.show()
```
</details>

### DIFF
--- a/abel/hansenlaw.py
+++ b/abel/hansenlaw.py
@@ -123,10 +123,10 @@ def hansenlaw_transform(image, dr=1, direction='inverse', hold_order=0,
     background : float or None, optional
         Initial conditions at the outer edge. ``None`` reproduces the behavior
         in PyAbel < 0.10.0, were they were taken from the edge column. This
-        lead to the transformed image missing the outermost column (it was
-        filled by duplicating the previous one). Also, the inverse transform
-        used intensities relative to the edge pixel, such that its intensity
-        (if not zero) was essentially subtracted from the whole row.
+        lead to the transformed image missing the outermost column. Also, the
+        inverse transform used intensities relative to the edge pixel, such
+        that its intensity (if not zero) was essentially subtracted from the
+        whole row.
         Default: ``0``.
 
     Returns
@@ -205,12 +205,12 @@ def hansenlaw_transform(image, dr=1, direction='inverse', hold_order=0,
 
     # missing axial column
     aim[:, 0] = aim[:, 1]
-    if background is not None:
+    if background is None:
+        # edge column is zero
+        aim[:, -1] = 0
+    else:
         # crop to original size
         aim = aim[:, :cols]
-    else:
-        # missing edge column
-        aim[:, -1] = aim[:, -2]
 
     if rows == 1:
         aim = aim[0]  # flatten to a vector


abel/hansenlaw.py
```diff
@@ -193,9 +203,14 @@ def I(n, lam, a):
...
-    # missing column at each side of image
+    # missing axial column
     aim[:, 0] = aim[:, 1]
-    aim[:, -1] = aim[:, -2]
+    if background is not None:
+        # crop to original size
+        aim = aim[:, :cols]
+    else:
+        # missing edge column
+        aim[:, -1] = aim[:, -2]
```
MikhailRyazanov: Another thought (suggested by the plots in the PR description): the “missing” last column, most probably, should be filled with zeros instead. This would also make more sense under the assumption that the input is zero at the edge.

MikhailRyazanov: OK, done!
I’ve also removed the statement that previously it was duplicated — hopefully, nobody relied on that artifact.

Issue #422, opened 2026-02-27
Implementation of new methods / Extending regularization


Ashwolaa:
Hello everyone,

First of all, thanks a lot for the work done in this library! I keep recommending it to students and colleagues as it really gathers most of the tools required to get your hands on Abel inversion.
In this issue, I would like to raise two comments on the current implementation in the hope of triggering some discussions.

# Context
- I think adding new methods is key, one place to find them all is really the gold criteria for any users. It simplifies comparison and narrows down the field of research. Keeping similar syntax between methods makes the testing trivial and gives the most freedom.
- Commenting and sourcing the methods are the first key factors when one wants to understand the approach.

I believe the popularity of the library can easily be connected to these two points and I believe it can also lead in interest and possibily increase the number of contributions. 
I think the work realized by [ADrone7](https://github.com/ADrone7) in #421 is a good start up, and congrats to him for making this very impressive review of the different methods available. I propose in this issue to start listing the missing methods that could be worth implementing. I am not sure on the best way to approach this. But here is the first draft that I (or a maintainer) can edit/complete according to suggestions.

# Transformation Methods
## Abel based
- [ ]  PBasex (implementation of https://github.com/e-champenois/CPBASEX with pyabel convention)
- [ ] Abel Davis (DAVIS: A direct algorithm for velocity-map imaging system 10.1063/1.5025057
- [ ] Fourier Transform (A fast algorithm for the inversion of Abel’s transform 10.1016/j.amc.2016.12.009 / https://arxiv.org/abs/2206.00448 )
- [ ] Chebyshev wavelets (Abel inversion method using Chebyshev wavelets for tomographic reconstruction of an axisymmetric medium 10.1364/AO.424172
##  Non-Abel based
- [ ]  Inverting ion images without Abel inversion: maximum entropy reconstruction of velocity map ( 10.1039/C3CP53673D)
- [ ] MELEXIR: maximum entropy Legendre expanded image reconstruction. A fast and efficient method for the analysis of velocity map imaging or photoelectron imaging data 10.1039/C9CP03353)

# Regularization Methods
I am wondering if regularization methods should be factorized for methods dealing with matrix inversion. This is not a strong opinion but mostly aim at collecting feedbacks.

# Additional points

- Couldn't we make the default import as pyabel instead of abel? This triggered me for some time as there is a clear conflict with https://pypi.org/project/abel/ if one is not careful enough

MikhailRyazanov:
[...]

PR #423, opened 2026-02-28
Conversion from abel to pyabel

Ashwolaa:
Fixes import issue from #422

This PR implements a modification of the import name from abel to pyabel making it consistent with the library name and avoiding possible conflict with other library (https://pypi.org/project/abel/)

`import abel` or `import abel.module` will still work as before but will now trigger a deprecation warning.  

I went for this line:
` DeprecationWarning: The 'abel' package has been renamed to 'pyabel'. Please update your imports: replace 'import abel' with 'import pyabel'. The 'abel' compatibility shim will be removed in a future version.`

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -52,7 +52,7 @@ Before submitting a pull request, be sure to run the unit tests. The test suite 
     
 For more detailed information, the following can be used::
 
-    pytest abel/  -v  --cov=abel
+    pytest pyabel/  -v  --cov=abel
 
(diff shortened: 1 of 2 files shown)


DanHickstein:
Hi @Ashwolaa! Thanks for putting the effort into this PR. However, I think that I agree with @MikhailRyazanov (https://github.com/PyAbel/PyAbel/issues/422#issuecomment-3978336250) that it seems like the abel package doesn't have many users and it doesn't justify adding the "py" to our import line. Is the abel package under active development? Does it have a broad user base?

Ashwolaa:
I understand your view, that is fine by me . I will close it if you are not interested.

PR #421, opened 2026-02-23
Add Nestor-Olsen transform method

ADrone7:
Added Nestor, Olsen transform method (linear probe variant) as described in O. H. Nestor and H. N. Olsen, "Numerical Methods for Reducing Line and Surface Probe Data", `SIAM Review, vol. 2, no. 3, 1960, pp. 200–07 <https://doi.org/10.1137/1002042>`__.
This is one of the first numerical Abel transform methods published in literature. The method can be derived similarly to "Three point", but the coefficient of the linear term is assumed to be zero. It makes the method resemble "Three point" in terms of accuracy and computational performance. 
plot_example_gaussian_nestorolsen

PR #424, opened 2026-03-05
Forward transform for the Dasch methods

MikhailRyazanov:
As discussed in the implementation of the Nestor–Olsen method (https://github.com/PyAbel/PyAbel/pull/421#issuecomment-3982844833 and reply), the Dasch methods can be complemented with the *forward* Abel transform using the same approach, namely, inverting the inverse-transform matrix. So this PR does exactly that.

Some representative benchmarks (decade-old i5 with 2 cores, 4 threads, so don’t pay attention to absolute numbers):
```
=== Basis generation ==========================================
Method              n = 257     n = 513    n = 1025    n = 2047
---------------------------------------------------------------
onion_peeling           1.2         3.4        17.3        90.5
three_point             3.0         9.4        43.8       179.9
two_point               0.7         2.4        13.7        49.9

=== Forward Abel transform ====================================
Method              n = 257     n = 513    n = 1025    n = 2047
---------------------------------------------------------------
onion_peeling           0.2         1.6         8.6        52.5
three_point             0.9         3.3        24.0       315.5
two_point               0.2         1.4         7.7        51.4

=== Inverse Abel transform ====================================
Method              n = 257     n = 513    n = 1025    n = 2047
---------------------------------------------------------------
onion_peeling           0.2         1.5         9.9        97.0
three_point             0.2         1.3         9.9        93.5
two_point               0.2         1.3         9.8        93.5
```
the “Basis” and “Inverse” timings should not be affected, so this only shows that the forward transform is somewhat faster than the inverse for ``onion_peeling`` and ``two_point`` but ~3 times slower for ``three_point`` (which is expected and probably fine because we don’t care about forward transforms so much).

P.S. When adding the forward-transform tests, I also took opportunity to strengthen other checks (they were orders of magnitude too loose) and do some code cleanup (Pylint complaints were partially useful).


stggh:
Fast work @MikhailRyazanov 

Timing for i7-9700 MKL (~/anaconda3/envs/MKL_env/lib/python3.14/)

```
=== Basis generation ==========================================

Method              n = 257     n = 513    n = 1025    n = 2047
---------------------------------------------------------------
onion_peeling           0.4         1.2         7.2        29.0
three_point             1.3         4.4        16.8        76.3
two_point               0.3         1.2         6.2        25.1


=== Forward Abel transform ====================================

Method              n = 257     n = 513    n = 1025    n = 2047
---------------------------------------------------------------
onion_peeling           0.1         0.4         2.1        13.2
three_point             0.5         1.5         7.3        91.2
two_point               0.1         0.6         2.1        13.0


=== Inverse Abel transform ====================================

Method              n = 257     n = 513    n = 1025    n = 2047
---------------------------------------------------------------
onion_peeling           0.1         0.3         2.0        20.3
three_point             0.1         0.3         1.9        19.7
two_point               0.1         0.6         1.9        19.9
```

This is with `repeat=5`, as there is some variation in the `three_point` forward-Abel benchmarks, that can vary between 65 and 90 seconds.

Perhaps "interestingly" the non-MKL python version, but still Anaconda python 3.14, gives:
```
=== Basis generation ==========================================

Method              n = 257     n = 513    n = 1025    n = 2047
---------------------------------------------------------------
onion_peeling           2.3         1.5         9.4        27.1
three_point             1.4         4.9        17.2        74.8
two_point               0.3         1.7         6.5        26.1


=== Forward Abel transform ====================================

Method              n = 257     n = 513    n = 1025    n = 2047
---------------------------------------------------------------
onion_peeling           0.4         0.7         4.4        15.9
three_point             7.5        10.7        30.5       184.7
two_point               0.3         0.9         3.5        16.9


=== Inverse Abel transform ====================================

Method              n = 257     n = 513    n = 1025    n = 2047
---------------------------------------------------------------
onion_peeling           0.1         1.4         5.9        24.7
three_point             0.1         0.3         1.9        20.8
two_point               0.1         0.3         2.0        21.2

```

i.e. x2 larger time to evaluate the `three_point` forward transform

stggh:
There may be an issue with the intensity scaling for the  "forward" transform:
       forward

forward transform requires a reciprocal radial spacing to give the same intensity as the other methods (in the test below "direct"):

```python
import abel

im = abel.tools.analytical.TransformPair(n=101, profile=(s:=3))

x = abel.dasch.two_point_transform(im.func, dr=1/im.dr, direction='forward')
y = abel.dasch.three_point_transform(im.func, dr=1/im.dr, direction='forward')
d = abel.direct.direct_transform(im.func, dr=im.dr, direction='forward')

plt.plot(im.r, im.abel, lw=3, label=f'profile {s}')
plt.plot(im.r, x, label='2pt')
plt.plot(im.r, y, label='3pt')
plt.plot(im.r, d, label='direct')
plt.legend()

plt.savefig('forward.svg')
plt.show()
```

PS: Re benchmarks (above) my system timings are sensitive to system loads

MikhailRyazanov:
> There may be an issue with the intensity scaling for the "forward" transform

Yes, the forward transform definitely must *multiply* by $dr$, by definition of the integral. I’ve corrected this and also updated the tests to make sure that $dr \neq 1$ cases are tested (also using our standard and tested ``GaussianAnalytical`` instead of an error-prone made-up gaussian).

> PS: Re benchmarks (above) my system timings are sensitive to system loads

I typically use ``t_min=10`` for final measurements (not sure whether this will affect your results) and make sure that nothing else runs during the benchmarks. It is more surprising to me that these “non-MKL” results are much worse for smaller image sizes, like there is some large overhead. I don’t see how this can be improved, except caching the forward-transform matrix, but this will complicate the code, which is hardly justified in this case...

abel/dasch.py
```diff
@@ -147,6 +145,34 @@ def dasch_transform(IM, D):
...
+    Returns
+    -------
+    fwd_IM : 2D numpy array
+        forward Abel transform according to inverted deconvolution operator D
+    """
+
+    if D[1, 0] == 0:
+        # D for 'two_point' and 'onion_peeling' is upper-triangular, so back
+        # substitution can be used (it is more efficient)
+        return solve_triangular(D, IM.T).T
```
stggh, line 170: This is neat. Thanks!

stggh (approved):
Looks very good to me. 

Given that the Abel forward transform is now quite common in the `pyabel` transform methods, the statement in `doc/transform_methods/hansenlaw.rst`, line 25:

> The algorithm is efficient, one of the few methods to provide both the **forward** Abel and **inverse** Abel transform.

is incorrect. Perhaps, delete the words "one of the few methods to" and change "provide" to "providing"  ?

MikhailRyazanov:
> Perhaps, delete the words "one of the few methods to" and change "provide" to "providing" ?

Done, thanks!

BTW, I’m now looking whether ``hansenlaw`` can be sped up similarly to ``direct``. The Python implementation seems to be as efficient as it could be but the recursion cannot be parallelized efficiently, so maybe implementing that critical part in Cython will help. It’s sad to see that an $O(n^2)$ method is slower than $O(n^3)$ for most practical sizes...

stggh:
> BTW, I’m now looking whether hansenlaw can be sped up similarly to direct ... implementing that critical part in Cython will help. 

I don't doubt that you can achieve significant gains, but, personally, I see Cython as extra baggage, complicating the installation process. For my own needs the reduction in execution time is not needed.

DanHickstein (approved):
Very cool! Good to have another method for the forward transform.

Issue #246, opened 2019-03-18
Subnormals and other performance issues


MikhailRyazanov:
This is a continuation of the discussion from #244 about the strange benchmark results obtained on Intel CPUs.

Basically, the observation was that on various Intel CPUs `basex` and `basex(var.reg.)` do the forward transform much slower than the inverse transform (https://github.com/PyAbel/PyAbel/pull/244#issuecomment-468479433), whereas on an AMD CPU they have the equal speeds (https://github.com/PyAbel/PyAbel/pull/244#issuecomment-468353490), as expected.

This is apparently caused by a significant penalty of working with [subnormal floating-point numbers](https://en.wikipedia.org/wiki/Denormal_number) that Intel CPUs have (and at least AMD Ryzen does not).

The “proper” way to avoid this penalty, recommended by Intel is to [set the related CPU flags](https://software.intel.com/en-us/mkl-linux-developer-guide-operating-on-denormals) such that subnormals are treated as zeros. The problem is that neither NumPy (numpy/numpy#4581), nor the underlying matrix libraries (OpenBLAS (xianyi/OpenBLAS#1237), Intel MKL, Apple's Accelerate Framework) expose this capability.

There is, however, a proof-of-concept module [daz](https://github.com/chainer/daz), which does just that (sets the CPU flags) from Python. It works as expected with MKL (see also [matrix benchmarks](https://gist.github.com/MikhailRyazanov/63823e1bc7028085389d787b765a4f2c)):
```
Python 2.7.15 |Anaconda, Inc.| (default, Dec 14 2018, 19:04:19) 
[GCC 7.3.0] on linux2
using Intel MKL

=== Basis generation ==============================   with daz:

Method              n = 513    n = 1025    n = 2049   n = 513    n = 1025    n = 2049
---------------------------------------------------   -------------------------------
basex                1486.2      9070.2     43671.3    1452.8      8901.9     43356.8
linbasex               19.8        74.2       269.9      19.8        68.7       280.1
onion_peeling           4.1        18.6       101.7       4.2        18.1        97.2
three_point            14.4        50.7       212.7      14.4        49.5       205.1
two_point               3.2        12.7        50.7       3.2        12.8        51.8


=== Forward Abel transform ========================   with daz:

Method              n = 513    n = 1025    n = 2049   n = 513    n = 1025    n = 2049
---------------------------------------------------   -------------------------------
basex                   1.2        50.5       340.1       1.2         8.6        64.9
[...]

PR #425, opened 2026-03-08
Cython backend for the Hansen–Law method

MikhailRyazanov:
Although @stggh seems to dislike Cython extensions (https://github.com/PyAbel/PyAbel/pull/424#issuecomment-4014979985), I’ve tried to see whether implementing the Hansen–Law recursion in Cython can help to make this potentially faster method (asymptotically $O(n^2)$ compared to $O(n^3)$ for others) actually faster for practically relevant image sizes.

It turned out that this indeed can make it almost an order of magnitude faster, and finally faster than ``two_point`` (the fastest method with very similar transform results) at 1-megapixel size:
HL
The speed-up, in turn, revealed that calculations of the driving function (and padding) actually take a noticeable fraction of time. They should be memory-bound, so there is not much to improve, but I’ve tried several alternatives to find the fastest approach. This also improved the Python timings slightly. However, the major speed-up for Python is the elimination of ``B0`` from calculations for ``hold_order=0``.

The benchmarks above are for an old 4-threads Intel CPU. Here is the relatively recent AMD (16 threads):
HL-AMD
As mentioned above, the speed becomes mostly memory-bound, so the improvement with more threads is not much better (in contrast to matrix multiplication for ``two_point``), but still keeps ``hansenlaw_C`` faster than ``two_point`` for $n \gtrsim 1000$. (Notice also the throughput drop at $n \gtrsim 2000$, where the CPU cache size is exceeded.) Nonetheless, multithreading does help:
<details>
<summary>
Same AMD but single-threaded
</summary>

Running with ``OMP_NUM_THREADS=1``:
HL-AMD-ST
Cython is still ~2× faster than Python + NumPy.
</details>

I’ve tried to keep the extension not intrusive, with minimal modifications of the original code and falling back to it if Cython is not available, so hopefully @stggh will not have strong objections. :–) In any case, we could keep at least the first 3 commits, which improve the Python code.

P.S. This extension is combined in a single file with ``direct_C`` because Cython does not allow compiling multiple source files into a single library. If compiled separately, the overall size would increase by a factor of 2 because each library includes all the interfacing stuff, which is larger than the function code itself. But the size of combined library is only ~10% larger than is was with ``direct_C`` alone.


stggh:
Impressive @MikhailRyazanov! I knew that I would be eating my words. The `hansenlaw` Abel transform is now very fast. Here is a benchmark from a Macbook M3 MacOS Tahoe 26.3.1 Anaconda Python 3.14.3(*):
```
=== Basis generation ===========================================

Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
two_point                0.2         0.6         2.6        13.2


=== Forward Abel transform =====================================

Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
hansenlaw_C              0.2         0.7         2.9        11.9
hansenlaw_Python         0.7         2.3         8.0        59.5
two_point                0.1         0.3         2.0        11.2


=== Inverse Abel transform =====================================

Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
hansenlaw_C              0.2         0.8         3.6        13.9
hansenlaw_Python         0.8         2.4         8.7        61.2
two_point                0.0         0.2         1.8        10.4
```

(*) i7-9700 unavailable for a fews days, due to home carpet installation = every non-wet area room emptied of furniture. This M3 appears faster in any case (pity it does not yet run Linux).

MikhailRyazanov:
Great, thanks!

> Macbook M3 MacOS

Then it probably runs ``hansenlaw_C`` in one thread (``setup.py`` currently doesn’t use OpenMP in macOS because its support is disabled in default Apple’s Clang). I would expect multithreaded performance to be additionally ~2 times better.

> i7-9700 unavailable for a fews days

Not a problem. :–)

P.S.
>pity it does not yet run Linux

<s>I don’t know about those new ARM Macbooks, but previous, Intel-based ones, were booting Debian Live from a USB stick or SD card just fine. The only problem was that specific WiFi kernel modules were not included, so either they had to be downloaded separately, or some more common WiFi/Ethernet USB adapter be used.</s> (Debian can’t run on M3 yet; Asahi Linux says that “Apple Silicon hardware does not support booting from USB storage, at all. ... This is by design, for security reasons.”)

MikhailRyazanov:
``AbelTiming`` docstring was missing ``daun`` and ``rbasex``, plus ``onion_bordas`` has no closing quote, so these are corrected.

For convenience, I also modified ``pyabel-timing.py`` to allow specifying just ``direct`` or ``hansenlaw`` to benchmark both C and Python backends. Inputs ``..._C`` and ``..._Python`` still do the specific one, as before.
By the way, what was the order of methods there? I though that it looked like from faster to slower, but this wasn’t exactly true. I’ve reordered them alphabetically, but can do differently if needed (and add an explaining comment there).

stggh:
>  modified pyabel-timing.py to allow specifying just direct or hansenlaw

Great, can this simplification naming feature be added to `benchmark.py` too?

> Asahi Linux says that “Apple Silicon hardware does not support booting from USB storage, at all. ... This is by design, for security reasons.”

Sure, thanks very much Apple.

MikhailRyazanov:
> > modified `pyabel-timing.py` to allow specifying just `direct` or `hansenlaw`
> 
> Great, can this simplification naming feature be added to `benchmark.py` too?

Yes, I was thinking whether it should be done there but was not sure how. The concern is that the results have “methods as keys”, thus passing ``select='direct'`` and then getting a ``KeyError`` exception for ``res.iabel['direct']`` might be confusing. Some possible solutions what the do in this case:
1. Just document that ``'direct'`` is a shorthand for ``['direct_C', 'direct_Pyhton']``.
2.  Add ``'direct'`` to the resulting dictionaries as an alias to the default backend (then this alias probably should be skipped in ``__repr__``).
3. Benchmark *only the default backend*. This is probably most friendly for end users who don’t care about internals; and those who care can still use full names, as it is now.

(In any case, `pyabel-timing.py` should expand ``'direct'`` to ``['direct_C', 'direct_Pyhton']`` because it needs to write to different files.)

MikhailRyazanov:
@stggh, when you get to i7-9700, before running full benchmarks, would you please check how the performance depends on the number of threads? I’ve made a script that runs benchmarks with different values of ``OMP_NUM_THREADS``:
<details>
<summary>script</summary>

```python
import os
from subprocess import run

import numpy as np
from matplotlib import pyplot as plt

#method = 'direct'
method = 'hansenlaw'
n = 1025
#n = 8193
script = f"""
import abel
res = abel.benchmark.AbelTiming({n=}, select='{method}_C',
                                repeat=3, t_min=5, verbose=False)
ftime = res.fabel['{method}_C'][0]
itime = res.iabel['{method}_C'][0]
print(f'{{ftime:.5g}}\t{{itime:.5g}}', end='')
"""

threads = []
ftimes = []
itimes = []

for t in range(os.cpu_count() + 1):
    if t != 0:
        os.environ['OMP_NUM_THREADS'] = str(t)
    out = run(['python3', '-c', script], capture_output=True).stdout.decode()
    print(f'{t}\t{out}')
    threads.append(t)
    ftime, itime = out.split()
    ftimes.append(float(ftime))
    itimes.append(float(itime))

threads = np.array(threads[1:])
ftime0 = ftimes[0]
ftimes = np.array(ftimes[1:])
itime0 = itimes[0]
itimes = np.array(itimes[1:])

plt.suptitle(f'{method}, {n=}')

plt.subplot(211)
plt.xlabel('Threads')
plt.ylabel('Transform time, ms')
plt.axhline(ftime0, c='C0', ls='--')
plt.plot(threads, ftimes, label='forward')
plt.axhline(itime0, c='C1', ls='--')
plt.plot(threads, itimes, label='inverse')
plt.xlim(left=0)
plt.ylim(bottom=0)
plt.legend()

fsp = ftimes[0] / ftimes
isp = itimes[0] / itimes

plt.subplot(223)
plt.xlabel('Threads')
plt.ylabel('Speed-up')
plt.plot(threads, fsp, label='forward')
plt.plot(threads, isp, label='inverse')
plt.xlim(left=0)
plt.ylim(bottom=0)

plt.subplot(224)
plt.xlabel('Threads')
plt.ylabel('Efficiency')
plt.plot(threads, fsp / threads, label='forward')
plt.plot(threads, isp / threads, label='inverse')
plt.xlim(left=0)
plt.ylim(bottom=0)

plt.tight_layout()
plt.savefig(f'OMPeff_{method}_{n}.svg')
plt.show()
```
</details>

On the old laptop with 4 logical CPUs, the results get better monotonically, so the default OpenMP behavior (using 1 thread per logical CPU) gives the best results. However, on the recent AMD with 16 logical CPUs, the picture is not so good (dashed lines = default):

| Linux | Windows
| --- | --- |
| OMPeff_hansenlaw_1025 | OMPeff_hansenlaw_1025
| OMPeff_hansenlaw_8193 | OMPeff_hansenlaw_8193

The speed doesn’t increase much with more than 4–6 threads and in some cases actually drops (despite increasing the overall CPU load).

If you see similar behavior on i7, then I think, it would make sense to limit ``parallel()`` to ``parallel(num_threads=4)`` or so.

For comparison, ``direct`` is much more computationally intensive and is not limited by memory bandwidth as much, so its scaling looks reasonable, especially for large images:
<details>
<summary>similar plots for direct</summary>

| Linux | Windows
| --- | --- |
| OMPeff_direct_1025 | OMPeff_direct_1025
| OMPeff_direct_8193 | OMPeff_direct_8193

</details>

thus default OpenMP behavior works fine there.

stggh:
> please check how the performance depends on the number of threads

Will do, and thanks for the script

MikhailRyazanov:
> Will do

Thanks!

To bother you more: do you have any plans for updating the Raspbery Pi benchmarks? If not, maybe you will be interested in replacing them by Macbook M3 MacOS? (To have something for an ARM CPU and a third OS there, although I’m not sure whether the transform speed is important for typical PyAbel applications on macOS.)

stggh:
Finally, some `script` results. Sorry, for the delay, it took a while to put my filing cabinet desk back together, being too large to fit through the internal door!

**i7-9700 CPU**

_Ubuntu 25.10, Anaconda Python 3.14.3_
- 1025
 ```
0	3.0467	3.6097
1	4.8744	5.5248
2	3.6245	4.2889
3	3.111	3.918
4	2.9614	3.7239
5	2.8172	3.623
6	3.0085	3.6614
7	2.9079	3.66
8	3.0323	3.7656
 ```
- 8193
 ```
0	97.91	141.92
1	231.81	280.64
2	149.23	197.62
3	120.03	169.53
4	107.15	156.57
5	98.27	146.57
6	96.426	145.41
7	92.983	141.39
8	92.518	140.65
 ```
OMPeff_hansenlaw_1025
OMPeff_hansenlaw_8193

_MKL, Ubuntu 25.10, Anaconda Python 3.14.3_
- 1025
```
0	3.0008	3.6501
1	4.8623	5.5158
2	3.4673	4.3257
3	3.17	3.8375
4	2.8772	3.6958
5	3.2318	3.6396
6	2.9019	3.677
7	3.0355	3.6697
8	2.9939	3.7682
```
- 8193

```
0	93.752	142.32
1	233.99	282.88
2	148.29	197.85
3	120.87	170.24
4	107.09	156.3
5	100.3	148.69
6	95.763	145.24
7	93.647	142.34
8	93.923	143.02
```

1025-MKL
8193-MKL


**Macbook M3 pro, Tahoe 26.3.1**
_Anaconda Python 3.14.3_
- 1025
```
0	2.8491	3.5272
1	2.8456	3.5955
2	2.8369	3.4922
3	2.8439	3.5162
4	3.0532	3.5106
5	2.8312	3.5035
6	2.8266	3.4967
7	2.8371	3.5205
8	2.8347	3.5082
9	2.8362	3.5577
10	2.8835	3.8275
11	2.8419	3.5054
12	2.8382	3.5085
```
- 8193
```
0	182.78	206.65
1	176.5	212.76
2	176.94	208.84
3	180.1	214.55
4	176.62	207.74
5	177.44	206.96
6	176.59	213.56
7	178.15	207.94
8	186.86	217.97
9	186.5	217.72
10	179.29	209.46
11	187.19	231.52
12	180.89	211.64
```
1025-MacOS
8193-MacOS

stggh:
> Raspberry Pi benchmarks? If not,  ... replacing them by Macbook M3 MacOS

Include both? It is kind-of-neat that the Rpi is only ~13x slower (`two_point` inverse Abel), and about 1/13th the cost. 

**Rpi 5B, ARM Cortex-A76, 8GB RAM, NVMe storage**  (~$A400-, rough guide with accessories)
_Debian GNU/Linux 13 (trixie), Anaconda Python 3.14_
```
=== Basis generation ===========================================

Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
basex                  210.0      1097.0      6885.4     48399.7
daun                     1.3         2.9         6.7        16.4
linbasex                 8.4        24.0       101.7       411.3
nestorolsen              1.4         7.3        31.1       127.9
onion_peeling            1.4         7.2        19.1       335.9
rbasex                   5.0        13.0        41.7       191.8
three_point              3.0        10.5        45.8       227.6
two_point                0.7         2.8        13.2        54.9


=== Forward Abel transform =====================================

Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
basex                    0.4         4.1        21.1       184.3
basex(var.reg.)         27.9       119.6       174.9       827.8
daun                     0.4         5.5        22.6       144.2
direct_Python            6.4        37.8       384.2      4204.6
hansenlaw_Python         2.7         8.4        28.4       118.8
nestorolsen              8.0         3.5        16.9       103.0
onion_peeling            0.5         3.6        15.8       109.2
rbasex                   0.9         4.3        21.0        92.9
rbasex(None)             0.5         2.2         9.6        43.0
three_point             13.0        47.2       123.9       309.4
two_point                0.5         4.8        17.3        94.1


=== Inverse Abel transform =====================================

Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
basex                    0.4         2.8        22.2       144.2
basex(var.reg.)         30.5       100.6       109.0       502.1
daun                     0.3         4.7        15.8        97.4
daun(var.reg.)          52.9       171.1       149.9       481.6
direct_Python            8.1        40.6       396.4      3487.8
hansenlaw_Python         2.8         8.8        31.0       130.2
linbasex                51.6       280.2      1792.9     12780.5
nestorolsen              0.2         2.7        22.3       146.2
onion_bordas            80.8       362.4      1931.0     50033.4
onion_peeling            0.5         3.8        25.5       198.1
rbasex                   0.9         4.2        20.9        92.9
rbasex(None)             0.5         2.2         9.6        42.8
three_point              0.2         2.4        21.4       141.2
two_point                0.2         2.4        21.3       142.5
```

**Mac M3 OS, 36 GB RAM** (~$A5,000+)
```
=== Basis generation ===========================================

Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
basex                   48.1       254.5      1665.8     12467.0
daun                     0.3         0.7         1.6         3.9
linbasex                 1.6         6.5        27.6       109.1
nestorolsen              0.4         1.7         8.6        35.7
onion_peeling            0.2         0.9         4.1        19.1
rbasex                   1.3         3.1         8.6        35.9
three_point              0.8         2.3         9.3        46.9
two_point                0.2         0.6         2.6        12.8


=== Forward Abel transform =====================================

Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
basex                    0.0         0.2         1.5        10.4
basex(var.reg.)          0.4         1.3         7.2        41.1
daun                     0.0         0.2         1.7        12.5
direct_C                 0.4         2.9        20.7       150.3
direct_Python            2.1        13.9       108.8       954.6
hansenlaw_C              0.2         0.7         2.9        11.1
hansenlaw_Python         0.7         5.8         8.0        87.1
nestorolsen              0.1         0.3         2.2        11.9
onion_peeling            0.1         0.3         2.0        11.1
rbasex                   0.4         1.3         6.7        26.1
rbasex(None)             0.2         0.8         3.8        15.4
three_point              0.2         0.8         4.3        24.2
two_point                0.1         0.4         2.0        10.8


=== Inverse Abel transform =====================================

Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
basex                    0.0         0.2         1.5        10.5
basex(var.reg.)          0.4         1.3         7.4        43.3
daun                     0.1         0.3         2.2        11.9
daun(var.reg.)           0.2         1.0         6.3        37.8
direct_C                 0.4         3.0        21.0       153.4
direct_Python            2.2        14.1       111.0       719.1
hansenlaw_C              0.2         0.8         3.6        13.5
hansenlaw_Python         0.8         2.5         8.8        90.8
linbasex                14.7        61.4       278.5      1683.7
nestorolsen              0.1         0.3         2.0        12.1
onion_bordas            20.4        90.3       506.6      7200.2
onion_peeling            0.0         0.2         1.5        11.0
rbasex                   0.4         1.3         6.7        26.0
rbasex(None)             0.2         0.8         3.8        15.5
three_point              0.1         0.2         1.8        11.2
two_point                0.1         0.2         1.7        11.2
```

I will run `pyabel_timing.py` later

MikhailRyazanov:
> Finally, some `script` results.

Thanks! It seems that using over 4 threads indeed doesn’t help much in most cases (and even when helps, the gain is very small, +10–15% speed for +100% CPU load), so I’ve fixed ``num_threads=4`` in the code.

> Macbook M3 pro, Tahoe 26.3.1

It cannot use OpenMP and always runs single-threaded.

>> Raspberry Pi benchmarks? If not, ... replacing them by Macbook M3 MacOS
>
> Include both?

If it’s not too much work, then definitely! Feel free to push directly to this branch. We’ll need updated benchmarks for ``hansenlaw`` and ``direct`` (both backends for each) and missing ``nestorolsen`` benchmarks (i7-9700 already has them).

P.S. I’ve replaced AMD results for Ryzen 5 5600G by Ryzen 7 6800U because running new benchmarks on the first one would be difficult now. However, they both have very similar performance (this “Ryzen 7” is a low-power model).

stggh:
I have pushed new `i7-9700_Linux` benchmarks:

benchmarks_i7_9700_Linux

Not sure what happened to `dawn(var)`/`basex(var)`.

`sensors` and `top` output, during 3pt `n=32769`:
```
pch_cannonlake-virtual-0
Adapter: Virtual device
temp1:        +42.0°C  

acpitz-acpi-0
Adapter: ACPI interface
temp1:        +27.8°C  

coretemp-isa-0000
Adapter: ISA adapter
Package id 0:  +83.0°C  (high = +86.0°C, crit = +100.0°C)
Core 0:        +81.0°C  (high = +86.0°C, crit = +100.0°C)
Core 1:        +80.0°C  (high = +86.0°C, crit = +100.0°C)
Core 2:        +83.0°C  (high = +86.0°C, crit = +100.0°C)
Core 3:        +81.0°C  (high = +86.0°C, crit = +100.0°C)
Core 4:        +83.0°C  (high = +86.0°C, crit = +100.0°C)
Core 5:        +82.0°C  (high = +86.0°C, crit = +100.0°C)
Core 6:        +81.0°C  (high = +86.0°C, crit = +100.0°C)
Core 7:        +80.0°C  (high = +86.0°C, crit = +100.0°C)

nvme-pci-0100
Adapter: PCI adapter
Composite:    +35.9°C  (low  = -20.1°C, high = +77.8°C)
                       (crit = +81.8°C)
Sensor 1:     +35.9°C  (low  = -273.1°C, high = +65261.8°C)


Tasks: 303 total,   2 running, 296 sleeping,   0 stopped,   5 zombie
%Cpu0  :100.0 us,  0.0 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
%Cpu1  :100.0 us,  0.0 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
%Cpu2  :100.0 us,  0.0 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
%Cpu3  :100.0 us,  0.0 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
%Cpu4  :100.0 us,  0.0 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
%Cpu5  :100.0 us,  0.0 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
%Cpu6  :100.0 us,  0.0 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
%Cpu7  :100.0 us,  0.0 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
MiB Mem :  31363.4 total,  20943.6 free,  10281.6 used,    895.1 buff/cache     
MiB Swap:  65536.0 total,  56510.3 free,   9025.7 used.  21081.8 avail Mem 

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND  
  11279 stg121    20   0   17.1g   8.2g   9940 R 800.0  26.7     38,23 python   
```

Note, the machine for `i7-6700_Linux` has moved on. Still running Mac M3, and Raspberry Pi 5B timings. 

Edit: more info

stggh:
Added `RPi5_Linux` benchmarks: 8GB Cortex-A76 CPU-max 2.4 GHz, Debian 13, Anaconda python 3.14.3. No Cython

benchmarks_Rpi5B_Linux

Note, `RPi4_Linux` updated benchmarks, also unavailable. My RPi4Bs are used for `Home Assistant` and `PiHole`.

MikhailRyazanov:
It should be totally fine to remove i7-6700 and RPi4 if they are not available, since we just want to show some representative examples.

> Not sure what happened to ``dawn(var)``/``basex(var)``.

Do you use ``daz`` there? It would not explain this behavior, especially for ``daun``, but I have no other ideas.
In principle, we can use the old results for I7-9700, if they did not change much. However, seeing similar stuff on RPi5 is weird. I’ll try the newest NumPy on my Intel and single-board ARM to see whether they are affected (the AMD is apparently not).

BTW, why ``RPi4_Linux`` has no Cython? I suspect that smaller CPUs should benefit most from the Cython backends (see that even ``hansenlaw_Python`` gets close to the fastest methods already at $n \approx 10^3$, so ``hansenlaw_C`` should be the absolute winner for $n \gtrsim 500$).

stggh:
> BTW, why RPi4_Linux has no Cython? 

This is probably my ignorance, I meant that the `PyAbel` C-extensions do not build for the rasberry pi installation:   `pip install . --no-build-isolation`

stggh:
> Do you use daz there?

No, [daz](https://github.com/chainer/daz) no longer appears to be installable

stggh:
Macbook M3 pro benchmarks in  `M3_MacOS` pushed:
benchmarks_M3_MacOS

I will update the `doc` information (later), and rerun the contentious calculations

MikhailRyazanov:
> > BTW, why RPi4_Linux has no Cython?
> 
> This is probably my ignorance, I meant that the `PyAbel` C-extensions do not build for the rasberry pi installation: `pip install . --no-build-isolation`

Should work by default if you have ``gcc`` and ``python3-dev`` installed in Debian (plus Cython, as usual). I’ve tested right now on my [La Frite](https://libre.computer/products/aml-s805x-ac/), and everything looked fine:
```
=== Basis generation ===========================================
Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
two_point                6.3        27.4        92.4       326.2

=== Forward Abel transform =====================================
Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
hansenlaw_C              3.4        12.3        39.1       124.1
hansenlaw_Python        28.2       112.2       465.6      1732.3
two_point                2.9        11.2        59.4       385.8

=== Inverse Abel transform =====================================
Method               n = 257     n = 513    n = 1025    n = 2047
----------------------------------------------------------------
hansenlaw_C              4.6        16.7        51.9       170.8
hansenlaw_Python        29.7       115.5       478.4      1762.4
two_point                1.8        10.4        73.3       552.8
```
(the overall speed is not impressive, but I bought this board with 16 GB eMMC included for only 30 USD and not for such tasks...).

MikhailRyazanov:
> No, [daz](https://github.com/chainer/daz) no longer appears to be installable

Is it incompatible with recent pip/Setuptools, or has some other problems?

I did some experiment to see whether it can be ported to Cython to include in PyAbel. Takes ~5 kB of source code (most of which is documentation) and compiles into <30 kB library (about the same as the original). Another option would be to release it as a separate package on PyPI, with binary wheels, such that anyone can just install it, but then it also needs to be maintained... @DanHickstein, do you have any opinion on this? (#246 is the previous discussion; in short, we have noticed a significant effect on ``basex`` and ``basex(var.reg.)``. Maybe ``daun(var.reg.)`` is also affected, but it did not exist at that time.)

stggh:
> Should work by default if you have gcc and python3-dev installed in Debian (plus Cython

Great, thanks for that, `_C`'s now working on Rpi 5B. Benchmarks to come

> La Frite

Interesting. Only 1GB RAM though

MikhailRyazanov:
> Interesting. Only 1GB RAM though

More than enough for it. I also have a VPS which runs on only 512 MB RAM without any problems (current uptime is ~2 years). :–)

stggh:
> `daz` - Is it incompatible with recent pip/Setuptools, or has some other problems?

- `imp` module has been depreciated for python 12+. However, this [website](https://stackoverflow.com/questions/77401730/modulenotfounderror-no-module-named-imp) says install `zombie-imp` which provides a copy of `imp`. Yes, that works
- `pip install .` fails, however, `python setup.py install` works.

Running i7-9700 with `daz` now ...
RPi5 benchmarks still in progress ..

MikhailRyazanov:
@stggh, any news?

And regarding ``daz`` and porting it to Cython: it turned out that integrating it into PyAbel would complicate the setup too much, so I actually did “release it as a separate package on PyPI”: https://pypi.org/project/sseflags/
Now we should be able to replace ``daz`` in ``pyabel-timing.py`` and documentation.

stggh:
@MikhailRyazanov, sorry, for the long delay, running benchmarks is a commitment!  And I live in fear of doing something dumb(*), that you may need to correct me on. So, I do spend some time checking things, and re-running the codes.

Thanks for the new `sseflags` implementation of the python `daz/ftz` flags.  Very clever, and useful. I had some minor issues running your `sseflags` test programs, which I will report to your github site. However, I saw enough within `test.py` to add to `pyabel-timimg.py` 
lines 117-121:
```python
try:
    import sseflags as sse
    sse.set_flags(daz=True, ftz=True)
except ModuleNotFoundError:
    print('\nWarning! No daz module. Intel CPUs can show poor performance.\n')
```

I only see a minor improvement, in the benchmarks, in the context of my commend above(*). The benchmarks `pyabel-timing.py` were run with a quiet system, no other user processes running.

A sample of the results. I will upload the full directory later, following any suggestions/comments you might have:
```
(py314)[Gibbo:.+0_Linux_daz_ftz]$ cat basex.dat 
# basex           iabel(ms)           basis(ms)
    5             0.00063             0.52386
    9             0.00069             0.61059
   17             0.00088             0.89248
   33             0.00159             1.91295
   65             0.00531             5.72319
  129             0.01692            24.97231
  257             0.05988           126.78522
  513             0.30730           572.32660
 1025             2.19889          3515.34910
 2049            16.73988         24269.90239
 4097           138.53222        180591.31925
```
Note truncated at `n = 4097` as `basex` makes a significant demand on my 32GB system and I have limited patience.

```(py314)[Gibbo:.+/fig_benchmarks]$ cat i7-9700_Linux_no_daz_ftz/basex.dat
# basex           iabel(ms)           basis(ms)
    5             0.00068             0.52073
    9             0.00073             0.61293
   17             0.00090             0.87453
   33             0.00164             1.88229
   65             0.00546             5.66779
  129             0.01694            24.24313
  257             0.06042           122.75611
  513             0.31027           575.22231
 1025             2.18528          3511.84464
 2049            20.09893         24351.41109
 4097           369.04939        184164.60174
```
```
(py314)[Gibbo:.+0_Linux_daz_ftz]$ cat basex\(var\).dat 
# basex(var)      iabel(ms)           
    5             0.45053
    9             0.45149
   17             0.45708
   33             0.46949
   65             0.50402
  129             3.61059
  257            34.95868
  513            81.56267
 1025            56.92075
 2049           156.77931
 4097           529.82501
```
```(py314)[Gibbo:.+/fig_benchmarks]$ cat i7-9700_Linux_no_daz_ftz/basex\(var\).dat 
# basex(var)      iabel(ms)           
    5             0.44517
    9             0.43730
   17             0.44211
   33             0.45383
   65             0.50003
  129             3.39308
  257            49.60956
  513           114.21545
 1025           115.79919
 2049           201.71347
 4097           580.19884
```
```
(py314)[Gibbo:.+0_Linux_daz_ftz]$ cat hansenlaw_C.dat 
# hansenlaw_C     iabel(ms)           
    5             0.08000
    9             0.08111
   17             0.08460
   33             0.09044
   65             0.10041
  129             0.13383
  257             0.24623
  513             1.21788
 1025             3.70593
 2049            12.15905
 4097            45.25307
 8193           154.73900
16385           607.45730
```
```
(py314)[Gibbo:.+inux_no_daz_ftz]$ cat hansenlaw_C.dat 
# hansenlaw_C     iabel(ms)           
    5             0.08520
    9             0.08751
   17             0.08901
   33             0.09747
   65             0.10647
  129             0.14030
  257             0.25908
  513             0.68852
 1025             2.43629
 2049             9.69257
 4097            47.09259
 8193           163.36386
16385           635.75550
```
The `benchmarks_i7-9700_Linux_daz_ftz` timing plot has consistent drop-out behaviour for `basex(var)` and `daun(var)`.

benchmarks_i7-9700_Linux_daz_ftz

Edit:  typo 'quite' -> 'quiet', 'user processes',  each 2nd table is the run with `daz=False, ftz==False`

Edit2: PS my "correct me on" comment was meant in good humour

MikhailRyazanov:
> that you may need to correct me on

As if I knew what and how to correct. :–)

These new i7-9700 results look very weird. I’ve checked on my old i5 (the only Intel currently accessible) and do not see such worsened performance for “(var)”:
transform_time-daz
(However, one strange thing is that now I do not see any difference between using and not using ``daz``, although initially it was noticed and tested on the same machine, and running ``sseflags.benchmarks`` with the same Python/NumPy/SciPy does show a clear effect. **Correction:** Actually, running ``abel.benchmark.AbelTiming`` still shows a large effect for the *forward* transform, [as it was before](https://github.com/PyAbel/PyAbel/issues/246), but we do not show it the benchmark plots; the inverse transform is now in general slightly faster than before (perhaps due to some improvements in NumPy or OpenBLAS) and is affected even less than before, so that is why “I did not see any difference” in the plots.)

Maybe there is something strange with NumPy/SciPy or the underlying BLAS on that particular i7? Buy it seems that investigating this becomes too complicated, so maybe then just retain what we had for other methods on i7-9700 and update only ``direct`` and ``hansenlaw`` in this PR? If I understand correctly, Macbook M3 and the new RPi5 look fine, so they will replace i7-6700 and the old RPi4, as planned.

> I had some minor issues running your ``sseflags`` test programs, which I will report to your github site.

Yes please! It is important to make sure that everything works before recommending it.

